PR #4, opened 2025-06-25
Fix sounds and merge

LSRCT:
- Fix sounds not working
- Fix merge not being available when task merged -> follow up -> new changes


LSRCT:
Closing this, sound should be on OS level so that it works when no browser is open.

PR #9, opened 2025-06-27
feat: visualize running dev servers in project management

LSRCT:
I often start dev servers, get distracted and leave them running by mistake. It's hard to find running dev servers currently.

This adds an icon to tasks and projects showing they are running dev servers.


stunningpixels:
As discussed, instead let's add section to navbar that will display all ongoing execution processes including devservers making it easy to identify what's going on

PR #11, opened 2025-06-27
feat: configure global MCP server settings in-app

ggordonhall:
Add a new MCP Servers textbox to the Settings page. Here, users MCP server config which we then apply to the global settings of whatever default executor they have set up.

If the user pastes this into the textbox with Claude as their default executor:
```
    "exa": {
      "command": "npx",
      "args": [
        "-y",
        "mcp-remote",
        "https://mcp.exa.ai/mcp?exaApiKey=your-exa-api-key"
      ]
    }
```
We'll go in and add this configuration to the global `mcpServers` object in `.claude.json`.

Screenshot 2025-06-30 at 14 11 40

### DIFF
--- a/frontend/src/pages/Settings.tsx
+++ b/frontend/src/pages/Settings.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
 import {
   Card,
   CardContent,


stunningpixels:
- As discussed yesterday, would be ideal to decouple the "Default Executor" setting from the MCP server configuration.
- Could be helpful to also print the path to the config file that will be updated (eg ~/.claude.json)
- Could also be helpful to show the user which part of the JSON they are editing, I initially added:

```
{
  "mcpServers": {
    "fetch": {
      "command": "uvx",
      "args": ["mcp-server-fetch"]
    }
  }
}
```

Which meant that my config was incorrect as it had `mcpServers` inside `mcpServers`

ggordonhall:
> * As discussed yesterday, would be ideal to decouple the "Default Executor" setting from the MCP server configuration.
> * Could be helpful to also print the path to the config file that will be updated (eg ~/.claude.json)
> * Could also be helpful to show the user which part of the JSON they are editing, I initially added:
> 
> ```
> {
>   "mcpServers": {
>     "fetch": {
>       "command": "uvx",
>       "args": ["mcp-server-fetch"]
>     }
>   }
> }
> ```
> 
> Which meant that my config was incorrect as it had `mcpServers` inside `mcpServers`

These have been addressed

ggordonhall:
@stunningpixels MCP Server configuration has been moved to a separate tab

PR #13, opened 2025-06-27
Support sound notification on windows

abcpro1:


ggordonhall:
Is this not better handled by https://github.com/BloopAI/vibe-kanban/pull/4?

abcpro1:
Tested again on Windows. The sound notification works.

PR #16, opened 2025-06-30
chore: remove checked-in sqlitedb

ggordonhall:


stunningpixels:
This can happen when the prepare-db script fails midway through, we should also add the db to gitignore

PR #23, opened 2025-06-30
Cross-platform sound support

abcpro1:
WAV files work on linux, macos, and windows with the builtin commands. Particularly `aplay` in Linux, which is the only preinstalled command in Ubuntu, only works with .wav files.

Added sound notification support for WSL2.


stunningpixels:
Looks good pls merge

PR #28, opened 2025-06-30
Code signing in CI

anastasiya1155:
Latest action example: https://github.com/BloopAI/vibe-kanban/actions/runs/15981616922
Release with assets: https://github.com/BloopAI/vibe-kanban/releases/tag/v0.0.26

Verified MacOS is running ok, but we should also test other platforms

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -165,10 +165,10 @@ jobs:
           mkdir -p dist
           if [[ "${{ matrix.os }}" == "windows-latest-l" ]]; then
             cp target/${{ matrix.target }}/release/vibe-kanban.exe dist/vibe-kanban-${{ matrix.name }}.exe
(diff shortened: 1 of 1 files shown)


stunningpixels:
Nit: can we normalise the name of the MCP server, we have mcp_task_server, vibe-kanban-mcp and mcp-server-{DIST} all going on. I would suggest we just stick to vibe-kanban-mcp and vibe-kanban-mcp-{DIST}

anastasiya1155:
> Nit: can we normalise the name of the MCP server, we have mcp_task_server, vibe-kanban-mcp and mcp-server-{DIST} all going on. I would suggest we just stick to vibe-kanban-mcp and vibe-kanban-mcp-{DIST}

Updated the workflow to use vibe-kanban-mcp. Here is the latest run https://github.com/BloopAI/vibe-kanban/actions/runs/15994630329, tested on macOS only

stunningpixels:
I think we should merge

PR #58, opened 2025-07-03
feat: PostHog product analytics

ggordonhall:
Lightweight usage analytics. Users can opt-out on either the Settings tab.

We only track events on the backend to ensure that analytics are robust to ad-blockers.

We track the following events:
- `session_start`
- `project_created`
- `task_created`
- `task_attempt_started`
- `task_attempt_finished` (with execution status)
- `task_changes_merged`
- `github_pr_created`
- `dev_server_started`

To Do:
- [x] Package analytics key in the production package

### DIFF
--- a/backend/src/models/config.rs
+++ b/backend/src/models/config.rs
@@ -261,7 +261,15 @@ impl Config {
 
             // Try to deserialize as is first
             match serde_json::from_str::<Config>(&content) {
-                Ok(config) => Ok(config),
+                Ok(mut config) => {
+                    if config.analytics_enabled.is_none() {
+                        config.analytics_enabled = Some(true);
+                    }
+
+                    // Always save back to ensure new fields are written to disk
+                    config.save(config_path)?;
+                    Ok(config)
+                }
                 Err(_) => {
                     // If full deserialization fails, merge with defaults
                     let config = Self::load_with_defaults(&content, config_path)?;


backend/src/main.rs
```diff
@@ -141,7 +141,13 @@ async fn main() -> anyhow::Result<()> {
...
 
     // Create app state
-    let app_state = AppState::new(pool.clone(), config_arc.clone());
+    let app_state = AppState::new(pool.clone(), config_arc.clone()).await;
+
+    // Track session start event
+    if app_state.get_analytics_enabled().await {
+        let analytics = app_state.analytics.read().await;
+        analytics.track_event(&services::generate_user_id(), "session_start", None);
+    }
```
anastasiya1155: Seems like this logic is duplicated in every place we track analytics events. Would be nice to have some utility for this, so we have less boilerplate and also we don't accidentally forget is analytics is enabled before tracking an event

ggordonhall: Good point, will sort

ggordonhall:
@stunningpixels Looked into moving analytics events out of `routes`. This is doable but a decently large piece of work to do well, so would rather defer to a future PR. The best long-term approach is to move logic out of both the `model` and `routes` layers into the `service` layer. For example, we'd create a `TaskAttemptService` that was responsible for starting a task, handling any errors, and analytics tracking, and a `ProjectService` that would be responsible for all project lifecycle events, as well as analytics tracking.

PR #78, opened 2025-07-04
Stderr Display & Gemini output display in conversation history

abcpro1:
- Implements normalized output for Gemini-CLI
- Adds error message display in the UI


abcpro1:
Yes, I'll work on that next in addition to supporting resume.

PR #80, opened 2025-07-07
Call claude via `npx`

abcpro1:


abcpro1:
@stunningpixels Tested and working. Ready to merge?

PR #86, opened 2025-07-07
fix: stop passing stale `task_id` from the frontend

ggordonhall:
This fixes issues where `get_branch_status` and `get_diff` in `TaskAttempt` would throw `TaskAttemptError::TaskNotFound`. This was because the frontend was passing a stale `task_id` that did not line up with the `task_attempt.id` in the `task_attempts` DB table.

### DIFF
--- a/frontend/src/components/tasks/TaskDetailsPanel.tsx
+++ b/frontend/src/components/tasks/TaskDetailsPanel.tsx
@@ -119,6 +119,9 @@ export function TaskDetailsPanel({
     handleSendFollowUp,
   } = useTaskDetails(task, projectId, isOpen);
 
+  // Use ref to track loading state to prevent dependency cycles
+  const diffLoadingRef = useRef(false);
+
   // Fetch diff when attempt changes
   const fetchDiff = useCallback(async () => {
     if (!projectId || !selectedAttempt?.id || !selectedAttempt?.task_id) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/tasks/TaskDetailsPanel.tsx
```diff
@@ -121,17 +121,22 @@ export function TaskDetailsPanel({
...
   const fetchDiff = useCallback(async () => {
-    if (!projectId || !task?.id || !selectedAttempt?.id) {
+    if (!projectId || !selectedAttempt?.id || !selectedAttempt?.task_id) {
       setDiff(null);
       setDiffLoading(false);
       return;
     }
 
+    // Prevent multiple concurrent requests
+    if (diffLoading) {
```
anastasiya1155: If it is not passed in the array of dependencies to useCallback it might cause bugs because it can be not the latest state.

ggordonhall: Would it work to just remove this `if` statement?

anastasiya1155: if we remove this if statement it would continue execution and make a diff request even if it is already in progress, not critical, but better to leave and just add it to the array

ggordonhall: The last commit led to some sort of infinite loop on the frontend. Rather than adding `diffLoading` to the list of dependencies, I've implemented Claude Code's suggestion which I don't fully understand, but which appears to work.

@anastasiya1155 Are there obvious issues with this approach?

anastasiya1155: seems like a good solution

PR #94, opened 2025-07-08
feat: improve agent conversation rendering

ggordonhall:
- Make diff box collapsible
- Format messages as markdown
- Render TODO and Glob tool arguments
- Improve consistency between Amp and Claude

**Amp**
Screenshot 2025-07-08 at 11 58 01

**Claude**
Screenshot 2025-07-08 at 11 58 49

### DIFF
--- a/frontend/src/components/tasks/TaskDetailsPanel.tsx
+++ b/frontend/src/components/tasks/TaskDetailsPanel.tsx
@@ -19,7 +19,14 @@ import {
   DialogHeader,
   DialogTitle,
 } from '@/components/ui/dialog';
-import { FileText, ChevronDown, ChevronUp, Trash2 } from 'lucide-react';
+import {
+  FileText,
+  ChevronDown,
+  ChevronUp,
+  Trash2,
+  Eye,
+  EyeOff,
+} from 'lucide-react';
 import type {
   TaskWithAttemptStatus,
   EditorType,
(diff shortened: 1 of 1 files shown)


frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <div className={className}>
+      {lines.map((line, index) => (
+        <div key={index}>
+          {parseMarkdown(line)}
+          {index < lines.length - 1 && <br />}
+        </div>
+      ))}
+    </div>
+  );
```
anastasiya1155: this component can cause performance issues, every time it is rendered (it happens a lot, e.g. on each change to any conversation prop, can happen many times per second) it splits content and renders each line with parseMarkdown, which uses regex and so on. Not sure if this is an issue for today, but can cause laggy behaviour in future, especially for long conversations.
Sometimes during development I can't reload the page as it gets stuck, but in prod didn't have any issues yet.

ggordonhall:
> Overall looks much better now, I still find that the area with conversation is too crammed and quite hard to read the steps and explanations (the red rectangle in the screenshot):
> 
> Image

Yes that's a good point. We could make the whole top part of the conversation box collapsible too, so you just see the task title.

anastasiya1155:
Also, minor one - can we add some favicon to the page? it usually uses some favicon from any other project I ran before or shows default empty icon.
Screenshot 2025-07-08 at 13 23 02

ggordonhall:
@anastasiya1155 For now have added a "Hide Diffs" toggle that allows you to completely hide all of the diff cards.

PR #125, opened 2025-07-10
Setup script for restored worktrees

abcpro1:


abcpro1:
Setup is executed correctly when dev server or followup is needed.

UI still doesn't show setup script output in this case for some reason but can be solved in a follwup PR as the current fix is useful enough on its own.

PR #126, opened 2025-07-11
Refactor TaskDetailsPanel

anastasiya1155:
- add useCallback to functions that are passed as props and useMemo to computed values (less rerenders)
- avoid passing whole project object in props (less rerenders)
- split TaskDetailsPanel into separate components for better maintainability and passing only relevant props (less rerenders)
- move common fields and handlers into context, so that children can take values directly instead of props drilling

We should test that functionality hasn't changed, code review can be a bit overwhelming

To be done in the next PRs:
LogsPanel
TaskDetailsToolbar

### DIFF
--- a/frontend/src/components/context/TaskDetailsContextProvider.tsx
+++ b/frontend/src/components/context/TaskDetailsContextProvider.tsx
@@ -1,5 +1,4 @@
 import {
-  createContext,
   Dispatch,
   FC,
   ReactNode,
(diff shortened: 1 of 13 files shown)


stunningpixels:
Could follow this with further refactoring to:
- frontend/src/components/tasks/TaskDetailsToolbar.tsx
- frontend/src/components/tasks/TaskDetails/NormalizedConversationViewer.tsx

Can probably move all network requests and state separate from UI components.

PR #136, opened 2025-07-11
Refactor TaskDetailsToolbar and LogsPanel, improve performance

anastasiya1155:
- extract smaller components for better maintainability
- split big context into smaller ones to avoid unnecessary re-renders when some part of state changes
- add useCallback and React.memo to reduce rerenders
- when polling some data check if it changed from previous time before setting to state to avoid unnecessary state updates and re-renders


LSRCT:
Tried this branch and tested:
- Project creating
- Github auth
- Basic tasks with Gemini, Claude and AMP
- Open a PR
- Ask follow up
Feels a lot snappier!

There is a bug when viewing a task that's in the 'merged' state where the diff fails. Probably backend related. @abcpro1

PR #145, opened 2025-07-11
feat: Allow custom GitHub OAuth app client ID via environment variable

westonplatter:
Enable devs to use their own GitHub OAuth app instead of the hardcoded Bloop AI client ID when building Vibe Kanban from source.

### Changes

  - Modified `backend/src/routes/auth.rs` to use option_env!("GITHUB_CLIENT_ID") instead of hardcoded client ID
  - Follows the same pattern as the existing `POSTHOG_API_KEY` environment variable
  - Falls back to the original Bloop AI client ID (`Ov23li9bxz3kKfPOIsGm`) if no environment variable is set
  - updated `README.md` with a descriptive table of this setting and other settings

### Usage

  To build with your own GitHub OAuth app:

```bash
GITHUB_CLIENT_ID=your_client_id_here pnpm run build
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,23 @@ This will start the frontend and backend with live reloading. A blank DB will be
 2. In the `npx-cli` folder run `npm pack`
 3. You can run your build with `npx [GENERATED FILE].tgz`
 
-### Custom GitHub OAuth App (Optional)
+
+### Environment Variables
+
+The following environment variables can be configured at build time or runtime:
+
+| Variable | Type | Default | Description |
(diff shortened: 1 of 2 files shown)


anastasiya1155:
Hey @westonplatter thanks for the contribution. The code looks good. If you could fix the 'cargo fmt' issue, we can merge it

westonplatter:
@anastasiya1155 thanks for pointing out the cargo fmt issue. Pushed up code changes for it.

zoosky:
Thx

PR #146, opened 2025-07-11
Add a privacy opt in during the onboarding (vibe-kanban)

stunningpixels:
Add a new dialog that triggers after the coding agent and IDE selection step, but before the GitHub login that asks the user whether they'd like to opt in to telemetry or not.

Please inform the user that telemetry sends high level usage data for example how many tasks are being created, but not the contents of any tasks, and can be opted out at any point later on in the settings, and that it really helps the developers if they opt in


stunningpixels:
Screenshot 2025-07-11 at 22 27 16

Preview, open for comments

PR #173, opened 2025-07-14
Normalise API calls on FE

anastasiya1155:
- all api calls are defined in one file - `lib/api.ts` and all components use it instead of custom fetch and makeRequest calls
- use the same loader everywhere
- minor fixes

### DIFF
--- a/frontend/src/components/context/TaskDetailsContextProvider.tsx
+++ b/frontend/src/components/context/TaskDetailsContextProvider.tsx
@@ -18,11 +18,7 @@ import type {
   TaskWithAttemptStatus,
   WorktreeDiff,
 } from 'shared/types.ts';
-import {
-  attemptsApi,
-  executionProcessesApi,
-  withErrorHandling,
-} from '@/lib/api.ts';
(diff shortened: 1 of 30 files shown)


frontend/src/components/context/TaskDetailsContextProvider.tsx
```diff
@@ -94,30 +93,30 @@ const TaskDetailsProvider: FC<{
...
-        if (isBackgroundRefresh) {
-          setIsBackgroundRefreshing(true);
-        } else {
-          setDiffLoading(true);
-        }
-        setDiffError(null);
-        const response = await makeRequest(
-          `/api/projects/${projectId}/tasks/${selectedAttempt.task_id}/attempts/${selectedAttempt.id}/diff`
-        );
+      diffLoadingRef.current = true;
```
stunningpixels, line 92: 🤔

Unrelated to this change, but not sure this ref pattern is necessary?

anastasiya1155: this ref makes sure we don't make this request multiple times, while previous response is still loading

PR #177, opened 2025-07-15
Add Slash Command input for Create Attempt when CC is selected

kylelol:
- Add slash_command column to task_attempts table via migration
- Update TaskAttempt model to include slash_command field
- Update Claude executor to use task attempt slash command
- Add slash command input to CreateAttempt component (Claude executor only)
- Remove slash command from Task model and TaskFormDialog
- Update TypeScript types to reflect new structure
- Regenerate .sqlx files for clean database schema

This allows each task attempt to have its own optional slash command, providing flexibility for different execution strategies without cluttering the task creation interface.

Screenshot 2025-07-14 at 10 15 58 PM

Screenshot 2025-07-14 at 10 18 59 PM


LSRCT:
Hi, thanks for opening a PR!

Unfortunately we cannot merge this for two reasons:
- Maintaining compatibility with all possible slash commands is a lot of work
- Given example `/read-files` does not actually exist for claude: https://docs.anthropic.com/en/docs/claude-code/slash-commands#built-in-slash-commands

Feel free to open an issue with the behaviour you tried to add with the slash commands.

PR #189, opened 2025-07-15
Better keyboard navigation

anastasiya1155:


stunningpixels:
FYI might be worth moving some of the logic to frontend/src/lib/keyboard-shortcuts.ts so we can centralise and re-use, especially if there are conflicting shortcuts

ggordonhall (approved):

PR #192, opened 2025-07-15
Make GitHub auth dialog dismissible  (vibe-kanban)

stunningpixels:
Currently the GitHub auth dialog shows any time the user isn't logged into GitHub.

Let's change this, instead the user should be able to exit the dialog if they don't want to sign in with github.

Once this is changed, when the user's creating a PR in frontend/src/components/tasks/Toolbar/CreatePRDialog.tsx and they click "create pr" and they are not logged in, it will show the message "GitHub authentication not configured. Please sign in with GitHub." in the frontend/src/components/tasks/TaskDetailsToolbar.tsx which is behind the dialog, so the user can't really see the message.


frontend/src/App.tsx
```diff
@@ -102,15 +94,48 @@ function AppContent() {
...
-      const notAuthenticated =
-        !updatedConfig.github?.username || !updatedConfig.github?.token;
-      setShowGitHubLogin(notAuthenticated);
     } catch (err) {
       console.error('Error saving config:', err);
     }
   };
 
+  const handleGitHubLoginComplete = async () => {
+    if (!config) return;
```
anastasiya1155: why do we check for previous config?

frontend/src/App.tsx
```diff
@@ -102,15 +94,48 @@ function AppContent() {
...
+
+      // If user skipped (no GitHub token), we need to manually set the acknowledgment
+      if (!latestConfig.github?.token) {
+        const updatedConfig = {
+          ...latestConfig,
+          github_login_acknowledged: true,
+        };
+        updateConfig(updatedConfig);
+        await configApi.saveConfig(updatedConfig);
+      }
```
anastasiya1155: shouldn't we just update config with github_login_acknowledged as soon as the user either closes the popup or logs in? Also, we could have some function in useConfig() to just refetch config instead of requesting it here and updating it twice

frontend/src/App.tsx
```diff
@@ -102,15 +94,48 @@ function AppContent() {
...
+      // After GitHub login (success or skip), show privacy opt-in if not acknowledged
+      if (!config.telemetry_acknowledged) {
+        setShowPrivacyOptIn(true);
+      }
+    } catch (err) {
+      console.error('Error refreshing config:', err);
+      // Still show privacy opt-in even if config refresh fails
+      if (!config.telemetry_acknowledged) {
+        setShowPrivacyOptIn(true);
+      }
```
anastasiya1155: we can move it to finally to avoid duplication

frontend/src/components/GitHubLoginDialog.tsx
```diff
@@ -60,7 +61,10 @@ export function GitHubLoginDialog({
           setPolling(false);
           setDeviceState(null);
           setError(null);
-          window.location.reload(); // reload config
+          // Add a small delay to ensure backend has time to save the token
+          setTimeout(() => {
+            onOpenChange(false);
+          }, 500);
```
anastasiya1155: we don't necessarily need this delay as we already have one inside handleGitHubLoginComplete

frontend/src/App.tsx
```diff
@@ -102,15 +94,48 @@ function AppContent() {
...
 
+  const handleGitHubLoginComplete = async () => {
+    if (!config) return;
+
+    setShowGitHubLogin(false);
+
+    try {
+      // Add a delay to ensure backend has time to save the GitHub token
+      // (if user authenticated) before refreshing the config
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
anastasiya1155: Not sure if we need this delay as `/poll` response will come after the token is saved and github login is acknowledged, so the popup won't close before the token is saved unless the user manually just closed the popup

Issue #176, opened 2025-07-15
Send message input box is small and non-resizable


sammcj:
It would be great if the send message input box in the active agent was larger and could be resized.

At present it's quite cramped when writing more than a short sentence:

<img width="840" height="347" alt="Image" src="https://github.com/user-attachments/assets/b0325380-202c-4f3c-ba70-4ad1fafe95d8" />

PR #194, opened 2025-07-15
Textareas should expand with content (vibe-kanban)

stunningpixels:
The Description field when creating a new task, and the "ask a follow up question..." textarea should expand vertically as content increases, up to a maximum size


stunningpixels:
Fixes https://github.com/BloopAI/vibe-kanban/issues/176

PR #197, opened 2025-07-15
feat: task templates (vibe-kanban)

ggordonhall:
Adds templates for task titles and descriptions. These can be either configured globally (for all projects) via the settings page, or for a particular project via the project settings page.

**Project Settings:**
Screenshot 2025-07-15 at 17 48 12

**Global Settings:**
Screenshot 2025-07-15 at 17 48 40

**Task creation:**
Screenshot 2025-07-15 at 17 49 02


frontend/src/components/projects/project-form.tsx
```diff
@@ -150,216 +152,252 @@ export function ProjectForm({
...
+
+        {isEditing ? (
+          <Tabs defaultValue="general" className="w-full -mt-2">
+            <TabsList className="grid w-full grid-cols-2 mb-4">
+              <TabsTrigger value="general">General</TabsTrigger>
+              <TabsTrigger value="templates">Task Templates</TabsTrigger>
+            </TabsList>
+            <TabsContent value="general" className="space-y-4">
+              <form onSubmit={handleSubmit} className="space-y-4">
+                {renderForm()}
```
anastasiya1155: It is better to have this form as a component rather than a function inside its parent. Alternatively, adding useMemo() can reduce re-renders

frontend/src/components/TaskTemplateManager.tsx
```diff
@@ -0,0 +1,308 @@
...
+    description: '',
+  });
+  const [saving, setSaving] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    fetchTemplates();
+  }, [projectId, isGlobal]);
+
+  const fetchTemplates = async () => {
```
anastasiya1155: We should add useCallback to all these functions. It is not very important for this component as it is not a major one, but leaving it like this causes LLM to generate more code like this, which in the end worsens performance a lot.

frontend/src/components/tasks/TaskFormDialog.tsx
stunningpixels, line 1: Dialog should respond to global keyboard shortcuts in `frontend/src/lib/keyboard-shortcuts.ts`, so enter = create

backend/src/routes/task_templates.rs
```diff
@@ -0,0 +1,178 @@
...
+            StatusCode::INTERNAL_SERVER_ERROR,
+            Json(ApiResponse::error(&format!(
+                "Failed to fetch template: {}",
+                e
+            ))),
+        )),
+    }
+}
+
+pub async fn create_template(
```
stunningpixels, line 84: From the error message it looks like we want to reject tasks that have the same name, but this isn't being enforced

Screenshot 2025-07-16 at 11 16 12

stunningpixels (changes requested):
To increase adoption I would add an icon next to the create button that opens a dropdown list of your templates, clicking on an entry would open the `create task dialog` with the template info pre-filled.

Screenshot 2025-07-16 at 11 18 27

We could also add some really good default templates as part of the DB migration. Perhaps one for bug fix, debug, plan, cleanup

backend/src/routes/task_templates.rs
ggordonhall: Interestingly, although we already have a unique index configured on `(project_id, template_name)` SQLite considers all `NULL` values unique: https://sqlite.org/lang_createindex.html

Will address.

PR #196, opened 2025-07-15
feat: Claude Code Router (CCR) executor

abcpro1:
Implement Claude Code Router executor.

resolves #182.

PR #209, opened 2025-07-16
chore: standardise executor type naming to kebab-case

abcpro1:
Follow-up to #196.


ggordonhall (commented):
Tested a migration from a project where I'd created a couple of tasks with `charmopencode`. I receive this error on start:

```
[0] Error: unknown variant `charmopencode`, expected one of `echo`, `claude`, `amp`, `gemini`, `setup-script`, `claude-code-router`, `charm-opencode`
[0] [Finished running. Exit status: 1]
```
Note that I'd also set my default executor to `charmopencode`.

PR #215, opened 2025-07-16
Open project in IDE doesn't work (vibe-kanban)

stunningpixels:
Open task attempt in IDE button does, but open project in IDE doesn't. It gives `Failed to parse the request body as JSON: EOF while parsing a value at line 1 column 0` as the frontend doesn't send any body with POST request


ggordonhall:
I think this has just been fixed by https://github.com/BloopAI/vibe-kanban/pull/213

PR #224, opened 2025-07-16
Enable follow up after stopping attempt (vibe-kanban)

LSRCT:
I want the user to be able to continue a task after he pressed the stop attempt button via the follow-up functionality


stunningpixels:
One thing to check - attempt has lived long enough to get a session ID

LSRCT:
> One thing to check - attempt has lived long enough to get a session ID

You're absolutely right! Follow ups should not be possible for the setup script, changed that.
We already handle the case when the agent process starts but is killed before getting a session ID.

stunningpixels (approved):

PR #238, opened 2025-07-17
Improve conversation logs rendering

anastasiya1155:
Render only last 100 entries from the conversation and add a button to load previous ones to reduce the rendering load.


backend/src/routes/task_attempts.rs
stunningpixels, line 1: Should we delete the old endpoint for normalized logs?

anastasiya1155: sure

stunningpixels (approved):

PR #239, opened 2025-07-17
sst Opencode

### LINKED ISSUE #169: OpenCode has no `--output-format=json` param.
when launching opencode, vibe-kanban might pass like:
```
opencode -p "project_id: 7a8b9c0d-1e2f-3a4b-5c6d-7e8f9a0b1c2d\n\nTask title: 实现用户认证功能\nTask description: 添加通过邮箱和密码进行注册和登录的功能。" --output-format=json
```, 
but sst/opencode has no --output-format=json param.

so opencode is not working in vibe-kanban.

since kimi k2 is so good at tool calling, k2 + opencode is really an alternative to claude code.

### PR DESCRIPTION
**Pull request by abcpro1**
resolves #169.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -26,6 +26,7 @@ chrono = { version = "0.4", features = ["serde"] }
 uuid = { version = "1.0", features = ["v4", "serde"] }
 ts-rs = { version = "9.0", features = ["uuid-impl", "chrono-impl", "no-serde-warnings"] }
 dirs = "5.0"
+xdg = "3.0"
 git2 = "0.18"
 async-trait = "0.1"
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by ggaabe** [comment]
good work!

File: backend/src/executors/sst_opencode/tests.rs [hunk 1]
```diff
@@ -0,0 +1,329 @@
+use super::*;
+use crate::{
+    executor::ActionType,
+    executors::sst_opencode::{format_opencode_content_as_normalized_json, SstOpencodeExecutor},
+};
+
+#[test]
```
ggordonhall: Let's move these tests inline

ggordonhall (approved):

PR #241, opened 2025-07-17
fix: distinguish between stopped and failed task attempts

ggordonhall:
- Display different messages depending on whether a `TaskAttempt` errored or was stopped.

### DIFF
--- a/frontend/src/components/tasks/TaskDetails/LogsTab.tsx
+++ b/frontend/src/components/tasks/TaskDetails/LogsTab.tsx
@@ -76,8 +76,8 @@ function LogsTab() {
     );
   }
 
-  // When setup failed, show error message and conversation
-  if (isSetupFailed) {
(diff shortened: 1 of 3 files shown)


anastasiya1155 (approved):

frontend/src/components/tasks/TaskDetails/LogsTab.tsx
```diff
@@ -99,6 +102,27 @@ function LogsTab() {
...
+        );
+
+    return (
+      <div className="h-full overflow-y-auto">
+        <div className="mb-4">
+          <p className="text-lg font-semibold mb-2">Setup Script Stopped</p>
+        </div>
+
+        {setupProcess && (
+          <NormalizedConversationViewer executionProcess={setupProcess} />
```
stunningpixels: Would be nice to avoid this duplication

PR #254, opened 2025-07-18
fix: config migration and CI windows build fixes for opencode

### LINKED ISSUE #250: opencode update broke install
Now getting :

Error: unknown variant opencode, expected one of echo, claude, amp, gemini, setup-script, setup_script, claude-code-router, charm-opencode, charmopencode

How do I clean my install? npx and npn uninstall do not fix this

abcpro1:
fixes #250

### DIFF
--- a/backend/src/executor.rs
+++ b/backend/src/executor.rs
@@ -428,7 +428,14 @@ impl ExecutorConfig {
                 dirs::home_dir().map(|home| home.join(".gemini").join("settings.json"))
             }
             ExecutorConfig::SstOpencode => {
-                xdg::BaseDirectories::with_prefix("opencode").get_config_file("opencode.json")
+                #[cfg(unix)]
+                {
+                    xdg::BaseDirectories::with_prefix("opencode").get_config_file("opencode.json")
+                }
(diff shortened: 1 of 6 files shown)


backend/src/executor.rs
```diff
@@ -428,7 +429,14 @@ impl ExecutorConfig {
                 dirs::home_dir().map(|home| home.join(".gemini").join("settings.json"))
             }
             ExecutorConfig::SstOpencode => {
-                xdg::BaseDirectories::with_prefix("opencode").get_config_file("opencode.json")
+                #[cfg(unix)]
+                {
+                    xdg::BaseDirectories::with_prefix("opencode").get_config_file("opencode.json")
```
ggordonhall, line 434: Why do we need this library in this case? Why can't we do `dirs::config_dir().join(...)`?

ggordonhall (approved):

backend/src/executor.rs
abcpro1: `sst opencode` uses XDG on macOS. `dirs::config_dir` returns `~/Library/Application Support/...` instead of `~/.config`

PR #261, opened 2025-07-18
Prevent response being created by non-generators (vibe-kanban)

stunningpixels:
Currently API responses are created using the success and error generators, and also using manual creation of the ApiResponse struct.

Let's enforce using generators and refactor the code by updating backend/src/models/api_response.rs to:

```
mod response {
    use serde::Serialize;
    use ts_rs::TS;

    #[derive(Debug, Serialize, TS)]
    #[ts(export)]
    pub struct ApiResponse<T> {
        success: bool,
        data: Option<T>,
        message: Option<String>,
    }

    impl<T> ApiResponse<T> {
        /// Creates a successful response, with `data` and no message.
        pub fn success(data: T) -> Self {
            ApiResponse {
                success: true,
                data: Some(data),
                message: None,
            }
        }

        /// Creates an error response, with `message` and no data.
        pub fn error(message: &str) -> Self {
            ApiResponse {
                success: false,
                data: None,
                message: Some(message.to_string()),
            }
        }
    }
}

// Re-export the type, but its fields remain private
pub use response::ApiResponse;
```

Then we can check for cargo errors and refactor the code incrementally until all the patterns have been changed to the generators

### DIFF
--- a/backend/src/models/api_response.rs
+++ b/backend/src/models/api_response.rs
@@ -1,28 +1,35 @@
-use serde::Serialize;
-use ts_rs::TS;
+mod response {
+    use serde::Serialize;
+    use ts_rs::TS;
 
-#[derive(Debug, Serialize, TS)]
-#[ts(export)]
-pub struct ApiResponse<T> {
-    pub success: bool,
-    pub data: Option<T>,
-    pub message: Option<String>,
-}
-
-impl<T> ApiResponse<T> {
-    pub fn success(data: T) -> Self {
-        Self {
-            success: true,
-            data: Some(data),
-            message: None,
-        }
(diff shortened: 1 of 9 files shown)


backend/src/models/api_response.rs
```diff
@@ -1,28 +1,35 @@
...
+            ApiResponse {
+                success: false,
+                data: None,
+                message: Some(message.to_string()),
+            }
         }
     }
 }
+
+// Re-export the type, but its fields remain private
```
ggordonhall, line 34: I don't see why this is necessary. Why not just define the type directly in the `api_response` module and get rid of `mod response { ... }`?

PR #282, opened 2025-07-20
Adds SST OpenCode model selection for task execution

### LINKED ISSUE #266: Cannot select model to use in SST Opencode
Opencode lets you select many different models and configure your own providers from local models and openrouter models. It would be great if we could select which one to use from all available models in a dropdown

ggaabe:
When SST Opencode is the selected coding agent, all available models are listed (via `opencode models` command) and the user can select which one to use for the task when it starts (via `opencode --model provider/model-id` command).

Resolves #266 

I had put up another PR before but other updates caused merge conflicts; this is a new up-to-date PR

### DIFF
--- a/backend/src/executor.rs
+++ b/backend/src/executor.rs
@@ -441,6 +441,10 @@ impl FromStr for ExecutorConfig {
 
 impl ExecutorConfig {
     pub fn create_executor(&self) -> Box<dyn Executor> {
+        self.create_executor_with_model(None)
+    }
+
+    pub fn create_executor_with_model(&self, model: Option<String>) -> Box<dyn Executor> {
         match self {
             ExecutorConfig::Echo => Box::new(EchoExecutor),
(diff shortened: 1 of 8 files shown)


ggordonhall:
Hi @ggaabe,

I appreciate the contribution, but we're trying to minimise custom coding agent configuration logic. Each agent has lots of it, and we can't support it all in Vibe Kanban. In this case, can't you configure the default opencode model in `opencode.json`?

Closing this because if we were to support this configuration I'd rather handle it in Vibe Kanban's own configuration file.

ggaabe:
Hi @ggordonhall 

You can set a default that opencode can always start with in the configuration file, but I don't find that it makes sense to always use the same model for every kind of task. Having a lot more control over which model opencode uses would be a lot more powerful and gives us a wider range of models and providers to choose from that we can configure on a per-task basis.

Otherwise, why even provide multiple agent CLI options to begin with?

PR #322, opened 2025-07-22
Cross-platform project script UI placeholders

### LINKED ISSUE #307: [Bug] Script execution on Windows has issues with spaces, quotes, and shebangs
There are two related issues with how Vibe Kanban handles the "Setup Script" and "Dev Server Script" fields on Windows, making it very difficult to configure projects with standard file paths or scripts.

1.  **The input fields do not handle paths with spaces or quotes.** When the repository or script is located at a path containing spaces, it's impossible to enter the path correctly in the Project Settings UI. The input field does not seem to correctly parse quoted paths (e.g., `"C:\My Project\script.bat"`) or paths with unquoted spaces. This forces users to create workarounds like symbolic links to space-free paths.
2.  **Shebangs (`#!`) in shell scripts are ignored.** When providing a path to a `.sh` script, the application does not seem to respect the shebang line (e.g., `#!/bin/bash`). It appears to execute the script with the default Windows command prompt (`cmd.exe`), which cannot interpret bash syntax, causing the script to fail. This forces users to create `.bat` file wrappers that explicitly call a bash interpreter.

## To Reproduce

**To Reproduce the Spaces/Quotes Issue:**
1. On Windows, create a project where the repository path contains spaces (e.g., `C:\My Project`).
2. Place a setup script (e.g., `setup.bat`) inside that project.
3. In the Vibe Kanban Project Settings, attempt to enter the full path `C:\My Project\setup.bat` into the "Setup Script" field.
4. Observe that the configuration fails or the path is not correctly used.

**To Reproduce the Shebang Issue:**
1. In Project Settings, set the "Setup Script" path directly to a shell script (e.g., `setup.sh`).
2. Ensure the script contains a shebang line like `#!/bin/bash` and uses bash-specific syntax.
3. Run a task that triggers the setup script.
4. Observe that the script fails because it is not executed by `bash`.

## Expected behavior
1. The UI input fields for script paths should correctly handle paths that contain spaces, ideally by accepting standard quoting.
2. The application should respect the shebang line in script files on Windows to allow for the direct use of `.sh` scripts without requiring `.bat` wrappers.

## Desktop:
* **OS:** Windows

## Additional context
These UI and execution limitations force users into complex workarounds (like creating symlinks or moving scripts to new locations), which can expose deeper bugs in the application's handling of file paths and worktrees. Resolving these usability issues would greatly improve the setup experience on Windows.

abcpro1:
resolves #307

### DIFF
--- a/backend/src/routes/health.rs
+++ b/backend/src/routes/health.rs
@@ -1,20 +1,7 @@
 use axum::response::Json;
-use serde_json::{json, Value};
 
 use crate::models::ApiResponse;
 
-pub async fn health_check() -> Json<ApiResponse<Value>> {
-    let info = os_info::get();
-
-    let system_info = json!({
-        "status": "OK",
-        "system": {
-            "os_type": info.os_type().to_string(),
-            "os_version": info.version().to_string(),
-            "architecture": info.architecture().unwrap_or("unknown").to_string(),
-            "bitness": info.bitness().to_string(),
-        }
-    });
-
-    Json(ApiResponse::success(system_info))
+pub async fn health_check() -> Json<ApiResponse<String>> {
+    Json(ApiResponse::success("OK".to_string()))
 }
--- a/backend/src/routes/health.rs
+++ b/backend/src/routes/health.rs
@@ -1,7 +1,20 @@
 use axum::response::Json;
+use serde_json::{json, Value};
 
 use crate::models::ApiResponse;
 
-pub async fn health_check() -> Json<ApiResponse<String>> {
-    Json(ApiResponse::success("OK".to_string()))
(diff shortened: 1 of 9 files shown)


backend/src/routes/health.rs
```diff
@@ -1,7 +1,20 @@
 use axum::response::Json;
+use serde_json::{json, Value};
 
 use crate::models::ApiResponse;
 
-pub async fn health_check() -> Json<ApiResponse<String>> {
-    Json(ApiResponse::success("OK".to_string()))
+pub async fn health_check() -> Json<ApiResponse<Value>> {
```
stunningpixels: Add as "environment" object to config endpoint

PR #345, opened 2025-07-25
MCP support for Codex executor

abcpro1:


crates/server/src/routes/config.rs
```diff
@@ -337,3 +331,46 @@ fn set_mcp_servers_in_config_path(
 
     Ok(())
 }
+
+/// Helpers for handling different config file formats (JSON vs TOML)
```
ggordonhall: Can you stick these in a new file?

We should also make it clear that these are for reading and writing agent config files, not ours: `read_agent_config`, `write_agent_config`

ggordonhall (approved):

crates/server/src/routes/config.rs
abcpro1: `crates/server/src/mcp/agent_config.rs`?

PR #352, opened 2025-07-28
refactor: move stranded `git2` logic out of `models`

ggordonhall:


crates/core/src/routes/projects.rs
stunningpixels, line 1: I think `create_project_branch` might be dead, can't think of anywhere we would use

stunningpixels (approved):

PR #355, opened 2025-07-29
Refactor auth endpoints

LSRCT:
Add auth service

Move auth logic to service

Add auth router and service integration to deployment

Refactor auth service and routes to use octocrab

Refactor auth error handling and improve token validation responses


crates/core/src/routes/mod.rs
```diff
@@ -26,6 +26,7 @@ pub fn router(deployment: DeploymentImpl) -> IntoMakeService<Router> {
         .merge(task_attempts::router(&deployment))
         .merge(execution_processes::router(&deployment))
         .merge(task_templates::router(&deployment))
+        .merge(auth::auth_router(&deployment))
```
stunningpixels: naming consistency router not auth_router

crates/core/src/routes/auth.rs
```diff
@@ -1,262 +1,103 @@
...
-                .track_analytics_event("$identify", Some(props))
-                .await;
-        }
-
-        ResponseJson(ApiResponse::success("GitHub login successful".to_string()))
-    } else {
-        ResponseJson(ApiResponse::error("No access token yet"))
+    // Save to config
+    {
+        let config_path = utils::assets::config_path();
```
stunningpixels, line 54: I think we can use the config service for this

LSRCT: Currently we have `utils::assets::config_path()` in any place the path is required. (e.g. routes/config.rs)
Leaving this to limit this PR to auth changes only.

PR #362, opened 2025-07-30
fix log normalizer selection

abcpro1:


crates/services/src/services/container.rs
```diff
@@ -156,10 +156,61 @@ pub trait ContainerService {
...
+                    return None;
+                }
+                Err(e) => {
+                    tracing::error!("Failed to fetch execution process {}: {}", id, e);
+                    return None;
+                }
+            };
+
+            // Get the task attempt to determine correct directory
+            let task_attempt =
```
stunningpixels: We should create `parent_task_attempt` helper for ExecutionProcess, I have been creating these for other models as needed, eg:

https://github.com/BloopAI/vibe-kanban/blob/deployments/crates/db/src/models/task_attempt.rs#L179

PR #365, opened 2025-07-31
Claude Code

abcpro1:


crates/executors/src/executors/claude.rs
```diff
@@ -0,0 +1,981 @@
...
+    /// Create a new Claude executor with default settings
+    pub fn new() -> Self {
+        Self {
+            executor_type: "Claude Code".to_string(),
+            command: "npx -y @anthropic-ai/claude-code@latest -p --dangerously-skip-permissions --verbose --output-format=stream-json".to_string(),
+        }
+    }
+
+    /// Create a new Claude executor in plan mode with watchkill script
+    pub fn new_plan_mode() -> Self {
```
stunningpixels, line 177: Is this used?

crates/executors/src/executors/claude.rs
```diff
@@ -0,0 +1,981 @@
...
+
+    async fn spawn_follow_up(
+        &self,
+        current_dir: &PathBuf,
+        prompt: &str,
+        session_id: &str,
+    ) -> Result<AsyncGroupChild, ExecutorError> {
+        let (shell_cmd, shell_arg) = get_shell_command();
+        let claude_command = format!(
+            "npx -y @anthropic-ai/claude-code@latest continue {} --output-format=stream-json",
```
stunningpixels: If you can override the initial, you should also be able to override the follow up command

crates/executors/src/executors/claude.rs
abcpro1: will be used in a followup. otherwise will be removed.

PR #380, opened 2025-08-01
Codex

abcpro1:


crates/executors/src/executors/codex.rs
```diff
@@ -0,0 +1,952 @@
...
+        }
+
+        Ok(child)
+    }
+
+    fn normalize_logs(&self, msg_store: Arc<MsgStore>, current_dir: &PathBuf) {
+        let entry_index_provider = EntryIndexProvider::new();
+
+        // Process stderr logs for session extraction only (errors come through JSONL)
+        SessionHandler::start_session_id_extraction(msg_store.clone());
```
stunningpixels, line 213: Does this need to happen in a separate thread? Could it just be another match on the listener below in this method

crates/executors/src/executors/codex.rs
```diff
@@ -0,0 +1,952 @@
...
+        tokio::spawn(async move {
+            let mut stream = msg_store.stdout_lines_stream().await;
+
+            while let Some(Ok(line)) = stream.next().await {
+                let trimmed = line.trim();
+                if trimmed.is_empty() {
+                    continue;
+                }
+
+                match serde_json::from_str::<serde_json::Value>(trimmed) {
```
stunningpixels: Is there a reason not to do `match serde_json::from_str::< CodexJson >(json)` here? Rather than in `parse_codex_json()`?

crates/executors/src/executors/codex.rs
```diff
@@ -0,0 +1,952 @@
...
+                    timestamp: None,
+                    entry_type: NormalizedEntryType::SystemMessage,
+                    content,
+                    metadata: Some(serde_json::to_value(self).unwrap_or(serde_json::Value::Null)),
+                }])
+            }
+            CodexJson::Prompt { .. } => None, // Skip prompt messages
+            CodexJson::StructuredMessage { msg, .. } => msg.to_normalized_entries(
+                current_dir,
+                serde_json::to_value(self).unwrap_or(serde_json::Value::Null),
```
stunningpixels: Are we passing the `msg` as a Value ? If so wouldn't this be duplicate

stunningpixels (changes requested):

crates/executors/src/executors/codex.rs
abcpro1: It is cleaner breaking code following the "Separation of Concerns" and "Single Responsibility" principles.

crates/executors/src/executors/codex.rs
abcpro1: `msg` is of type `CodexMsgContent` not `Value`.

abcpro1: method has been inlined

crates/executors/src/executors/codex.rs
stunningpixels: If there's no technical reason, let's just use a simple match - much more readable

PR #388, opened 2025-08-04
refactor: introduce `AgentProfiles`

ggordonhall:
- Introduces `AgentProfiles` which contains custom user-defined agent configuration
- Automatically trigger a `CodingAgent` `ExecutionProcess` after a `SetupScript` has finished executing. We store an optional `next_action` field on `ScriptRequest` which stores the configuration of the agent action
- Centralise command building logic for coding agents in a `CommandBuilder` struct


crates/executors/src/actions/script.rs
```diff
@@ -29,6 +32,8 @@ pub struct ScriptRequest {
     pub script: String,
     pub language: ScriptRequestLanguage,
     pub context: ScriptContext,
+    /// Action to execute once the script has completed
+    pub next_action: Option<Box<ExecutorActions>>,
```
stunningpixels: We could move next_action up a level:

```
pub enum ExecutorActions {
    CodingAgentInitialRequest,
    CodingAgentFollowUpRequest,
    ScriptRequest,
}
```

becomes

```
pub enum ExecutorActionTypes {
    CodingAgentInitialRequest,
    CodingAgentFollowUpRequest,
    ScriptRequest,
}

pub struct ExecutorActions {
    type: ExecutorActionTypes,
    next_action: Option<Box<ExecutorActions>>
}
```

If the next action is outside of the inner structs (ScriptRequest etc...) might make it easier to chain multiple in a row.

crates/db/src/models/task_attempt.rs
LSRCT, line 1: Should we rename `executor_action_type` to `coding_agent_executor` here?
The values are variants of `CodingAgentExecutors`, not `ExecutorActionType`.

In the other table `execution_processes.executor_action_type` has variants of `ExecutorActionType`, so if we leave it it's ab it confusing

crates/server/src/routes/task_attempts.rs
```diff
@@ -258,10 +257,19 @@ pub async fn create_task_attempt(
         .executor
         .unwrap_or(deployment.config().read().await.executor.to_string());
 
+    let executor_action_type = executors::command::AgentProfiles::get_cached()
```
LSRCT: This wouldn't work with the current config, the `executor` there is a `CodingAgentExecutors`.tostring(), e.g. `CLAUDE_CODE`. I guess the solution is to have the config safe the default profile label instead?

PR #389, opened 2025-08-05
Refactor remaining task routes

LSRCT:
- Re-add tasks/create-and-start route
- Re-add delete task
- Remove executor field from TaskAttempt model


crates/server/src/routes/tasks.rs
```diff
@@ -74,104 +75,85 @@ pub async fn create_task(
     Ok(ResponseJson(ApiResponse::success(task)))
 }
 
-// TODO: create and start
```
stunningpixels, line 77: delete commented stuff

LSRCT: Agree, but the comment you metnion was deleted and a bunch after, too. What exactly do you mean?

crates/local-deployment/src/container.rs
```diff
@@ -426,6 +427,50 @@ impl ContainerService for LocalContainerService {
         Ok(worktree_path.to_string_lossy().to_string())
     }
 
+    async fn delete(&self, task_attempt: &TaskAttempt) -> Result<(), ContainerError> {
+        // stop all execution processes for this attempt
+        if let Ok(processes) =
+            ExecutionProcess::find_by_task_attempt_id(&self.db().pool, task_attempt.id).await
+        {
+            for process in processes {
+                self.stop_execution(&process).await.unwrap_or_else(|e| {
```
stunningpixels: Filter to running processes

PR #403, opened 2025-08-06
refactor: commit changes after successful execution

ggordonhall:
Try to commit changes after successful `CodingAgent` or `CleanupScript` execution

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     io,
-    path::PathBuf,
+    path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
 };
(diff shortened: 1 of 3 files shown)


crates/local-deployment/src/container.rs
```diff
@@ -701,4 +708,71 @@ impl ContainerService for LocalContainerService {
         let combined_stream = select(initial_stream, live_stream);
         Ok(combined_stream.boxed())
     }
+
+    async fn try_commit_changes(&self, ctx: &ExecutionContext) -> Result<(), ContainerError> {
+        if !matches!(
```
LSRCT, line 713: Is there a reason the SetupScript and the CleanupScript are treated differently here?

ggordonhall: We could commit on `SetupScript`, but it is less clear what the use case is. This mirrors the previous behaviour.

stunningpixels (approved):

PR #404, opened 2025-08-06
Alex/fix GitHub auth frontend

LSRCT:


crates/server/src/routes/task_attempts.rs
```diff
@@ -469,6 +471,47 @@ pub async fn merge_task_attempt(
     Ok(ResponseJson(ApiResponse::success(())))
 }
 
+#[derive(strum_macros::Display, Debug, Deserialize, TS)]
+#[strum(serialize_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
```
stunningpixels: SCREAMING_SNAKE_CASE to be consistent with all other shared enums

crates/server/src/routes/task_attempts.rs
```diff
@@ -469,18 +471,66 @@ pub async fn merge_task_attempt(
     Ok(ResponseJson(ApiResponse::success(())))
 }
 
+#[derive(Debug, Serialize, Deserialize, TS)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+#[ts(export)]
+#[ts(use_ts_enum)]
+pub enum CreateGitHubPRErrorData {
```
ggordonhall:
```suggestion
pub enum CreateGitHubPrErrorData {
```

crates/utils/src/response.rs
```diff
@@ -26,6 +28,16 @@ mod response {
                 success: false,
                 data: None,
                 message: Some(message.to_string()),
+                error_data: None,
+            }
+        }
+        /// Creates an error response, with `message` and no data.
```
ggordonhall:
```suggestion
        /// Creates an error response, with no `data`, no `message`, but with arbitrary `error_data`.
```

crates/server/src/routes/task_attempts.rs
```diff
@@ -469,18 +471,66 @@ pub async fn merge_task_attempt(
...
+    GithubTokenInvalid,
+    GithubRepoNotFoundOrNoAccess,
+    InsufficientGithubPermissions,
+}
+
+impl CreateGitHubPRErrorData {
```
ggordonhall: Could we remove this struct and add new error variants to `GitHubServiceError`?

```
impl GitHubServiceError {
    from git_error(error: &GitServiceError) {
        ...
    }
}
```

PR #405, opened 2025-08-06
Config upgrades

stunningpixels:
- Migrates from v1 (pre deployments) to v2 config (post deployments).
- Outlines a system that can be used for future config migrations.
- Virtually all config options needed mapping from old -> new due to schema changes or adoption of SCREAMING_SNAKE_CASE as the standard for constants.


crates/services/src/services/config/versions/v2.rs
```diff
@@ -28,33 +21,89 @@ pub struct Config {
...
+            tracing::warn!("Config reset to default");
+            Self::default()
+        }
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            config_schema: "v2".to_string(),
```
ggordonhall:
```suggestion
            config_version: "v2".to_string(),
```

PR #408, opened 2025-08-06
refactor: conditional notification

ggordonhall:
Only play sound notifications when a cleanup script has finished, or a coding agent with no cleanup script has finished.


stunningpixels (approved):
lgtm

Noticed:

https://github.com/BloopAI/vibe-kanban/blob/d8e45550051c3e5f1ba46c9a9a8afab634b4a204/crates/local-deployment/src/container.rs#L338C1-L343C30

I think this log message will fire in case where there is no next action, atm after a cleanup script.

PR #410, opened 2025-08-07
Claude Code Router

abcpro1:


crates/executors/src/command.rs
```diff
@@ -239,7 +255,7 @@ mod tests {
     #[test]
     fn test_default_profiles() {
         let profiles = AgentProfiles::from_defaults();
-        assert!(profiles.profiles.len() == 6);
+        assert!(profiles.profiles.len() == 7);
```
ggordonhall, line 258: I think we can just remove these tests, I don't think they add much value.

ggordonhall (approved):

PR #416, opened 2025-08-07
Fix cards not updating (vibe-kanban)

LSRCT:
Currently, the cards in vk do not change when an attempt is started,   but they should move to the "in progress column"

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -95,23 +95,13 @@ impl LocalContainerService {
         map.remove(id);
     }
 
-    /// Notifications are sent when:
-    /// - A CodingAgent completes without a next_action
-    /// - A CleanupScript completes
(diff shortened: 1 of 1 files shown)


crates/local-deployment/src/container.rs
```diff
@@ -343,6 +344,21 @@ impl LocalContainerService {
                             let notify_cfg = config.read().await.notifications.clone();
                             NotificationService::notify_execution_halted(notify_cfg, &ctx).await;
                         }
+                        // After the last action is completed, update task status to InReview
```
ggordonhall: We could move this into the `Self::should_notify` block. Is true when a `CodingAgent` has no next action or when a `CleanupScript` completes.

LSRCT: Whats the reasoning behind that condition? Is there a case where checking if a next action exists isn't enough?

PR #421, opened 2025-08-07
Re-add stop task attempt endpoint

LSRCT:
- Cleanup api.ts
- Re-add /api/task-attempts/{id}/stop endpoint


ggordonhall (commented):
We should also stop all `next_action` from executing when an attempt is stopped

PR #430, opened 2025-08-08
Qwen code

abcpro1:


crates/executors/src/command.rs
```diff
@@ -132,6 +139,16 @@ impl AgentProfile {
                 "--dangerously-bypass-approvals-and-sandbox",
                 "--skip-git-repo-check",
             ]),
+            mcp_config_path: None,
+        }
+    }
+
+    pub fn qwen() -> Self {
+        Self {
+            label: "qwen".to_string(),
```
ggordonhall:
```suggestion
            label: "qwen-code".to_string(),
```
Clearer if we refer to this as `qwen-code` everywhere.

crates/executors/src/command.rs
```diff
@@ -132,6 +139,16 @@ impl AgentProfile {
                 "--dangerously-bypass-approvals-and-sandbox",
                 "--skip-git-repo-check",
             ]),
+            mcp_config_path: None,
+        }
+    }
+
+    pub fn qwen() -> Self {
```
ggordonhall:
```suggestion
    pub fn qwen_code() -> Self {
```

crates/executors/src/command.rs
```diff
@@ -62,6 +62,8 @@ pub struct AgentProfile {
     pub agent: BaseCodingAgent,
     /// Command builder configuration
     pub command: CommandBuilder,
+    /// Optional profile-specific MCP config file path (absolute; supports leading ~ only)
```
ggordonhall:
```suggestion
    /// Optional profile-specific MCP config file path (absolute; supports leading ~ only). Overrides the default `BaseCodingAgent` config path
```

PR #441, opened 2025-08-08
fix: display profile in the task attempt card

ggordonhall:
- Display `profile_label` in the task attempt view rather than base coding agent
- Store `profile_label` as an additional column on `TaskAttempt` in the DB

### DIFF
--- a/crates/db/src/models/task_attempt.rs
+++ b/crates/db/src/models/task_attempt.rs
@@ -77,11 +77,12 @@ pub struct TaskAttempt {
     pub merge_commit: Option<String>,
     pub base_coding_agent: String, // Name of the base coding agent to use ("AMP", "CLAUDE_CODE",
     // "GEMINI", etc.)
-    pub pr_url: Option<String>,                    // GitHub PR URL
-    pub pr_number: Option<i64>,                    // GitHub PR number
-    pub pr_status: Option<String>,                 // open, closed, merged
-    pub pr_merged_at: Option<DateTime<Utc>>,       // When PR was merged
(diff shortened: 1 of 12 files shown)


crates/db/src/models/task_attempt.rs
```diff
@@ -77,11 +77,12 @@ pub struct TaskAttempt {
     pub merge_commit: Option<String>,
     pub base_coding_agent: String, // Name of the base coding agent to use ("AMP", "CLAUDE_CODE",
     // "GEMINI", etc.)
-    pub pr_url: Option<String>,                    // GitHub PR URL
-    pub pr_number: Option<i64>,                    // GitHub PR number
-    pub pr_status: Option<String>,                 // open, closed, merged
-    pub pr_merged_at: Option<DateTime<Utc>>,       // When PR was merged
-    pub worktree_deleted: bool, // Flag indicating if worktree has been cleaned up
+    pub profile_label: String,     // Profile label used for this attempt
```
LSRCT, line 80: The profile label may not be constant for a task attempt (claude-code-plan -> claude-code). Can we get this from the execution process instead for the frontend?
If not, we should rename it to `latest_profile_label` or `initial_profile_label`, depending on what it is.

LSRCT: See also #442

ggordonhall: Will address

ggordonhall:
Closing in favour of https://github.com/BloopAI/vibe-kanban/pull/442

PR #442, opened 2025-08-08
feat: add profile field to follow-up attempt

LSRCT:
- enable switching plan mode / normal mode: add profile field for follow up attempts
- fix no new execution_session being created for followups


crates/executors/src/actions/coding_agent_follow_up.rs
```diff
@@ -7,20 +7,21 @@ use ts_rs::TS;
...
+    command::ProfileVariant,
     executors::{CodingAgent, ExecutorError, StandardCodingAgentExecutor},
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS)]
 pub struct CodingAgentFollowUpRequest {
     pub prompt: String,
     pub session_id: String,
-    pub profile: String,
+    pub profile: ProfileVariant,
```
ggordonhall: Surprised that this takes a `ProfileVariant` and not just `Profile`. Is a `Profile` a `ProfileVariant` too (e.g. if you follow `claude-code` with `claude-code`)?

crates/executors/src/command.rs
```diff
@@ -53,6 +52,17 @@ impl CommandBuilder {
...
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS)]
+pub struct AgentVariantProfile {
+    /// Unique identifier for this profile (e.g., "MyClaudeCode", "FastAmp")
+    pub label: String,
+    /// The coding agent this profile is associated with
+    #[serde(flatten)]
+    pub agent: CodingAgent,
+    /// Optional profile-specific MCP config file path (absolute; supports leading ~). Overrides the default `BaseCodingAgent` config path
+    pub mcp_config_path: Option<String>,
```
ggordonhall, line 72: Could just make `variants` optional and do away with the duplicate `Variant` struct? Do we expect long-term divergence here?

crates/executors/src/command.rs
```diff
@@ -62,6 +72,42 @@ pub struct AgentProfile {
...
+    pub fn get_mcp_config_path(&self) -> Option<PathBuf> {
+        match self.mcp_config_path.as_ref() {
+            Some(path) => Some(PathBuf::from(path)),
+            None => self.agent.default_mcp_config_path(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS)]
+pub struct ProfileVariant {
```
ggordonhall: Does this uniquely identify a `Profile` and `Variant`? If so could just rename this:

```suggestion
pub struct ProfileVariantLabel {
```

crates/executors/src/executors/mod.rs
```diff
@@ -51,91 +57,104 @@ fn unknown_executor_error(s: &str) -> ExecutorError {
...
-            .map(|variant| variant.to_string())
-            .collect()
-    }
-
-    pub fn iter_discriminants() -> impl Iterator<Item = BaseCodingAgent> {
-        BaseCodingAgent::iter()
-    }
-
-    /// Create an executor from a profile string
+    /// Create an executor from a profile variant
```
ggordonhall:
```suggestion
    /// Create an CodingAgent from a profile variant
```

crates/services/src/services/container.rs
```diff
@@ -413,7 +414,7 @@ pub trait ContainerService {
     async fn start_attempt(
         &self,
         task_attempt: &TaskAttempt,
-        profile_label: String,
+        profile_id: ProfileVariant,
```
ggordonhall: Would like to have consistent naming for these

crates/executors/src/actions/coding_agent_follow_up.rs
LSRCT: To uniquely identify an agent we need a profile label and a variant label:
```
pub struct ProfileVariant {
    pub profile: String,
    pub variant: Option<String>,
}
```

For a normal claude code execution this would be
```
ProfileVariant {
   profile: "claude-code",
    variant: None
}
```

For a claude code plan execution this would be
```
ProfileVariant {
   profile: "claude-code",
    variant: Some("plan")
}
```

For the follow up specifically we only need `ProfileVariant.variant` and can get the `ProfileVariant.profile` from the initial request and we do that, just happens earlier in `routes/task_atttempts.rs` `follow_up`.

crates/executors/src/command.rs
LSRCT: Removing the duplicate in favour of a wrapper

PR #469, opened 2025-08-14
Display file edit content in log view

abcpro1:


frontend/src/components/NormalizedConversation/EditDiffRenderer.tsx
```diff
@@ -0,0 +1,134 @@
...
+    hideLineNumbers: hideNums,
+    totalHunks,
+    additions,
+    deletions,
+  };
+}
+
+function EditDiffRenderer({ path, hunks: hunksProp, hasLineNumbers }: Props) {
+  const { config } = useConfig();
+  const [expanded, setExpanded] = useState(false);
```
stunningpixels, line 61: Can we default to expanded: true

frontend/src/components/NormalizedConversation/EditDiffRenderer.tsx
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Process hunks for @git-diff-view/react
+ * - Extract additions/deletions for display
+ * - Decide whether to hide line numbers based on backend data
+ */
+function processHunks(path: string, hunks: string[], hasLineNumbers: boolean) {
+  const totalHunks = hunks.length;
+
+  // Prepend file headers, which are required by the diff view
+  const header = `--- a/${path}\n+++ b/${path}\n`;
```
stunningpixels: This could be done on the backend?

crates/executors/src/logs/mod.rs
```diff
@@ -49,12 +41,47 @@ pub struct NormalizedEntry {
...
+    Other {
+        description: String,
+    },
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+#[serde(tag = "action", rename_all = "snake_case")]
+pub enum FileChange {
+    /// Create a file if it doesn't exist, and overwrite its content.
+    Write { content: String },
```
stunningpixels, line 89: Rename this 'create'

stunningpixels (changes requested):
Looks good, couple minor points

frontend/src/components/NormalizedConversation/EditDiffRenderer.tsx
abcpro1: The purpose of displaying edit changes is to inspect the agent's actions. Edit changes take too much space in the log and do not accurately match the current state of the repo. There is a dedicated diff tab for up-to-date expanded diffs. I want the default state of the log view to be an overview of the conversation, with optional expandable details.

crates/executors/src/logs/mod.rs
abcpro1: I thought about that initially, but "create" isn't accurate. "Write" overwrites an existing file, or creates a new one, except we don't get the previous value.

frontend/src/components/NormalizedConversation/EditDiffRenderer.tsx
abcpro1: I'm looking into it

abcpro1: done

PR #478, opened 2025-08-14
Variant cycle keyboard shortcut (vibe-kanban)

LSRCT:
Add a keyboard shortcut to cycle through the different profile variants for follow ups. Make the shortcut lshift+tab


stunningpixels (commented):
I haven't tested but looks good - fyi there is frontend/src/lib/keyboard-shortcuts.ts which  tries to centralise all keyboard shortcuts to deal with conflicting ones

LSRCT:
> I haven't tested but looks good - fyi there is frontend/src/lib/keyboard-shortcuts.ts which tries to centralise all keyboard shortcuts to deal with conflicting ones

Good point, claude moved it over.

PR #507, opened 2025-08-18
Prepend and append prompt arguments to executors (vibe-kanban)

stunningpixels:
- You can override the prompt of each executor

CC does have a `--append-system-prompt <prompt>` CLI option, but this is different from modifying the prompt


crates/executors/src/executors/claude.rs
stunningpixels, line 1: Look into CLI route for this

Issue #333, opened 2025-07-24
ESC key closes input window and discards all content


mokizzz:
When creating a new task or interacting with AI, pressing the ESC key closes the input window and discards everything I’ve typed. This is especially problematic when using a Chinese input method, since I often press ESC to cancel in-progress compositions—causing the window to unexpectedly close and lose all input.

PR #508, opened 2025-08-18
Frontend: Fix esc/click discarding existing input (vibe-kanban)

LSRCT:
When creating a task, pressing escape or clicking outside the dialog will close it, discarding all of the content. When the title or description have content, there should be a warning about this instead.


LSRCT:
Should fix 
https://github.com/BloopAI/vibe-kanban/issues/333

PR #510, opened 2025-08-18
Allow multiple merges

LSRCT:
- Enable follow-ups on merged tasks
- Enable multiple merges
- Add merges table
- Show merged diff if worktree is clean and branch is not ahead
- Show error when trying to merge into a dirty base branch

### DIFF
--- a/frontend/src/components/tasks/Toolbar/CurrentAttempt.tsx
+++ b/frontend/src/components/tasks/Toolbar/CurrentAttempt.tsx
@@ -700,62 +700,57 @@ function CurrentAttempt({
                   {rebasing ? 'Rebasing...' : `Rebase`}
                 </Button>
               )}
-              {
-                // Normal merge and PR buttons for regular tasks
-                ((branchStatus.commits_ahead ?? 0) > 0 ||
-                  pushSuccess ||
-                  mergeSuccess) && (
-                  <>
-                    <Button
-                      onClick={handlePRButtonClick}
-                      disabled={
-                        creatingPR ||
-                        pushing ||
-                        Boolean((branchStatus.commits_behind ?? 0) > 0) ||
-                        isAttemptRunning ||
-                        (mergeInfo.hasOpenPR &&
-                          branchStatus.remote_commits_ahead === 0)
-                      }
-                      variant="outline"
-                      size="xs"
-                      className="border-blue-300 text-blue-700 hover:bg-blue-50 gap-1 min-w-[120px]"
-                    >
-                      <GitPullRequest className="h-3 w-3" />
-                      {mergeInfo.hasOpenPR
-                        ? pushSuccess
-                          ? 'Pushed!'
-                          : pushing
-                            ? 'Pushing...'
-                            : branchStatus.remote_commits_ahead === 0
-                              ? 'Push to PR'
-                              : branchStatus.remote_commits_ahead === 1
-                                ? 'Push 1 commit'
-                                : `Push ${branchStatus.remote_commits_ahead || 0} commits`
-                        : creatingPR
-                          ? 'Creating...'
-                          : 'Create PR'}
-                    </Button>
-                    <Button
-                      onClick={handleMergeClick}
-                      disabled={
-                        mergeInfo.hasOpenPR ||
-                        merging ||
-                        Boolean((branchStatus.commits_behind ?? 0) > 0) ||
-                        isAttemptRunning
-                      }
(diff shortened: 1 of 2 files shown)


crates/server/src/routes/task_attempts.rs
```diff
@@ -221,6 +221,11 @@ pub async fn follow_up(
         cleanup_action,
     );
 
+    // Clear merge_commit if it exists - follow-ups after merge should allow re-merging
+    if task_attempt.merge_commit.is_some() {
```
stunningpixels: Some scenarios not covered that I have experienced:
- Making further changes manually
- Starting a new task attempt, using another task attempt as the base and then merging

LSRCT: > * Making further changes manually

Do you mean manual changes to the task attempt branch after merging into a base branch? Why not just make them in the base branch?

> * Starting a new task attempt, using another task attempt as the base and then merging

Is this different from a normal attempt -> follow up -> merge?

LSRCT:
> Conceptually we currently we use 'merged' as the final state of a task attempt. Perhaps we should introduce a new concept of 'archived' which can be triggered manually after a developer finishes working on a task attempt.
> 
> Merges become just something that happens during the lifecycle of the task attempt, rather than the final state.

I don't know about a `final` state, is there a reason to ever lock a task? I'd like to follow-up whenever.
We could remove the final state logic (like disabling the merge button) entirely IMO.

crates/db/src/models/merge.rs
```diff
@@ -0,0 +1,284 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, Type)]
+#[sqlx(type_name = "TEXT", rename_all = "snake_case")]
+pub enum MergeType {
+    Direct,
+    Pr,
+}
+
+#[derive(FromRow)]
+struct MergeRow {
```
stunningpixels, line 60: Should we track the branch that the changes were merged into?

frontend/src/components/tasks/Toolbar/CurrentAttempt.tsx
```diff
@@ -456,21 +459,70 @@ function CurrentAttempt({
             Status
           </div>
           <div className="flex items-center gap-1.5">
-            {selectedAttempt.merge_commit ? (
-              <div className="flex items-center gap-1.5 overflow-hidden">
+            {mergeInfo.hasMergedPR &&
+            mergeInfo.mergedPR &&
+            mergeInfo.mergedPR.type === 'pr' ? (
+              <div className="flex items-center gap-1.5">
```
stunningpixels: This section could be restructured to be clearer, you could have an array of status text and dot colour.

`<div className="flex items-center gap-1.5">` is duplicated a lot.

frontend/src/components/tasks/Toolbar/CurrentAttempt.tsx
```diff
@@ -456,21 +459,70 @@ function CurrentAttempt({
...
+                    ? '1 commit ahead'
+                    : `${branchStatus?.commits_ahead} commits ahead`}
+                  {branchStatus?.has_uncommitted_changes && ' (dirty)'}
+                </span>
+              </div>
+            ) : (
+              <div className="flex items-center gap-1.5">
+                <div className="h-2 w-2 bg-gray-500 rounded-full" />
+                <span className="text-sm font-medium text-gray-700">
+                  Up to date
```
stunningpixels: Currently there is nothing in the UI that indicates to the user that the changes have been merged, perhaps when the task branch isn't ahead of the base, we could show status 'merged' ?

frontend/src/components/context/TaskDetailsContextProvider.tsx
```diff
@@ -176,6 +187,42 @@ const TaskDetailsProvider: FC<{
...
+    };
+
+    fetchBranchStatus();
+  }, [selectedAttempt]);
+
+  // Add periodic polling for branch status
+  useEffect(() => {
+    if (!selectedAttempt) return;
+
+    const interval = setInterval(async () => {
```
stunningpixels: Does this need to be on a loop? I'm trying to eliminate these as much as possible atm.

There should be triggers that we can hook into when we need to check.

LSRCT: Yes, needs to be in a loop to display PR status. Moved it to the existing task attempt loop.

PR #514, opened 2025-08-18
Re-rank file search (vibe-kanban)

stunningpixels:
Using the git history of a project, we should improve the ranking of search results when a user is searching for a file in a project.

We should iterate over each commit (up to 100), keeping track of:
- Which files were changes
- How many times each file was changed
- Which were changed most recently

This data should be calculated when needed and cached to improve performance.

You should then devise an algorithm to re-rank search results.

crates/server/src/routes/projects.rs
crates/services/src/services/git.rs

Please plan this using the oracle, confirm with me before making changes

### DIFF
--- a/crates/services/src/services/git.rs
+++ b/crates/services/src/services/git.rs
@@ -659,22 +659,22 @@ impl GitService {
     pub fn get_head_info(&self, repo_path: &Path) -> Result<HeadInfo, GitServiceError> {
         let repo = self.open_repo(repo_path)?;
         let head = repo.head()?;
-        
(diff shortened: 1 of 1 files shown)


stunningpixels:
Some before and afters:

Attempt before:

Screenshot 2025-08-18 at 17 08 18

Attempt after:

Screenshot 2025-08-18 at 17 08 23

Task before:

Screenshot 2025-08-18 at 17 08 30

Task after:

Screenshot 2025-08-18 at 17 08 41

PR #521, opened 2025-08-19
Add user message to log history

abcpro1:


crates/executors/src/executors/amp.rs
```diff
@@ -410,7 +410,7 @@ impl AmpContentItem {
         match self {
             AmpContentItem::Text { text } => {
                 let entry_type = match role {
-                    "user" => NormalizedEntryType::UserMessage,
```
stunningpixels, line 413: This breaks how amp removes old conversation turns from the latest turn

crates/services/src/services/container.rs
```diff
@@ -321,6 +321,16 @@ pub trait ContainerService {
                     if let Ok(executor) =
                         CodingAgent::from_profile_variant_label(&request.profile_variant_label)
                     {
+                        // Inject the initial user prompt before normalization (DB fallback path)
+                        let user_entry = NormalizedEntry {
```
stunningpixels, line 325: Feels like we are duplicating a lot of the same logic, maybe we should extract this to a single method?

crates/executors/src/logs/utils/entry_index.rs
```diff
@@ -28,6 +31,36 @@ impl EntryIndexProvider {
     pub fn reset(&self) {
         self.0.store(0, Ordering::Relaxed);
     }
+
+    /// Create a provider seeded from the maximum existing normalized-entry index
+    /// observed in prior JSON patches in `MsgStore`.
+    pub fn seeded_from_msg_store(msg_store: &MsgStore) -> Self {
```
stunningpixels, line 37: This feels a bit hacky, we could pass 'starts_from' as a param to the normalise function?

abcpro1: Why is it hacky? It just counts entries so we don't have to keep manual count up-to-date.

crates/executors/src/executors/amp.rs
abcpro1: Does it have a practical use-case? In my amp JSONL samples, the first message is always user message.

stunningpixels: If you make a follow up, the logs from previous follow ups are now duplicated because amp sends the entire conversation history each time you make a follow up

crates/executors/src/logs/utils/entry_index.rs
stunningpixels: I think 'starts from' would be more intuitive here

abcpro1: we can rename it to start_from but I prefer automatic count vs manually tracking how many message we have prepended.

crates/executors/src/executors/amp.rs
abcpro1: will look into it

PR #577, opened 2025-08-26
Overhaul UI

stunningpixels:


jtomaszewski:
Great work! I like it that one can see way more on the small screen now. 

Although I find the font a bit hard to read.

stunningpixels:
Cheers @jtomaszewski - will continue to refine over the next couple of days, but this is a start

PR #607, opened 2025-09-02
feature: Checkpoint restore

abcpro1:


crates/db/src/models/execution_process.rs
```diff
@@ -360,6 +373,28 @@ impl ExecutionProcess {
...
+    pub async fn set_restore_boundary(
+        pool: &SqlitePool,
+        task_attempt_id: Uuid,
+        boundary_process_id: Uuid,
+    ) -> Result<(), sqlx::Error> {
+        // Monotonic drop: only mark newer records as dropped; never undrop.
+        sqlx::query!(
+            r#"UPDATE execution_processes
+               SET dropped = 1
+             WHERE task_attempt_id = $1
```
stunningpixels, line 431: And not a dev server

stunningpixels:
I was thinking about feature flagging recently, and how we could do this in the most typesafe way possible - what do you think of defining capabilities as traits:

```
use async_trait::async_trait;
use enum_dispatch::enum_dispatch;
use serde::{Deserialize, Serialize};
use strum_macros::{Display, VariantNames};
use ts_rs::TS;

use crate::executors::{amp::Amp, claude::ClaudeCode, codex::Codex};

#[enum_dispatch]
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS, Display, VariantNames)]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]
pub enum CodingAgentRandomCapability {
    ClaudeCode,
    Amp,
    Codex,
}

impl CodingAgentRandomCapability {}

#[async_trait]
#[enum_dispatch(CodingAgentRandomCapability)]
pub trait RandomCapabilityTrait {
    async fn perform_random_capability(&self) {}
}

```

The list of which executors support the trait is fully exportable to the frontend via TS-RS:

```
export type CodingAgentRandomCapability = { "CLAUDE_CODE": ClaudeCode } | { "AMP": Amp } | { "CODEX": Codex };
```

(You can also make this available at runtime by exporting an enum)

crates/db/src/models/execution_process.rs
abcpro1: Dev servers, clean-up scripts, and processes in general are also obsoleted after worktree (git) restore. Tested that restore is unavailable when dev server is running, and dev server process is also marked as deleted if it had been executed in a previous repo state that no-longer exists.

abcpro1:
The challenge with the trait design is that not all capabilities are independent actions. E.g. restore process management is managed in vibe-kanban, so we only need a boolean from the executor.

stunningpixels:
Could those be default method implementations on the trait?

kloodylemonh-tech (commented):
Loll

PR #619, opened 2025-09-03
UI form builder for editing executor config

stunningpixels:
Screenshot 2025-09-04 at 00 12 15

### DIFF
--- a/crates/executors/src/executors/codex.rs
+++ b/crates/executors/src/executors/codex.rs
@@ -40,7 +40,7 @@ pub enum SandboxMode {
 }
 
 /// Approval policy for Codex
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS, AsRefStr)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS, AsRefStr, JsonSchema)]
 #[serde(rename_all = "kebab-case")]
 #[strum(serialize_all = "kebab-case")]
 pub enum ApprovalPolicy {
--- a/shared/schemas/codex.json
+++ b/shared/schemas/codex.json
@@ -24,6 +24,20 @@
         null
       ]
     },
+    "approval": {
+      "description": "Approval policy for Codex",
+      "type": [
+        "string",
+        "null"
+      ],
+      "enum": [
+        "untrusted",
+        "on-failure",
+        "on-request",
+        "never",
+        null
+      ]
+    },
     "oss": {
       "type": [
         "boolean",
--- a/shared/types.ts
+++ b/shared/types.ts
@@ -58,13 +58,11 @@ export type CreateImage = { file_path: string, original_name: string, mime_type:
 
 export type ApiResponse<T, E = T> = { success: boolean, data: T | null, error_data: E | null, message: string | null, };
 
-export type UserSystemInfo = {
-    config: Config, environment: Environment,
-    /**
-     * Capabilities supported per executor (e.g., { "CLAUDE_CODE": ["RESTORE_CHECKPOINT"] })
-     */
-    capabilities: { [key in string]?: Array<BaseAgentCapability> }, executors: { [key in BaseCodingAgent]?: ExecutorConfig },
-};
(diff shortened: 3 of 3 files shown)


frontend/src/App.tsx
```diff
@@ -237,8 +247,17 @@ function AppContent() {
                   path="/projects/:projectId/tasks/:taskId"
                   element={<ProjectTasks />}
                 />
-                <Route path="/settings" element={<Settings />} />
-                <Route path="/mcp-servers" element={<McpServers />} />
+                <Route path="/settings/*" element={<SettingsLayout />}>
+                  <Route index element={<Navigate to="general" replace />} />
+                  <Route path="general" element={<GeneralSettings />} />
+                  <Route path="executors" element={<ExecutorsSettings />} />
```
ggordonhall:
```suggestion
                  <Route path="agents" element={<ExecutorsSettings />} />
```

We should refer to these as `agents` rather than `executors`. Unclear to the average user what an executor is.

frontend/src/pages/settings/ExecutorsSettings.tsx
```diff
@@ -0,0 +1,666 @@
...
+        <CardContent className="space-y-4">
+          {/* Editor type toggle */}
+          <div className="flex items-center space-x-2">
+            <Checkbox
+              id="use-form-editor"
+              checked={useFormEditor}
+              onCheckedChange={(checked) => setUseFormEditor(!!checked)}
+              disabled={profilesLoading || !localParsedProfiles}
+            />
+            <Label htmlFor="use-form-editor">Edit visually</Label>
```
ggordonhall: This would be clearer if was unchecked by default and was labelled "Edit JSON" (e.g. the reverse toggle)

frontend/src/pages/settings/ExecutorsSettings.tsx
```diff
@@ -0,0 +1,666 @@
...
+  const validateConfigName = (name: string): string | null => {
+    const trimmedName = name.trim();
+    if (!trimmedName) return 'Configuration name cannot be empty';
+    if (trimmedName.length > 40)
+      return 'Configuration name must be 40 characters or less';
+    if (!/^[a-zA-Z0-9_-]+$/.test(trimmedName)) {
+      return 'Configuration name can only contain letters, numbers, underscores, and hyphens';
+    }
+    if (localParsedProfiles?.executors?.[selectedExecutorType]?.[trimmedName]) {
+      return 'A configuration with this name already exists';
```
ggordonhall: We should be consistent in our terminology. We refer to "configurations" here, but "profiles" elsewhere. Imo we should continue to refer to these as profiles here ("configure a profile").

frontend/src/pages/settings/SettingsLayout.tsx
```diff
@@ -0,0 +1,71 @@
...
+const settingsNavigation = [
+  {
+    path: 'general',
+    label: 'General',
+    icon: Settings,
+    description: 'Theme, notifications, and preferences',
+  },
+  {
+    path: 'executors',
+    label: 'Executors',
```
ggordonhall:
```suggestion
    label: 'Agents',
```

frontend/src/pages/settings/ExecutorsSettings.tsx
```diff
@@ -0,0 +1,666 @@
...
+
+          {useFormEditor &&
+          localParsedProfiles &&
+          localParsedProfiles.executors ? (
+            // Form-based editor
+            <div className="space-y-4">
+              <div className="grid grid-cols-2 gap-4">
+                <div className="space-y-2">
+                  <Label htmlFor="executor-type">Executor</Label>
+                  <Select
```
ggordonhall: `DEFAULT` should always be the first item in the drop down. `ROUTER` is currently the first item for `CLAUDE_CODE`.

ggordonhall (commented):
I like the JSON builder. It's now much easier to edit profiles. But now its also easier configure profiles that don't work, and to break default profiles. Could we have some sort of undo/restore?

PR #625, opened 2025-09-04
feat: rework log view

ggordonhall:


frontend/src/components/tasks/TaskDetails/LogsTab.tsx
```diff
@@ -196,7 +196,12 @@ function LogsTab({ selectedAttempt }: Props) {
     if (toExpand.length > 0) {
       dispatch({ type: 'AUTO_EXPAND', ids: toExpand });
     }
-  }, [filteredProcesses, state.userCollapsed, state.autoCollapsed]);
+  }, [
+    filteredProcesses,
+    state.userCollapsed,
```
stunningpixels: This component is infinitely re-rendering, eventually crashing chrome.

Somewhere in here the state is being modified and also listened to, which causes an infinite loop.

PR #639, opened 2025-09-05
Add spinoff button (vibe-kanban)

LSRCT:
Add a button for task attempts that creates a new task with the same base branch


frontend/src/components/tasks/Toolbar/CurrentAttempt.tsx
stunningpixels, line 1: Rebase thing need to import GitFork

PR #652, opened 2025-09-08
Remove ccr (vibe-kanban)

LSRCT:
Remove the claude code router from the default profiles

### DIFF
--- a/crates/executors/default_profiles.json
+++ b/crates/executors/default_profiles.json
@@ -10,12 +10,6 @@
         "CLAUDE_CODE": {
           "plan": true
         }
-      },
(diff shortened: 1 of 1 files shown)


tomzx:
Is CCR still supported or is support dropped?
If it's been dropped, is there a reason?
Also https://www.vibekanban.com/docs/supported-coding-agents/ccr should probably be updated if support is dropped.

abcpro1:
CCR is still supported. Docs will be updated shortly.

abcpro1:
https://www.vibekanban.com/docs/supported-coding-agents/ccr

PR #655, opened 2025-09-08
Collect richer logs from Opencode

abcpro1:


crates/executors/src/executors/opencode_share_bridge.rs
```diff
@@ -0,0 +1,196 @@
+use std::{collections::HashMap, net::SocketAddr, sync::Arc};
```
ggordonhall: Can you stick this in an `opencode/` directory

crates/executors/Cargo.toml
```diff
@@ -40,3 +40,5 @@ strum = "0.27.2"
 strum_macros = "0.27.2"
 convert_case = "0.6"
 sqlx = "0.8.6"
+axum = { workspace = true }
+once_cell = "1.19"
```
ggordonhall: I don't think this is used anywhere

abcpro1: `axum` is used in `opencode_share_bridge.rs`. `once_cell` is no longer used

PR #657, opened 2025-09-08
feat: one click installation for popular MCP servers

ggordonhall:
Adds support for one click installation for popular MCP servers on the settings page. Implement adapters to handle divergent MCP formats.

### DIFF
--- a/crates/executors/src/mcp_config.rs
+++ b/crates/executors/src/mcp_config.rs
@@ -184,36 +184,69 @@ fn adapt_codex(mut servers: ServerMap, mut meta: Option<Value>) -> Value {
     attach_meta(servers, meta)
 }
 
-fn adapt_opencode() -> Value {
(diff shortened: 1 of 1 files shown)


crates/executors/src/mcp_config.rs
```diff
@@ -79,3 +84,180 @@ pub async fn write_agent_config(
...
+
+    if let Some(Value::Object(ref mut m)) = meta {
+        m.retain(|k, _| servers.contains_key(k));
+        servers.insert("meta".to_string(), Value::Object(std::mem::take(m)));
+        meta = None; // already attached above
+    }
+    attach_meta(servers, meta)
+}
+
+fn adapt_opencode() -> Value {
```
LSRCT: Is there a way to use the same pattern as with the others here? Looks like the .json is ignored for opencode.

PR #658, opened 2025-09-09
Codex follow-ups fail with “Couldn't find a prior session_id”

erukiti:
Overview

- After a Codex run, attempting “Continue working on this task attempt...” fails to start a follow-up.
- Error: Failed to start follow-up execution: TaskAttemptError: Validation error: Couldn't find a prior session_id, please create a new task attempt
- This does not occur with Claude Code.

Affected Versions

- Codex CLI @openai/codex 0.30.0 and 0.31.0

Steps To Reproduce

- Set executor to Codex.
- Run an initial execution on a task and let it complete.
- In the same task, send a follow-up prompt (“Continue working on this task attempt...”).
- Observe the error above.

Expected Behavior

- The follow-up reuses the prior session and starts normally (new ExecutionProcess created, logs stream).

Actual Behavior

- API can’t find the prior session_id, and the follow-up fails with a validation error.

Root Cause (Investigation)

- The follow-up API requires the last session_id persisted in DB (executor_sessions.session_id) and fetches it with find_latest_session_id_by_task_attempt.
- For Claude, the session_id appears in the JSON stdout and is reliably captured.
- For Codex:
      - The JSON stdout does not include a session_id.
      - stderr may include a “SessionConfigured … session_id: ” line, but on some versions/environments it’s absent or formatted differently.
      - As a result, no session_id is saved, and the follow-up fails.

  Temporary Fix (Implemented)

  - Added a fallback in Codex log normalization to scan ~/.codex/sessions/
  rollout-*.jsonl for the latest rollout after start time and extract the id
  (used as session_id).
  - If found, emit it via msg_store.push_session_id(...) so it’s persisted to DB.
  - Files:
      - crates/executors/src/executors/codex.rs (added
  SessionHandler::start_rollout_scan_fallback and invoked it in normalize_logs)

### DIFF
--- a/crates/executors/src/executors/codex.rs
+++ b/crates/executors/src/executors/codex.rs
@@ -1,4 +1,6 @@
 use std::{
+    fs,
+    io::{BufRead, BufReader},
     path::{Path, PathBuf},
     process::Stdio,
     sync::Arc,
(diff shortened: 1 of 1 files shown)


abcpro1:
Resolved in #700

PR #660, opened 2025-09-09
fix: stop showing diff spinner after timeout (vibe-kanban)

ggordonhall:
This should just be a frontend change. Look at the frontend/src/components/tasks/TaskDetails/DiffTab.tsx

### DIFF
--- a/frontend/src/components/tasks/TaskDetails/DiffTab.tsx
+++ b/frontend/src/components/tasks/TaskDetails/DiffTab.tsx
@@ -20,10 +20,29 @@ function DiffTab({ selectedAttempt }: DiffTabProps) {
   );
 
   useEffect(() => {
+    setLoading(true);
+  }, [selectedAttempt?.id]);
+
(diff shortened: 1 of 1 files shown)


frontend/src/components/tasks/TaskDetails/DiffTab.tsx
```diff
@@ -18,12 +18,27 @@ function DiffTab({ selectedAttempt }: DiffTabProps) {
...
+    setLoading(true);
+  }, [selectedAttempt?.id]);
+
   useEffect(() => {
     if (diffs.length > 0 && loading) {
       setLoading(false);
     }
   }, [diffs, loading]);
 
+  // If no diffs arrive within 3 seconds, stop showing the spinner
```
LSRCT: On my machine diffs regularly take more than 3 seconds to show up (unfortunately). Not sure what a good value is, probably >5.

LSRCT (approved):

PR #664, opened 2025-09-09
feat: support Codex `--full-auto` mode

ggordonhall:
Codex has a `--full-auto` flag that allows it to read files, make edits, and run commands from within a sandbox (using Seatbelt on MacOS or Landlock on Linux). Approval is required to make web requests. https://github.com/openai/codex/blob/main/docs/sandbox.md

This is a safer default than `--dangerously-bypass-approvals-and-sandbox` which bypasses the sandbox completely.

### DIFF
--- a/crates/executors/default_profiles.json
+++ b/crates/executors/default_profiles.json
@@ -41,7 +41,7 @@
       },
       "HIGH": {
         "CODEX": {
-          "sandbox": "danger-full-access",
+          "sandbox": "auto",
           "model_reasoning_effort": "high"
         }
       }


crates/executors/default_profiles.json
```diff
@@ -36,7 +36,7 @@
     "CODEX": {
       "DEFAULT": {
         "CODEX": {
-          "sandbox": "danger-full-access"
+          "sandbox": "auto"
         }
       },
       "HIGH": {
```
abcpro1, line 42: please change the `HIGH` variant as well.

PR #666, opened 2025-09-09
Improve repo discovery performance (vibe-kanban)

LSRCT:
Make sure we
- Search only the working directory
- Filter common non-git folders like Movies, Music

### DIFF
--- a/crates/services/src/services/filesystem.rs
+++ b/crates/services/src/services/filesystem.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashSet,
     fs,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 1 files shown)


crates/services/src/services/filesystem.rs
```diff
@@ -52,13 +52,47 @@ impl FilesystemService {
     ) -> Result<Vec<DirectoryEntry>, FilesystemError> {
         let base_path = path
             .map(PathBuf::from)
-            .unwrap_or_else(Self::get_home_directory);
+            .unwrap_or_else(Self::get_working_or_home_directory);
         Self::verify_directory(&base_path)?;
         let mut git_repos: Vec<DirectoryEntry> = WalkBuilder::new(&base_path)
             .follow_links(false)
-            .hidden(true)
+            .hidden(false)
```
ggordonhall: We want to ignore hidden files, no?

crates/services/src/services/filesystem.rs
```diff
@@ -52,13 +52,47 @@ impl FilesystemService {
...
-            .filter_entry(|entry| entry.path().is_dir())
+            .filter_entry(|entry| {
+                let path = entry.path();
+                if !path.is_dir() {
+                    return false;
+                }
+
+                // Skip common non-git folders
+                if let Some(name) = path.file_name().and_then(|n| n.to_str())
+                    && matches!(
```
ggordonhall: The `dirs` crate has cross-platform values for these paths: https://codeberg.org/dirs/dirs-rs

crates/services/src/services/filesystem.rs
LSRCT: True, changed it back.

PR #686, opened 2025-09-11
Navigate to fullscreen task (vibe-kanban)

LSRCT:
Fullscreen state is now preserved at:
- New Attempt
- Create Subtask
- Navigate to child/parent task

Before this any of the above actions would leave fullscreen mode.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,14 +1,9 @@
 import { useEffect } from 'react';
-import {
-  BrowserRouter,
-  Navigate,
-  Route,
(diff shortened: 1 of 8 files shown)


frontend/src/hooks/useFullscreenState.ts
stunningpixels, line 1: If this is a hook, we may as well add some more stuff like `toggleFullScreen(bool)` that would replace the prop passing of setFullScreen

stunningpixels (approved):

PR #692, opened 2025-09-11
Improve performance of conversation

stunningpixels:


frontend/src/components/ui/markdown-renderer.tsx
```diff
@@ -100,7 +100,7 @@ function MarkdownRenderer({
   return (
     <div className={`relative group`}>
       {enableCopyButton && (
-        <div className="sticky top-2 z-10 pointer-events-none">
```
abcpro1, line 103: Please revert these, the button UX has degraded.

PR #695, opened 2025-09-11
Rebase conflict resolution

abcpro1:


shared/types.ts
```diff
@@ -218,7 +218,19 @@ export type CommitInfo = { sha: string, subject: string, };
...
-export type BranchStatus = { commits_behind: number | null, commits_ahead: number | null, has_uncommitted_changes: boolean | null, head_oid: string | null, uncommitted_count: number | null, untracked_count: number | null, base_branch_name: string, remote_commits_behind: number | null, remote_commits_ahead: number | null, merges: Array<Merge>, };
+export type BranchStatus = { commits_behind: number | null, commits_ahead: number | null, has_uncommitted_changes: boolean | null, head_oid: string | null, uncommitted_count: number | null, untracked_count: number | null, base_branch_name: string, remote_commits_behind: number | null, remote_commits_ahead: number | null, merges: Array<Merge>, 
+/**
+ * True if a `git rebase` is currently in progress in this worktree
+ */
+is_rebase_in_progress: boolean, 
+/**
+ * Current conflict operation if any: one of "rebase", "merge", "cherry_pick", "revert"
+ */
+conflict_op: string | null,
```
LSRCT: Should this be an enum?

crates/services/src/services/git_cli.rs
```diff
@@ -355,14 +355,64 @@ impl GitCli {
...
-        match self.git(worktree_path, ["rev-parse", "--verify", "REBASE_HEAD"]) {
+        let rebase_merge = self.git(worktree_path, ["rev-parse", "--git-path", "rebase-merge"])?;
+        let rebase_apply = self.git(worktree_path, ["rev-parse", "--git-path", "rebase-apply"])?;
+        let rm_exists = std::path::Path::new(rebase_merge.trim()).exists();
+        let ra_exists = std::path::Path::new(rebase_apply.trim()).exists();
+        Ok(rm_exists || ra_exists)
+    }
+
+    /// Return true if a merge is in progress (MERGE_HEAD exists).
+    pub fn is_merge_in_progress(&self, worktree_path: &Path) -> Result<bool, GitCliError> {
```
LSRCT, line 370: Should this work the same way as rebase?

shared/types.ts
abcpro1: Good point. will do

crates/services/src/services/git_cli.rs
abcpro1: There isn’t a reliable “directory” equivalent like rebase’s rebase-merge/rebase-apply.

LSRCT: Are we fine with the way the way `in_progress` functions relate to the `abort` functions?

Let's say `in_progress` is true for an op in a worktree and I use `abort` I'd expect `in_progress` to be false after. 
Currently this may not happen, we may get the same problems we had with rebase, leaving a state where we detect an ongoing merge, but abort fails due to git thinking there is no merge.

frontend/src/components/tasks/Toolbar/CurrentAttempt.tsx
```diff
@@ -220,7 +243,13 @@ function CurrentAttempt({
       await rebaseMutation.mutateAsync(undefined);
       setError(null); // Clear any previous errors on success
     } catch (err) {
-      setError(err instanceof Error ? err.message : 'Failed to rebase branch');
+      // Distinguish conflicts/in-progress from other errors: rely on banner for conflicts
+      const isConflict = err instanceof ApiError && err.statusCode === 409;
```
LSRCT: We could have proper error types and use `ApiResponse::error_with_data` for this. We already use this for GithubService errors

abcpro1: good suggestion

frontend/src/lib/api.ts
```diff
@@ -486,6 +487,16 @@ export const attemptsApi = {
         body: JSON.stringify(data),
       }
     );
+    return handleApiResponse<void, GitOperationError>(response);
```
LSRCT:
``` suggestion
    return handleApiResponseAsResult<void, GitOperationError>(response);
```
With this you can change the return type like in `createPR` and get type safety without casting it from any (`err.error_data as GitOperationError | undefined`)

frontend/src/hooks/useBranchStatus.ts
```diff
@@ -6,6 +6,7 @@ export function useBranchStatus(attemptId?: string) {
     queryKey: ['branchStatus', attemptId],
     queryFn: () => attemptsApi.getBranchStatus(attemptId!),
     enabled: !!attemptId,
-    refetchInterval: 5000,
+    // Poll faster to promptly reflect rebase/abort transitions
+    refetchInterval: 1500,
```
stunningpixels: I think this checks GitHub? Not sure this is a good idea

frontend/src/hooks/useRebase.ts
```diff
@@ -17,7 +18,17 @@ export function useRebase(
       const data: RebaseTaskAttemptRequest = {
         new_base_branch: newBaseBranch || null,
       };
-      return attemptsApi.rebase(attemptId, data);
+      return attemptsApi.rebase(attemptId, data).then((res) => {
+        if (!res.success) {
+          // Throw a typed ApiError so callers can branch on error_data
+          throw new ApiError<GitOperationError>(
```
stunningpixels: Manually throwing a backend error on the frontend?

frontend/src/components/tasks/ConflictBanner.tsx
```diff
@@ -0,0 +1,101 @@
...
+  conflictedFiles,
+  isDraftLocked,
+  isDraftReady,
+  onOpenEditor,
+  onInsertInstructions,
+  onAbort,
+  op,
+}: Props) {
+  const displayFiles = conflictedFiles.slice(0, 8);
+  const opTitle =
```
stunningpixels: Duplicate, also in CurrentAttempt

frontend/src/lib/conflicts.ts
```diff
@@ -0,0 +1,39 @@
...
+  op?: ConflictOp | null
+): string {
+  const source = sourceBranch || 'current attempt branch';
+  const base = baseBranch ?? 'base branch';
+  const filesList = conflictedFiles.slice(0, 12);
+  const filesBlock = filesList.length
+    ? `\n\nFiles with conflicts:\n${filesList.map((f) => `- ${f}`).join('\n')}`
+    : '';
+
+  const opTitle: string =
```
stunningpixels: Duplicate

frontend/src/hooks/useBranchStatus.ts
abcpro1: reverted

PR #700, opened 2025-09-12
Update codex to the latest version

abcpro1:


crates/executors/src/executors/codex.rs
```diff
@@ -246,7 +277,7 @@ pub struct Codex {
 
 impl Codex {
     fn build_command_builder(&self) -> CommandBuilder {
-        let mut builder = CommandBuilder::new("npx -y @openai/codex@0.29.0 exec")
+        let mut builder = CommandBuilder::new("npx -y @openai/codex exec")
```
ggordonhall: We set `@latest` for most executors. Think we should do here too.

PR #717, opened 2025-09-15
open the frontend by default when running the dev command

britannio:


stunningpixels:
We used to have this but it got annoying when Claude would run the dev command while it was making changes

britannio:
I see, so I should flip the default to `false` instead?

stunningpixels:
Yes that'd work. In future we should have dev mode human, dev mode AI

PR #722, opened 2025-09-15
Change SSE to websocket (vibe-kanban)

stunningpixels:
We need to change SSE to websocket, but we're going to essentially use websockets as if they were SSE, so there won't be any client transmissions.

Let's start with the `stream_tasks` endpoint.

crates/server/src/routes/tasks.rs


frontend/src/hooks/useJsonPatchWsStream.ts
```diff
@@ -0,0 +1,173 @@
...
+            setData(dataRef.current);
+          }
+
+          // Handle Finished messages (same as SSE finished event)
+          if (msg.Finished !== undefined) {
+            ws.close();
+            wsRef.current = null;
+            setIsConnected(false);
+            // Treat finished as terminal and schedule reconnect; servers may rotate
+            retryAttemptsRef.current += 1;
```
abcpro1, line 128: `retryAttemptsRef.current` needs an overhaul. It is incremented in a few places and doesn't guard against graceful close.

abcpro1 (approved):

PR #739, opened 2025-09-16
Raise log truncation limit to 100kb

stunningpixels:
Noticing some legit logs are being truncated


abcpro1:
Truncating the middle is safer than truncating the tail

PR #744, opened 2025-09-16
Refactor: TaskFollowUpSection.tsx

abcpro1:


frontend/src/components/tasks/TaskFollowUpSection.tsx
```diff
@@ -226,561 +181,73 @@ export function TaskFollowUpSection({
...
 
-  const {
-    data: draftStream,
-    isConnected: draftStreamConnected,
-    error: draftStreamError,
-  } = useJsonPatchStream<DraftStreamState>(
-    draftStreamEndpoint,
-    !!draftStreamEndpoint,
-    makeInitialDraftData
+  const appendToFollowUpMessage = useCallback(
```
stunningpixels, line 190: This feels convoluted, can the code be restructured to avoid these callbacks

frontend/src/components/tasks/follow-up/FollowUpEditorCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export function FollowUpEditorCard({
+  placeholder,
+  value,
+  onChange,
+  onKeyDown,
+  disabled,
+  projectId,
+  rows = 1,
```
stunningpixels, line 25: Unnecessary props as this component is only used in one place, these should be constants

frontend/src/hooks/follow-up/useDraftAutosave.ts
```diff
@@ -0,0 +1,114 @@
...
+import type { FollowUpDraft, ImageResponse } from 'shared/types';
+
+export type SaveStatus = 'idle' | 'saving' | 'saved' | 'offline' | 'sent';
+
+type Args = {
+  attemptId?: string;
+  draft: FollowUpDraft | null;
+  message: string;
+  selectedVariant: string | null;
+  images: ImageResponse[];
```
stunningpixels, line 12: Are these essentially the same as FollowUpDraft shared type? Can we reduce bloat by passing that around instead of each field individually

frontend/src/hooks/follow-up/useDraftImages.ts
stunningpixels, line 1: This can be folded into useDraftEdits?

frontend/src/components/tasks/TaskFollowUpSection.tsx
abcpro1: Addressed in #762

frontend/src/components/tasks/follow-up/FollowUpEditorCard.tsx
abcpro1: Addressed in #762

frontend/src/hooks/follow-up/useDraftAutosave.ts
abcpro1: Addressed in #762

frontend/src/hooks/follow-up/useDraftImages.ts
abcpro1: Addressed in #762

PR #748, opened 2025-09-17
feat: manual approvals

ggordonhall:


crates/executors/src/logs/utils/patch.rs
```diff
@@ -22,6 +25,9 @@ pub enum PatchType {
     Stdout(String),
     Stderr(String),
     Diff(Diff),
+    ApprovalRequest(ApprovalRequest),
```
stunningpixels: Could these be normalized entries? Perhaps a field on the sub entry `requires_approval` true/false, `timeout` and `approval_status` pending/approved/rejected

stunningpixels:
Question
- Do users need to leave a comment when rejecting. Observed when rejecting the model tries to look for hooks in my repo

Comments
- Rounded styles should be sharp
- Need to disable the loading placeholder when showing approve/reject
- Can't see approval status in the log history, all historic approvals look like they have been accepted even when not

frontend/src/lib/api.ts
```diff
@@ -795,3 +796,24 @@ export const imagesApi = {
     return `/api/images/${imageId}/file`;
   },
 };
+
+// Approval API
+export const approvalsApi = {
+  respond: async (
+    approvalId: string,
+    payload: {
```
stunningpixels: ApprovalResponse in shared types

crates/server/src/main.rs
```diff
@@ -81,9 +81,7 @@ async fn main() -> Result<(), VibeKanbanError> {
     let actual_port = listener.local_addr()?.port(); // get → 53427 (example)
 
     // Write port file for discovery if prod, warn on fail
-    if !cfg!(debug_assertions)
-        && let Err(e) = write_port_file(actual_port).await
-    {
+    if let Err(e) = write_port_file(actual_port).await {
```
abcpro1, line 84: is this intentional?

ggordonhall: Is there an issue with doing this in debug? We need to know the port for approvals to work at all.

abcpro1: this system is hacky. it doesn't support multiple vibe-kanban instances. this is the reason it was disabled for debug.

crates/executors/src/executors/claude.rs
```diff
@@ -177,22 +223,86 @@ impl StandardCodingAgentExecutor for ClaudeCode {
...
+                }
+            ]
+        }
+    })
+    .to_string())
+}
+
 fn create_watchkill_script(command: &str) -> String {
-    let claude_plan_stop_indicator = concat!("Exit ", "plan mode?");
+    let claude_plan_stop_indicator = "Approval request timed out";
```
LSRCT: concat!("Approval ", "request timed out"), otherwise plan mode stops when claude reads claude.rs

crates/db/src/models/executor_session.rs
```diff
@@ -104,6 +104,29 @@ impl ExecutorSession {
         .await
     }
 
+    pub async fn find_by_session_id(
```
LSRCT, line 107: Do we still need `find_by_id` above?

PR #755, opened 2025-09-17
Fix todos and delete useProcessesLogs (vibe-kanban)

stunningpixels:
Currently the TodoPanel uses useProcessesLogs, which is being retired.

Let's take a new approach, creating a new hook and shared context which means we can share state (entries) from the onEntriesUpdated callback in VirtualizedList to the Todo component.

We can't modify the path of entries between useConversationHistory and VirtualizedList because it's highly optimised.

frontend/src/components/tasks/TodoPanel.tsx
frontend/src/hooks/useProcessesLogs.ts
frontend/src/components/logs/VirtualizedList.tsx

### DIFF
--- a/frontend/src/components/tasks/TaskDetailsPanel.tsx
+++ b/frontend/src/components/tasks/TaskDetailsPanel.tsx
@@ -207,7 +207,6 @@ export function TaskDetailsPanel({
 
                               <TaskFollowUpSection
                                 task={task}
-                                projectId={projectId}
                                 selectedAttemptId={selectedAttempt?.id}
                                 jumpToLogsTab={jumpToLogsTab}
                               />
@@ -252,7 +251,6 @@ export function TaskDetailsPanel({
 
                             <TaskFollowUpSection
                               task={task}
-                              projectId={projectId}
                               selectedAttemptId={selectedAttempt?.id}
                               jumpToLogsTab={jumpToLogsTab}
                             />


frontend/src/contexts/EntriesContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+const EntriesContext = createContext<EntriesContextType | null>(null);
+
+interface EntriesProviderProps {
+  children: ReactNode;
+}
+
+export const EntriesProvider = ({ children }: EntriesProviderProps) => {
+  const entriesRef = useRef<PatchTypeWithKey[]>([]);
```
abcpro1:
```suggestion
const [entries, setEntries] = useState<PatchTypeWithKey[]>([]);
```
Usage looks like a state.

frontend/src/contexts/EntriesContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [, forceRerender] = useState(0);
```
abcpro1: This doesn't seem to do anything as consumers of `value` won’t see changes since it doesn't depend on `_`.

frontend/src/components/tasks/TaskDetailsPanel.tsx
```diff
@@ -117,87 +118,137 @@ export function TaskDetailsPanel({
         <TabNavContext.Provider value={{ activeTab, setActiveTab }}>
           <ProcessSelectionProvider>
             <ReviewProvider>
-              {/* Backdrop - only on smaller screens (overlay mode) */}
-              {!hideBackdrop && (
+              <EntriesProvider>
```
abcpro1, line 121: Consider wrapping only the `TodoPanel` and `LogsTab` rather than the entire panel to reduce unrelated re-renders on log updates.

abcpro1 (approved):

frontend/src/contexts/EntriesContext.tsx
stunningpixels: The useConversationHistory makes use of useRef rather than useState for performance reasons, I think claude got confused - good feedback, it was unnecessary here

PR #768, opened 2025-09-17
Default project setup (vibe-kanban)

LSRCT:
When a user has no projects setup, this is the case for new users, instead of making them pick, already setup the three most recent repos. frontend/src/components/projects/github-repository-picker.tsx already has a way to fetch these.


stunningpixels:
Adds two projects but should it be 3?

LSRCT:
> Adds two projects but should it be 3?

It has limits on time used for searching, locations to search and depth. May only find two.

stunningpixels:
I'm just surprised that when I click add project I see 3 but the automatic project add only adds 2

LSRCT:
> I'm just surprised that when I click add project I see 3 but the automatic project add only adds 2

For add project the limits are a bit less restrictive, we use max_depth=4 and 1200ms timeout. The user clicks this and knows what they are waiting for.

For auto setup the user is unaware this is happening so I used max_depth=3 and 1000ms timeout to make sure this finishes in time and doesn't crawl to deep into the users files.

Should I increase these limits?

PR #775, opened 2025-09-18
Introduce i18n (vibe-kanban)

stunningpixels:
Incremental adoption of i18n, this lays out the framework and translates the settings pages to Japanese as a test


stunningpixels:
CI script will land in a follow up PR as need to seed main with the script and eslint setup first or it will fail

PR #784, opened 2025-09-19
WIP use Gemini-CLI --experimental-acp

abcpro1:


crates/local-deployment/src/container.rs
```diff
@@ -355,180 +360,197 @@ impl LocalContainerService {
...
-                        Ok(None) => None,
-                        Err(e) => Some(Err(e)),
+        tokio::spawn(async move {
+            let mut exit_signal_future = exit_signal
+                .map(|rx| rx.map(|_| ()).boxed()) // wait for signal
+                .unwrap_or_else(|| std::future::pending::<()>().boxed()); // no signal, stall forever
+
+            let status_result: std::io::Result<std::process::ExitStatus>;
+
+            // Wait for process to exit, or exit signal from executor
```
ggordonhall, line 372: Can you add a comment explaining why we have this separate wait condition? E.g. explaining that this is to handle the fact that ACP processes do not exit and have to be manually shutdown

abcpro1:
done!

PR #791, opened 2025-09-19
Fix/kanban infinite horizontal dragging

britannio:
Previously, if a card in the Kanban board was dragged to the far right edge of the screen, it would continue scrolling.
With this change, we introduce a variant of the `restrictToFirstScrollableAncestor` dnd-kit modifier (https://docs.dndkit.com/api-documentation/modifiers#restricttofirstscrollableancestor).
This modifier additionally limits the x delta of the dragged card by comparing the right edge of the card to the right edge of the scrollable ancestor and ensuring that it is at least 16 units.


britannio:
A simpler variant of this solution is to set `autoScroll` to `false` in `DndContext` but it disables some desirable auto scrolling demonstrated in this video: 

https://github.com/user-attachments/assets/38f7fdc8-d1db-41e6-b51d-c21be811d757

frontend/src/components/ui/shadcn-io/kanban/index.tsx
```diff
@@ -160,6 +161,66 @@ export const KanbanHeader = (props: KanbanHeaderProps) =>
     </Card>
   );
 
+function restrictToBoundingRectWithRightPadding(
```
britannio, line 167: This is a variant of https://github.com/clauderic/dnd-kit/blob/e9215e820798459ae036896fce7fd9a6fe855772/packages/modifiers/src/utilities/restrictToBoundingRect.ts

britannio:
ah rip

britannio:
Closed in favour of https://github.com/BloopAI/vibe-kanban/pull/853

PR #811, opened 2025-09-23
Darkmode favicon (vibe-kanban)

LSRCT:
When in dark mode, the favicon is hard to see.


stunningpixels:
vk_icon_dark
vk_icon_light

Works but can you pls use these icons

PR #814, opened 2025-09-23
Use CWD as default for new project creation (vibe-kanban)

LSRCT:
The default directory for creating new, empty projects in vibe kanban is currently home, but should instead be cwd.

### DIFF
--- a/crates/server/src/routes/projects.rs
+++ b/crates/server/src/routes/projects.rs
@@ -61,8 +61,7 @@ pub async fn create_project(
     tracing::debug!("Creating project '{}'", name);
 
     // Validate and setup git repository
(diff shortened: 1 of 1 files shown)


frontend/src/components/projects/project-form-fields.tsx
```diff
@@ -349,7 +349,7 @@ export function ProjectFormFields({
                   type="text"
                   value={parentPath}
                   onChange={(e) => setParentPath(e.target.value)}
-                  placeholder="Home"
+                  placeholder="Current Directory"
```
stunningpixels, line 352: Should probs say what the current directory is

PR #819, opened 2025-09-23
chore: remove unused exports, components and hooks

ggordonhall:


stunningpixels:
Looks fine, just concerned about rebases to a couple big PRs that are open atm

ggordonhall:
Sure, we can keep this open. It's easy to re-apply

PR #821, opened 2025-09-23
feat: add `Back to app` button to redirect to last non-settings page

ggordonhall:
Track the last non-settings page with a `usePreviousPath` hook


frontend/src/pages/settings/SettingsLayout.tsx
```diff
@@ -20,13 +22,24 @@ const settingsNavigation = [
...
+    navigate(getPreviousPath());
+  };
 
   return (
     <div className="container mx-auto px-4 py-8">
       <div className="flex flex-col lg:flex-row gap-8">
         {/* Sidebar Navigation */}
         <aside className="w-full lg:w-64 lg:shrink-0 lg:sticky lg:top-8 lg:h-fit lg:max-h-[calc(100vh-4rem)] lg:overflow-y-auto">
           <div className="space-y-1">
+            <Button variant="ghost" onClick={handleGoBack} className="mb-4">
```
stunningpixels: I think you can do `onClick={() => navigate(-1)}`

ggordonhall: That just does browser back. This implements back to app, which navigates back to the last non-settings page

stunningpixels: Gotcha, maybe something like this then:

```
import { useEffect, useRef } from "react";
import { useLocation, useNavigate } from "react-router-dom";

export function useBackToNonSettings() {
  const navigate = useNavigate();
  const location = useLocation();
  const visited = useRef<string[]>([]);

  // Track pathnames as user navigates
  useEffect(() => {
    visited.current.push(location.pathname);
  }, [location]);

  return () => {
    // Find last non-settings route in history
    const idx = [...visited.current]
      .reverse()
      .findIndex((p) => !p.startsWith("/settings"));

    if (idx === -1) {
      navigate("/"); // fallback if none
    } else {
      const target = visited.current[visited.current.length - 1 - idx];
      navigate(target);
    }
  };
}
```

This info is already stored in the browser history so just trying to avoid using session storage. We also don't use session storage anywhere else in the frontend, instead we use React useState for most similar things.

Not super important, would avoid new patterns.

PR #822, opened 2025-09-23
Show only ahead commits in diff

LSRCT:
- Show diff for commits ahead, ignore diff for commits behind

PR 1/n for deprecating the base branch concept

### DIFF
--- a/crates/services/src/services/git.rs
+++ b/crates/services/src/services/git.rs
@@ -72,22 +72,20 @@ pub struct HeadInfo {
 }
 
 #[derive(Debug, Clone)]
-pub struct Commit {
-    id: git2::Oid,
-}
+pub struct Commit(git2::Oid);
 
 impl Commit {
     pub fn new(id: git2::Oid) -> Self {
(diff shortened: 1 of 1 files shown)


crates/services/src/services/git.rs
```diff
@@ -71,13 +71,32 @@ pub struct HeadInfo {
     pub oid: String,
 }
 
+#[derive(Debug, Clone)]
+pub struct Commit {
+    id: git2::Oid,
```
stunningpixels: Could just use git2::Oid instead of wrapping in this struct?

If needs to be struct you could use a tuple struct, see example: https://github.com/BloopAI/vibe-kanban/blob/main/crates/executors/src/executors/mod.rs#L167

LSRCT: I like the struct:
- Makes it clear this is a commit, not any Oid, moving away from base branch I think it's good to have extra clarity
- We avoid using git2 directly in lots of places it's not currently used in

Changed it to a tuple struct.

PR #835, opened 2025-09-24
enable the esc shortcut from the global settings page (vibe-kanban)

britannio:


stunningpixels:
Confirm if unsaved changes maybe?

britannio:
can we add auto save to settings?

britannio:
Replaced by https://github.com/BloopAI/vibe-kanban/pull/1018

britannio:
ah, it's technically missing an unsaved changes confirmation

PR #836, opened 2025-09-24
Improve auto project setup (vibe-kanban)

LSRCT:
- Increase timeouts for auto project setup
- Fix auto project setup not being backgrounded
- track manual/auto source for project creation

### DIFF
--- a/crates/server/src/routes/github.rs
+++ b/crates/server/src/routes/github.rs
@@ -182,6 +182,7 @@ pub async fn create_project_from_github(
                         "clone_url": payload.clone_url,
                         "has_setup_script": has_setup_script,
                         "has_dev_script": has_dev_script,
+                        "source": "github",
                     })),
                 )
                 .await;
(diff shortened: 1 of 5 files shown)


crates/server/src/routes/github.rs
```diff
@@ -182,6 +182,7 @@ pub async fn create_project_from_github(
                         "clone_url": payload.clone_url,
                         "has_setup_script": has_setup_script,
                         "has_dev_script": has_dev_script,
+                        "source": "github",
```
ggordonhall, line 185: We should probably make this a `project_created` event too

crates/services/tests/filesystem_repo_discovery.rs
```diff
@@ -0,0 +1,179 @@
+use std::{fs, path::Path};
+
+use services::services::filesystem::FilesystemService;
+use tempfile::TempDir;
+
+/// Helper function to create a directory structure
```
ggordonhall: Can wrap the whole file so we only compile for test builds
```
#[cfg(test)]
mod tests {
...
}
```

PR #837, opened 2025-09-24
feat: add functionality to attach existing pull requests to tasks

jtomaszewski:


crates/server/src/routes/task_attempts.rs
```diff
@@ -1707,6 +1714,37 @@ pub async fn get_task_attempt_children(
...
+pub async fn attach_existing_pr(
+    Extension(task_attempt): Extension<TaskAttempt>,
+    State(deployment): State<DeploymentImpl>,
+    Json(request): Json<AttachExistingPrRequest>,
+) -> Result<ResponseJson<ApiResponse<PrMerge>>, ApiError> {
+    // Check if a PR is already attached to this task attempt
+    let existing_merges = Merge::find_by_task_attempt_id(deployment.pool(), task_attempt.id).await?;
+
+    // Check if PR is already attached
+    for merge in existing_merges {
```
LSRCT, line 1726: This should check if the task has an *open* PR attached using pr_merge.pr_info.status for open PRs. A PR could have been closed and re-opened. Same in task_server.rs

LSRCT (commented):
There may be an easier way to achieve your goal, could you explain what you're trying to do? As far as I can see attaching a PR gives you two things:
- Ability to push from the UI
- Automatic column change to `Done` once the PR is merged.

Your method of attaching the PR has a few possible edge cases:
- No check if PR branch == task branch, nothing prevents users form attaching any PR to any attempt, even if they don't match.
- PR url and number aren't verified, these may not actually exist on GitHub and there would be no way to detach them again.

Did you consider using the GitHub API instead? Could be a `discover_and_attach_pr` function that actually checks if there is an open PR and then attaches it. That would make the call simpler with fewer arguments and you'd be sure things are correct.

jtomaszewski:
Thanks for your feedback. 

I went with your `discover_and_attach_pr ` suggestion - see https://github.com/BloopAI/vibe-kanban/pull/842 . 

If that one feels better (I think it does), let's close this one.

LSRCT:
Closing this in favour of #842

PR #840, opened 2025-09-24
Increase diff file size limit and display diff hunks

abcpro1:


ggordonhall:
Despite limiting to display hunks, I still get OOM issues with large files. Closing this in favour of the simple size limit fix

PR #847, opened 2025-09-25
Louis/keyboard shortcut improve

stunningpixels:
- j navigate to next task
- k navigate to previous task
- d to delete task
- scroll kanban when navigating


stunningpixels:
> Would be nice to add `h` and `l` for left right navigation

Addressed

Issue #859, opened 2025-09-26
CMD+A or select all text to delete to remove task description doesnt work


KeentGG:
https://github.com/user-attachments/assets/5319ad5e-ae70-4ed7-a70a-bd1845079dcb

when you edit a task description to empty it via highlighting all text and deleting it, after saving the description it is still not deleted.

Only works when you partially delete the description or delete all and add some characters (what i did on the very last part of the video).

stunningpixels:
Reproduced, fixing

stunningpixels:
Will be included in the next release

PR #862, opened 2025-09-26
Editing task description to remove it doesn't work (vibe-kanban)

stunningpixels:
When you edit a task description to empty, it doesn't get deleted after saving.

crates/server/src/routes/tasks.rs


stunningpixels:
Fixes https://github.com/BloopAI/vibe-kanban/issues/859

PR #863, opened 2025-09-26
feat: add task buttons to Kanban headers

ggordonhall:


britannio:
CleanShot 2025-09-26 at 14 14 46@2x

Would a counter be more useful. If users want an easier way to create tasks, we should inform them of the `c` shortcut.

abcpro1 (approved):

ggordonhall:
> CleanShot 2025-09-26 at 14 14 46@2x
> Would a counter be more useful. If users want an easier way to create tasks, we should inform them of the `c` shortcut.

I'm not sure how useful it is to know the exact number of tasks in a column. If it's a reasonably small number then you should be able to eyeball it anyway. This just makes the add task dialog easier to find. I agree that we should also encourage the keyboard shortcut though.

PR #866, opened 2025-09-26
Make diff stats much faster

abcpro1:


frontend/src/hooks/useDiffEntries.ts
stunningpixels, line 1: I think this can be merged with `frontend/src/hooks/useDiffStream.ts`?

stunningpixels (approved):

PR #872, opened 2025-09-26
Prevent useConversationHistory conflicts (vibe-kanban)

stunningpixels:
frontend/src/hooks/useConversationHistory.ts

Process logs can overwrite each other, let's move away from directly writing to displayedExecutionProcesses.current and instead use this helper:

```
const mergeIntoDisplayed = (updates: ExecutionProcessStateStore) => {
  displayedExecutionProcesses.current = {
    ...displayedExecutionProcesses.current,
    ...updates,
  };
};
```


abcpro1:
To-do leakage hasn't been solved by this for running tasks

https://github.com/user-attachments/assets/5b43f38d-2d3f-4340-9676-735f432b1432

stunningpixels:
@abcpro1 that\s expected, this change should solve running task logs not showing when there is a long running tool call

PR #894, opened 2025-09-30
Speed-up running process check

abcpro1:
Significantly reduces the running-process detection delays.

Before:

https://github.com/user-attachments/assets/563d2d3d-782f-40fe-81f5-0f6ef6aa6950

After:

https://github.com/user-attachments/assets/acd240f9-0dc3-4439-aa35-6affe91f0720


stunningpixels:
We can probably delete useAttemptExecution

abcpro1:
Are you sure? What happens to `processDetailQueries` and `stopExecution`?

stunningpixels:
I think `runningProcessDetails` is unused and `stopExecution` should probably be a tanstack mutation,

stunningpixels:
Can do as follow up, this doesn't make things worse - only noticed as haven't looked at this hook in a while

stunningpixels (approved):

PR #903, opened 2025-10-01
Fix wrong git status for remote branches

LSRCT:
fix wrong ahead/behind commit info for remote branches

### DIFF
--- a/crates/server/src/routes/task_attempts.rs
+++ b/crates/server/src/routes/task_attempts.rs
@@ -917,70 +917,72 @@ pub async fn get_task_attempt_branch_status(
         .git()
         .find_branch_type(&ctx.project.git_repo_path, &task_attempt.target_branch)?;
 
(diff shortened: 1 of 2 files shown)


crates/server/src/routes/task_attempts.rs
```diff
@@ -917,70 +917,72 @@ pub async fn get_task_attempt_branch_status(
...
+                &task_attempt.branch,
+                &task_attempt.target_branch,
+            )?;
+            (Some(a), Some(b))
+        }
+        BranchType::Remote => {
+            let github_config = deployment.config().read().await.github.clone();
+            let token = github_config
+                .token()
+                .ok_or(ApiError::GitHubService(GitHubServiceError::TokenInvalid))?;
```
abcpro1, line 933: GtiHub token should not be required for this operation. It can be used if available, otherwise use current state of origin/main.
The `branch-status` endpoint fails if logged out of GitHub in vibe-kanban and base branch is a remote branch. The `Git` section become invisible in this case, disabling all git operations.

PR #904, opened 2025-10-01
feat: MCP upgades

ggordonhall:
- add `start_task_attempt` mcp tool
- connect to vk api in mcp rather than direct db access

### DIFF
--- a/crates/server/src/mcp/task_server.rs
+++ b/crates/server/src/mcp/task_server.rs
@@ -23,7 +23,7 @@ use crate::routes::task_attempts::CreateTaskAttemptBody;
 #[derive(Debug, Deserialize, schemars::JsonSchema)]
 pub struct CreateTaskRequest {
     #[schemars(description = "The ID of the project to create the task in. This is required!")]
-    pub project_id: String,
+    pub project_id: Uuid,
     #[schemars(description = "The title of the task")]
     pub title: String,
     #[schemars(description = "Optional description of the task")]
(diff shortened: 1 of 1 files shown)


crates/server/src/mcp/task_server.rs
```diff
@@ -161,19 +183,31 @@ pub struct DeleteTaskRequest {
...
 }
 
+#[derive(Debug, Deserialize, schemars::JsonSchema)]
+pub struct StartTaskAttemptRequest {
+    #[schemars(description = "The ID of the task to start")]
+    pub task_id: String,
+    #[schemars(
+        description = "The coding agent executor to run ('CLAUDE_CODE', 'CODEX', 'GEMINI', 'CURSOR', 'OPENCODE')"
+    )]
+    pub executor: String,
```
LSRCT, line 189: Could we use the enum here instead of String?

crates/server/src/mcp/task_server.rs
```diff
@@ -161,19 +183,31 @@ pub struct DeleteTaskRequest {
...
+    pub base_branch: String,
+}
+
 #[derive(Debug, Serialize, schemars::JsonSchema)]
-pub struct DeleteTaskResponse {
-    pub success: bool,
+pub struct StartTaskAttemptResponse {
     pub message: String,
-    pub deleted_task_id: Option<String>,
+    pub task_id: String,
```
LSRCT, line 199: Could the ids be typed Uuid instead of String?

crates/server/src/mcp/task_server.rs
```diff
@@ -186,27 +220,97 @@ pub struct GetTaskRequest {
...
 #[derive(Debug, Clone)]
 pub struct TaskServer {
-    pub pool: SqlitePool,
+    client: reqwest::Client,
+    base_url: String,
     tool_router: ToolRouter<TaskServer>,
 }
 
 impl TaskServer {
     #[allow(dead_code)]
```
LSRCT:
```suggestion
```

LSRCT (approved):

crates/server/src/mcp/task_server.rs
ggordonhall: There are some limitations with using enum types in MCP tool definitions. If I use a `BaseCodingAgent` here Claude Code fails to parse the tool definitions and can't interact with the server. Will look into in future

PR #922, opened 2025-10-03
Add variant/config cycle keyboard shortcut (vibe-kanban)

LSRCT:
Add a shortcut to cycle between agent configurations (previously called profile variants) when typing a follow up. Should be active when the user is editing the follow-up text box.
User shift tab to trigger the shortcut.


ggordonhall (approved):
We should hook up the retry box to these shortcuts too

LSRCT:
> We should hook up the retry box to these shortcuts too

Agree, will do in a follow-up

PR #924, opened 2025-10-03
Event for merged PRs (vibe-kanban)

LSRCT:
Add an event for when a PR is merged.
crates/services/src/services/pr_monitor.rs


stunningpixels:
Do we care about PRs merged? PRs created and work directly merged is probably decent enough metric

PR #926, opened 2025-10-03
create a default task for installing the companion (vibe-kanban)

britannio:


britannio:
I actually have a better idea that combines two other ideas for streamlining the process for new users.

britannio:
Closed in favour of https://github.com/BloopAI/vibe-kanban/pull/953

PR #928, opened 2025-10-03
No user deny messages in history (vibe-kanban)

LSRCT:
When interacting with an approval and the users chooses to deny the request with feedback, the feedback text should be displayed as a user message. Currently it's not displayed

### DIFF
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -39,6 +39,10 @@ async fn get_backend_port() -> std::io::Result<u16> {
 
 const CONFIRM_HOOK_SCRIPT: &str = include_str!("./hooks/confirm.py");
 
+/// Natural language marker we add in our Python hook to denote user feedback
+/// This marker is added by our confirm.py hook script and is robust to Claude Code format changes
+const USER_FEEDBACK_MARKER: &str = "User feedback: ";
+
 fn base_command(claude_code_router: bool) -> &'static str {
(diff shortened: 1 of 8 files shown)


crates/executors/src/executors/hooks/confirm.py
```diff
@@ -147,7 +147,10 @@ def main():
             json_success()
         elif status == "denied":
             reason = result.get("reason")
-            json_error(reason)
+            # Prefix user feedback with marker for extraction by Rust code
```
ggordonhall: Might be easier to pass the user feedback marker via the command line? Also, will be more robust to script version upgrades

crates/executors/src/executors/claude.rs
```diff
@@ -313,6 +317,39 @@ exit "$exit_code"
...
+    };
+
+    // Look for our natural language marker and extract everything after it
+    if let Some(pos) = content_str.find(USER_FEEDBACK_MARKER) {
+        let after_marker = &content_str[pos + USER_FEEDBACK_MARKER.len()..];
+        let feedback = after_marker.trim();
+        if !feedback.is_empty() {
+            return Some(feedback.to_string());
+        }
+    }
```
ggordonhall:
```suggestion
    flat.split_once(USER_FEEDBACK_MARKER)
        .map(|(_, rest)| rest.trim().to_string())
        .filter(|s| !s.is_empty())
```

PR #935, opened 2025-10-04
Support adding subtasks on tasks that are in TODO state.

gburch:
This is useful to be able to map out many tasks for both humans and AI before work has started. Currently the API/implementation prevents doing this.


gburch:
I do not know how to get the CI tests to pass? We don't check-in .sqlx/ but the database_url doesn't seem to be picked up for the gh actions test? What is the best practice here?

gburch:
> I do not know how to get the CI tests to pass? We don't check-in .sqlx/ but the database_url doesn't seem to be picked up for the gh actions test? What is the best practice here?

AI helped me, I was looking in the wrong place.

gburch:
Finally ready for review.

ggordonhall:
Hi @gburch, thanks for the PR. Unfortunately this is something that we want to tackle internally once we've refactored the `Task` model. Closing this one!

PR #936, opened 2025-10-04
Support multiple remotes for PRs

gburch:
Let users choose the remote:
Add a remote selector to the Create PR dialog, so users can explicitly choose which remote to create the PR against.

Default the branch's configured remote
The code already has logic to get the remote from a branch's configuration (get_remote_from_branch_ref), but get_github_repo_info() doesn't use it - it just uses the default remote.

image

### DIFF
--- a/crates/server/src/routes/task_attempts.rs
+++ b/crates/server/src/routes/task_attempts.rs
@@ -754,14 +754,12 @@ pub async fn create_github_pr(
             )));
         }
     }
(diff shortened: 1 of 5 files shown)


gburch:
Perhaps stick it under an "Advanced >" disclosure?

ggordonhall:
As of https://github.com/BloopAI/vibe-kanban/pull/1976 we now support PR'ing into upstream repos. Closing this.

PR #944, opened 2025-10-04
If PR open and no commits to push, link to open PR (vibe-kanban)

stunningpixels:
frontend/src/components/tasks/Toolbar/GitOperations.tsx

### DIFF
--- a/frontend/src/components/tasks/Toolbar/GitOperations.tsx
+++ b/frontend/src/components/tasks/Toolbar/GitOperations.tsx
@@ -142,6 +142,23 @@ function GitOperations({
     };
   }, [branchStatus?.merges]);
 
+  const remoteAhead = branchStatus?.remote_commits_ahead ?? 0;
+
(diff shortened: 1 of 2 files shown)


LSRCT:
Do we need this? Currently the PR is always viewable using the link we display above the `Push`-Button.
If it's not obvious that the `PR #00`-label is a link maybe we could make that more obvious instead of having a dual-purpose button?
Screenshot 2025-10-06 at 16 33 07

stunningpixels:
Got it, yes didn't realise that was a link - if you have a better idea for design happy for you to look at, otherwise I think button is very clear

LSRCT:
Screenshot 2025-10-06 at 16 43 53

Probably doesn't need a lot, what if we added an external link icon to the label?

stunningpixels:
Looks good - maybe make it blue

LSRCT:
#951 adds the blue `external link`-icon

PR #949, opened 2025-10-06
Setting to override branch prefix (vibe-kanban)

stunningpixels:
Screenshot 2025-10-06 at 15 58 36

Currently branches are prefixed `/vk` but we should allow the user to customise this with a git branch name compliant string from their global settings.

New settings can be appended to v7 without needing to create a new version.

frontend/src/pages/settings/GeneralSettings.tsx
crates/services/src/services/config/versions/v6.rs


crates/utils/src/git.rs
```diff
@@ -0,0 +1,66 @@
+pub fn is_valid_branch_prefix(prefix: &str) -> bool {
```
ggordonhall, line 1: Might be able to replace all of this logic with this? https://docs.rs/git2/latest/git2/struct.Branch.html#method.name_is_valid

frontend/src/pages/settings/GeneralSettings.tsx
```diff
@@ -533,6 +525,55 @@ export function GeneralSettings() {
...
+          <CardDescription>
+            {t('settings.general.git.description')}
+          </CardDescription>
+        </CardHeader>
+        <CardContent className="space-y-4">
+          <div className="space-y-2">
+            <Label htmlFor="git-branch-prefix">
+              {t('settings.general.git.branchPrefix.label')}
+            </Label>
+            <Input
```
ggordonhall, line 567: We could have some FE validation of the branch name? If the user does enter an invalid prefix then they end up in a state where they can't start a task but no error message is displayed

PR #960, opened 2025-10-07
Check i18n consistency (vibe-kanban)

stunningpixels:
Extend the scripts/check-i18n.sh script to check that each translation file contains the exact same keys


stunningpixels:
Example of the script catching missing keys here: https://github.com/BloopAI/vibe-kanban/actions/runs/18308409527/job/52131019030#step:6:40

ggordonhall (approved):

PR #962, opened 2025-10-07
Live Diff Service Refactor (vibe-kanban)

LSRCT:
- Fix initial diff not being sent before filesystem watcher is ready, causing slow diffs
- Refactor live diffs
- Improve filesystem watcher `.gitignore` discovery performance

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -10,7 +10,6 @@ use std::{
 };
 
 use anyhow::anyhow;
-use async_stream::try_stream;
 use async_trait::async_trait;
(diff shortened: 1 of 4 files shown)


ggordonhall (approved):
I think this makes sense. We could also decrease the time we wait before displaying to `No diffs` message to 3ish seconds?

LSRCT:
> I think this makes sense. We could also decrease the time we wait before displaying to `No diffs` message to 3ish seconds?

It is at 3 seconds already. If we wanted to make it really snappy we could try and make `No diffs` triggered by the backend, but that would be in another PR

PR #969, opened 2025-10-07
replace folder icon with IDE icon in the nav bar

britannio:


britannio:
image

mickaelxd:
Hey there, I think using IntelliJ might be too specific — could we switch to a more generic IDE icon?

britannio:
> Hey there, I think using IntelliJ might be too specific — could we switch to a more generic IDE icon?

CleanShot 2025-10-08 at 09 36 05@2x

The icon matches the one you choose in settings.

PR #997, opened 2025-10-11
Move project settings to global settings as a new tab (vibe-kanban)

LSRCT:
Move project settings to global settings, giving them a new tab, similar to agent settings. It should have a dropdown to select the project and then offer the same settings the current project settings offer


stunningpixels:
Works well but what has happened to task templates

LSRCT:
> Works well but what has happened to task templates

Project level templates can't be configured after we merge this, global templates are still available.
In a next step global templates will be replaced by tags, already working on it.
Should be fine, nobody uses templates AFAIK

PR #1012, opened 2025-10-14
Close settings X (vibe-kanban)

LSRCT:
Currently settings can be exited via a 'back to app' button in the top left. It should be an X in the top right, like in most other places.


britannio:
Do we want a confirmation dialog if you attempt to close settings while there are unsaved changed? https://github.com/BloopAI/vibe-kanban/pull/835 had this.

ggordonhall:
@britannio can layer this in a follow up PR

PR #1049, opened 2025-10-17
use existing config system instead of local storage for showcase tracking (vibe-kanban)

britannio:


crates/services/src/services/config/versions/v7.rs
```diff
@@ -11,6 +13,12 @@ fn default_git_branch_prefix() -> String {
     "vk".to_string()
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, TS, Default)]
+pub struct ShowcaseState {
+    #[serde(default)]
+    pub seen_versions: HashMap<String, u32>,
```
LSRCT: will this cause the users config files to get longer and longer over time?

britannio: In proportion to the number of showcases that they have seen, yes.

britannio: It could be reversed to `pending_showcases` and every time a new showcase is introduced, we add a migration to insert it into the user's config, then remove it once shown. Plus some state for performing migrations exactly once.

Or it could move to the db?

LSRCT: Got it, it's fine the way it is.

ggordonhall (approved):

PR #1073, opened 2025-10-21
feat(executors): use --append-system-prompt for Claude Code

namastex888:
## Summary

Replace string concatenation with native Claude CLI `--append-system-prompt` parameter for appending system instructions. This provides cleaner separation between user prompts and system instructions.

## Benefits

- **Better separation of concerns**: User prompts and system context are now handled separately
- **Cleaner code**: Removes the need for `combine_prompt()` calls in spawn methods
- **Native CLI support**: Leverages built-in Claude Code functionality

## Changes

- Add `--append-system-prompt` CLI parameter in `build_command_builder()`
- Remove `combine_prompt()` concatenation in `spawn()` method
- Remove `combine_prompt()` concatenation in `spawn_follow_up()` method

## Testing

- All existing tests pass
- Verified with real Claude Code sessions in production use

## Context

The `AppendPrompt` struct already supports this pattern, but was using string concatenation via `combine_prompt()`. The Claude CLI natively supports `--append-system-prompt`, which is the cleaner approach.

## Files Modified

- `crates/executors/src/executors/claude.rs`

---

**Note**: This PR is submitted from our fork where we've been using this improvement successfully.

### DIFF
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -117,6 +117,12 @@ impl ClaudeCode {
         if let Some(model) = &self.model {
             builder = builder.extend_params(["--model", model]);
         }
+
+        // Add append-system-prompt if configured
+        if let Some(append_text) = self.append_prompt.0.as_ref() {
+            builder = builder.extend_params(["--append-system-prompt", append_text]);
+        }
+
         builder = builder.extend_params([
             "--verbose",
             "--output-format=stream-json",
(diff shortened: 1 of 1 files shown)


stunningpixels:
Hey @namastex888 - as this doesn't change performance, functionality or significantly improve code quality I'm closing as it will cost the team time to test.

namastex888:
Appending system prompts instead of injecting it the input indeed brings extra performance to LLMs, but your call..

PR #1080, opened 2025-10-22
Claude approval refactor

LSRCT:

PR #1087, opened 2025-10-23
Address Claude Code planning mode

speleobrad:
There's an issue with Claude  Code planning mode on Windows. 

### Root Cause

The problem was in `crates/executors/src/executors/claude.rs:294-318`. The `create_watchkill_script` function generated a **bash-only script** that monitors Claude Code output and exits early when it detects "Approval request timed out". This bash script is incompatible with Windows, which uses `cmd.exe`.

### The Fix

`create_watchkill_script` function was modified to be platform-aware:
- **Windows**: Generates a batch script (`.bat`) using `cmd.exe` syntax with `for /f` loops and `findstr`
- **Unix/Linux/macOS**: Uses the existing bash script with process substitution and pipelines

The Windows batch script version is simpler than the bash version and doesn't preserve exit codes as elegantly, but it fulfills the core requirement of detecting the timeout phrase

See the changes in `crates/executors/src/executors/claude.rs:294-339`

### DIFF
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -295,10 +295,30 @@ fn create_watchkill_script(command: &str) -> String {
     // Hack: we concatenate so that Claude doesn't trigger the watchkill when reading this file
     // during development, since it contains the stop phrase
     let claude_plan_stop_indicator = concat!("Approval ", "request timed out");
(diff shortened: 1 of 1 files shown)


LSRCT:
`create_watchkill_script` is not going to exist after #1080. That should also fix your problem, closing this

Issue #1044, opened 2025-10-17
Diff preview with LF / CR LF


kfuglsang:
When running Vibe Kanban v0.0.108 on Windows, every task displayed in the Review stage shows an excessive number of line differences that appear unrelated to actual code changes.

### Observed Behavior

The “old” version of the files uses LF line endings.

The “new” (changed) version uses CRLF line endings.

As a result, Vibe Kanban highlights entire files as changed, even though there are no real content modifications.

### Expected Behavior

Line-ending differences (LF vs CRLF) should be ignored when computing or displaying diffs — similar to how most Git clients handle this automatically.

### Environment

OS: Windows 11

Vibe Kanban Version: 0.0.108

Git Version: 2.51.0.windows.2

Repo Settings: core.autocrlf enabled

### Additional Notes

Other Git tools (e.g., GitHub Desktop, VS Code, SourceTree) correctly ignore line-ending differences, so this behavior seems specific to Vibe Kanban’s diff implementation.

ggordonhall:
This is on the short term roadmap!

ggordonhall:
I selector to hide whitespace changes in diffs is in recent releases: https://github.com/BloopAI/vibe-kanban/releases/tag/v0.0.112-20251028183536

https://github.com/BloopAI/vibe-kanban/pull/1067

PR #1095, opened 2025-10-23
Fix Windows diff view false positives from CRLF line endings

speleobrad:
Normalize line endings to LF when comparing repository content with working tree files to prevent all lines showing as changed on Windows when core.autocrlf=true converts LF to CRLF.

Added normalize_line_endings() helper applied to both blob_to_string() and read_file_to_string() for consistent comparison across platforms.

### DIFF
--- a/crates/services/src/services/git.rs
+++ b/crates/services/src/services/git.rs
@@ -553,6 +553,13 @@ impl GitService {
             .unwrap_or_default()
     }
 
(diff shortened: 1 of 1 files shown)


speleobrad:
Should address #1044

stunningpixels:
https://github.com/BloopAI/vibe-kanban/pull/1067 should solve this

PR #1101, opened 2025-10-25
Fix Claude Code executor tasks being marked as failed on success

### LINKED ISSUE #1100: Claude Code tasks incorrectly marked as failed despite successful completion
## Description

Claude Code executor tasks are being marked as failed (exit code 1) in the UI even when they complete successfully internally.

## Steps to Reproduce

1. Create a task using the Claude Code executor
2. Task completes successfully (visible in logs with `"subtype":"success"`)
3. UI shows "Try Again" button as if the task failed
4. Database shows `exit_code: 1` and `status: failed`

## Root Cause

The Claude Code executor doesn't provide an `exit_signal` like the Codex executor does. Without an exit_signal, the container process monitoring logic falls back to using the raw process exit code. For some reason, the npx Claude Code process exits with code 1 even on successful completion.

The Codex executor works correctly because it provides an exit_signal that forces success regardless of the actual exit code.

## Expected Behavior

Claude Code tasks should be marked as completed when they finish successfully, matching the behavior of the Codex executor.

## Actual Behavior

Claude Code tasks are always marked as failed with exit code 1, even when the internal execution succeeds.

knowsuchagency:
## Summary
Fixes #1100

Claude Code tasks were being incorrectly marked as failed (exit code 1) even when they completed successfully. This resulted in the UI showing a "Try Again" button for successful executions.

### The Deeper Issue: Watchkill Script Bug

After thorough investigation, we discovered that the exit code 1 is **not coming from Claude Code itself**, but from a **bug in the `watchkill` wrapper script** used when plan mode is enabled.

**The Watchkill Script** (`crates/executors/src/executors/claude.rs:373-397`):
```bash
#!/usr/bin/env bash
set -euo pipefail

word="Approval request timed out"

exit_code=0
while IFS= read -r line; do
    printf '%s\n' "$line"
    if [[ $line == *"$word"* ]]; then
        exit 0
    fi
done < <(bash -lc {cmd} <&0 2>&1)

exit_code=${PIPESTATUS[0]}  # ← BUG: PIPESTATUS doesn't work with process substitution!
exit "$exit_code"
```

**The Bug**: `${PIPESTATUS[0]}` is designed for pipes (`cmd1 | cmd2`), not process substitution (`< <(cmd)`). After a `while read < <(command)` loop, `PIPESTATUS[0]` always returns 1, regardless of the command's actual exit code.

**Evidence**:
- Claude Code itself exits with code 0 when successful
- Database shows tasks with `exit_code=1` BUT logs contain `{"type":"result","subtype":"success"}`
- Only affects tasks with `"plan": true` in executor profile (which triggers the watchkill wrapper)
- Reproduced: `bash -c 'while read x; do :; done < <(exit 0); echo ${PIPESTATUS[0]}'` → Always prints `1`

### Why This Fix Works

Without an exit_signal, the container process monitoring logic falls back to using the raw process exit code from the watchkill script, which is always 1 due to the bug above.

This fix:
1. Monitors Claude Code's stdout stream for success indicators
2. Sends an `exit_signal` when `{"type":"result"...}` is detected
3. Container receives the signal and marks the task as completed (exit code 0), bypassing the broken exit code

The Codex executor already uses this pattern and works correctly.

### Future Considerations

While this fix works, the watchkill script itself should be fixed or removed:
- **Option A**: Fix the script to correctly capture exit codes (use background process + `wait` instead of process substitution)
- **Option B**: Remove the watchkill script and rely on exit_signal as the standard pattern
- **Option C**: Make exit_signal the standard approach for all executors (it's actually cleaner and more explicit)

The exit_signal approach is superior because:
- More explicit about success/failure detection
- Works regardless of wrapper scripts or shell quirks
- Can detect success before the process even exits
- Immune to PIPESTATUS bugs or signal handling issues

## Changes
- Added `exit_signal` support to Claude Code executor
- Duplicates stdout to monitor JSON output stream
- Detects the `{"type":"result"...}` message indicating successful completion
- Sends exit_signal when success is detected, forcing the task to be marked as completed
- Applied to both `spawn()` and `spawn_follow_up()` methods

## Testing
- Tested with multiple Claude Code tasks
- Verified tasks now correctly show as completed instead of failed
- Confirmed behavior now matches Codex executor

## Test plan
- [x] Create a Claude Code task that makes no file changes
- [x] Verify task is marked as completed (not failed)
- [x] Verify UI shows success state instead of "Try Again" button
- [x] Create a Claude Code task that makes file changes
- [x] Verify successful completion and commit

### DIFF
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -187,12 +187,12 @@ impl StandardCodingAgentExecutor for ClaudeCode {
                         }
 
                         // Check if this is a result message
-                        if let Ok(json) = serde_json::from_str::<serde_json::Value>(line) {
-                            if json.get("type").and_then(|v| v.as_str()) == Some("result") {
-                                // Found result message - signal completion
-                                let _ = exit_signal_tx.send(());
-                                return;
-                            }
+                        if let Ok(json) = serde_json::from_str::<serde_json::Value>(line)
+                            && json.get("type").and_then(|v| v.as_str()) == Some("result")
+                        {
+                            // Found result message - signal completion
+                            let _ = exit_signal_tx.send(());
+                            return;
                         }
                     }
                 }
@@ -272,12 +272,12 @@ impl StandardCodingAgentExecutor for ClaudeCode {
                         }
 
                         // Check if this is a result message
-                        if let Ok(json) = serde_json::from_str::<serde_json::Value>(line) {
-                            if json.get("type").and_then(|v| v.as_str()) == Some("result") {
-                                // Found result message - signal completion
-                                let _ = exit_signal_tx.send(());
-                                return;
-                            }
+                        if let Ok(json) = serde_json::from_str::<serde_json::Value>(line)
+                            && json.get("type").and_then(|v| v.as_str()) == Some("result")
+                        {
+                            // Found result message - signal completion
+                            let _ = exit_signal_tx.send(());
+                            return;
                         }
                     }
                 }


stunningpixels:
This will be solved in a different way by https://github.com/BloopAI/vibe-kanban/pull/1080

PR #1098, opened 2025-10-24
feat: Enhance executable resolution by refreshing PATH

abcpro1:
Improve the detection of CLI agents and IDEs/editors by refreshing the `PATH` environment variable. This resolves a common onboarding issue when newly installed executables weren't being discovered without a restart of vibe-kanban in a new terminal.

Can be reviewed commit-by-commit.

PR #1105, opened 2025-10-25
fix: respect user's default shell from SHELL env var

### LINKED ISSUE #1104: Shell selection ignores user's default shell
## Description

Currently, Vibe Kanban hardcodes a preference for zsh over bash when selecting which shell to use for command execution. This behavior ignores the user's configured default shell set in the `SHELL` environment variable.

## Current Behavior

The `get_shell_command()` function in `crates/utils/src/shell.rs` checks for shells in this order:
1. zsh (if /bin/zsh exists)
2. bash (if /bin/bash exists)
3. sh (fallback)

This means users with bash as their default shell will still get zsh if it's installed, leading to inconsistent behavior.

## Expected Behavior

Vibe Kanban should respect the user's `SHELL` environment variable, which reflects their configured default shell, and only fall back to hardcoded preferences if that variable is not set or invalid.

## Impact

Users may experience:
- Unexpected shell behavior
- Inconsistent environment across system commands
- Issues with shell-specific configurations not being honored

knowsuchagency:
## Summary

Fixes #1104

This PR updates the shell selection logic to respect the user's default shell from the `SHELL` environment variable instead of hardcoding a preference for zsh over bash.

## Changes

- Modified `get_shell_command()` in `crates/utils/src/shell.rs` to:
  1. Check the `SHELL` environment variable first
  2. Extract and validate the shell name from the path
  3. Support common shells: bash, zsh, fish, ksh, dash, sh
  4. Fall back to bash → sh if `SHELL` is not set or contains an unknown shell
  5. Removed hardcoded zsh preference over bash

## Testing

- All 112 workspace tests pass
- Code formatting verified with `cargo fmt --check`
- Linting passes with `cargo clippy --all --all-targets --all-features -- -D warnings`

## Behavior

**Before**: Always prefers zsh if installed, regardless of user's default shell
**After**: Uses the shell specified in `SHELL` env var, falls back to bash/sh if not set

### DIFF
--- a/crates/utils/src/shell.rs
+++ b/crates/utils/src/shell.rs
@@ -4,15 +4,38 @@
 ///
 /// Returns (shell_program, shell_arg) where:
 /// - Windows: ("cmd", "/C")
-/// - Unix-like: ("sh", "-c") or ("bash", "-c") if available
+/// - Unix-like: Respects the user's SHELL environment variable, falls back to bash or sh
 pub fn get_shell_command() -> (&'static str, &'static str) {
     if cfg!(windows) {
         ("cmd", "/C")
     } else {
-        // Prefer zsh or bash if available, fallback to sh
-        if std::path::Path::new("/bin/zsh").exists() {
-            ("zsh", "-c")
-        } else if std::path::Path::new("/bin/bash").exists() {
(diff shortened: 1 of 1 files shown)


abcpro1:
Superseded by #1098

PR #1115, opened 2025-10-28
Filter processes shown in conversation history

stunningpixels:
Previously we weren't filtering out dev servers from an upstream part of the conversation history pipeline, this caused issues.


britannio:
confirmed fix!

PR #1121, opened 2025-10-29
fix: prevent resuming completed Claude Code sessions

MakoTano:
## Problem

Follow-up requests were failing with "No assistant message found" error immediately after Claude Code spawned.

### Root Cause

The `find_latest_session_id_by_task_attempt` query didn't check the process status, so it returned session IDs from **completed** processes. Claude Code cannot resume terminated sessions using `--fork-session --resume`, causing immediate failures.

### Investigation Details

- Initial requests worked perfectly - Claude Code responded and completed normally
- Follow-up requests failed because they tried to resume a completed session
- Session files existed in `~/.claude/debug/{session_id}.txt` and showed `SessionEnd` events
- Manual invocation of Claude Code worked fine

## Solution

Added `ep.status = 'running'` condition to only return session IDs from active processes.

Now follow-up requests will:
- Resume an active session if Claude Code is still running
- Start a fresh initial request if the previous session completed

## Testing

Tested with multiple scenarios:
1. ✅ Initial comment → Claude Code responds successfully
2. ✅ Follow-up comment after completion → Creates new session successfully
3. ✅ Multiple follow-ups work as expected

---

### DIFF
--- a/crates/db/src/models/execution_process.rs
+++ b/crates/db/src/models/execution_process.rs
@@ -279,10 +279,11 @@ impl ExecutionProcess {
         let row = sqlx::query!(
             r#"SELECT es.session_id
                FROM execution_processes ep
-               JOIN executor_sessions es ON ep.id = es.execution_process_id  
+               JOIN executor_sessions es ON ep.id = es.execution_process_id
                WHERE ep.task_attempt_id = $1
                  AND ep.run_reason = 'codingagent'
                  AND ep.dropped = FALSE
+                 AND ep.status = 'running'
                  AND es.session_id IS NOT NULL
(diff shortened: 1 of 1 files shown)


LSRCT:
Hi, thanks for the PR!
I don't quite understand what problem this solves, could you explain how to reproduce your issue?

Follow-ups work just for for me when using Claude Code.
Generally it's not possible to follow-up on a running process, we only allow it for completed processes. The only time we continue in the same process is for passing feedback in approvals/plan mode, but in that case we don't need the session id at all.

ggordonhall:
@MakoTano Am closing this PR. Feel free to open an issue if the problem persists!

PR #1122, opened 2025-10-29
Fix profiles

stunningpixels:
Previous versions of profiles.json can't be parsed.

I can see we did add serde rename, but without also doing this for the strum_discriminants it won't parse correctly.

#[strum_discriminants(serde(alias = "CURSOR"))]


stunningpixels:
Have also changed the warn to an error so sentry will catch similar in future

PR #1126, opened 2025-10-29
Make Clerk session token handling more robust

abcpro1:
- Refresh the Clerk token in the frontend in fixed intervals.
- Improve token aquisiton in the backend to reduce spurious auth errors.

### DIFF
--- a/frontend/src/components/ClerkSessionRefresher.tsx
+++ b/frontend/src/components/ClerkSessionRefresher.tsx
@@ -1,7 +1,9 @@
 import { useAuth } from '@clerk/clerk-react';
 import { useEffect } from 'react';
-import { refreshClerkSession } from '../lib/api';
+import { refreshClerkSession } from '../lib/clerk';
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/ClerkSessionRefresher.tsx
```diff
@@ -0,0 +1,29 @@
+import { useAuth } from '@clerk/clerk-react';
+import { useEffect } from 'react';
+import { refreshClerkSession } from '../lib/api';
+
+const CLERK_SESSION_REFRESH_INTERVAL_MS = 25_000;
```
ggordonhall, line 7: Can you add a comment explaining this choice of interval?

frontend/src/lib/api.ts
```diff
@@ -121,6 +121,28 @@ async function getClerkToken(): Promise<string | null> {
   }
 }
 
+export async function refreshClerkSession(): Promise<void> {
```
ggordonhall: There's quite a lot of Clerk logic now. Can you move into a `lib/clerk.ts` and import here?

PR #1135, opened 2025-10-30
Catchup on WebSocket disconnection

abcpro1:
- Change the WebSocket auto-reconnection logic to perform the catchup procedure.
- Close the WebSocket when the remote sends a lag signal in the form of an Error message. A catchup is crucial in this case.
- Add a maximum delay between the end of last catchup and the start of a successful WebSocket connection to restart the catchup if the user-session is disconnected between the two operations.

### DIFF
--- a/crates/services/src/services/share.rs
+++ b/crates/services/src/services/share.rs
@@ -154,7 +154,7 @@ impl RemoteSync {
 
             let ws_url = self.config.websocket_endpoint(last_seq)?;
             let (close_tx, close_rx) = oneshot::channel();
-            let remote = match spawn_shared_remote(
+            let ws_connection = match spawn_shared_remote(
                 self.processor.clone(),
                 &self.sessions,
                 ws_url,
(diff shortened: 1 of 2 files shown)


crates/services/src/services/share.rs
```diff
@@ -166,24 +262,36 @@ async fn spawn_shared_remote(
...
     let session_source = sessions.clone();
     let ws_config = WsConfig {
         url,
-        autoreconnect: true,
-        reconnect_base_delay: std::time::Duration::from_secs(1),
-        reconnect_max_delay: std::time::Duration::from_secs(30),
+        autoreconnect: false,
+        // unused since autoreconnect is false
+        reconnect_base_delay: WS_BACKOFF_BASE_DELAY,
+        reconnect_max_delay: WS_BACKOFF_MAX_DELAY,
```
ggordonhall:
```suggestion
```

PR #1137, opened 2025-10-30
Close the gap between tasks snapshot and activity seq Database reads in `bulk_fetch()`

abcpro1:
Postgres doesn't prevent phantom reads by default.

### DIFF
--- a/crates/remote/src/db/tasks.rs
+++ b/crates/remote/src/db/tasks.rs
@@ -245,6 +245,11 @@ impl<'a> SharedTaskRepository<'a> {
         &self,
         organization_id: &str,
     ) -> Result<BulkFetchResult, SharedTaskError> {
+        let mut tx = self.pool.begin().await?;
+        sqlx::query("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ")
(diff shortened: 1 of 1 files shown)


crates/remote/src/db/tasks.rs
```diff
@@ -329,9 +335,12 @@ impl<'a> SharedTaskRepository<'a> {
             "#,
             organization_id
         )
-        .fetch_one(self.pool)
+        .fetch_one(&mut *tx)
         .await?;
 
+        // release the "repeatable-read" snapshot
```
ggordonhall:
```suggestion
```

crates/remote/src/db/tasks.rs
```diff
@@ -245,6 +245,12 @@ impl<'a> SharedTaskRepository<'a> {
         &self,
         organization_id: &str,
     ) -> Result<BulkFetchResult, SharedTaskError> {
+        let mut tx = self.pool.begin().await?;
+        // Ensure no gaps between tasks snapshot and activity seq read
```
ggordonhall:
```suggestion
```

PR #1144, opened 2025-10-30
feat(mcp): include attempt notes in get_task

mbernier:
Expose attempts-with-notes via REST and surface attempt summaries via MCP get_task when include_attempts=true.

Noticed that when the include_attempts param was being passed, the data was not being included in the response. 

It is now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,13 @@
+## Changelog
+
+### [Unreleased]
+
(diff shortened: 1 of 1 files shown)


stunningpixels:
Hey @mbernier thanks for the contribution. I was thinking it might be cleaner to have a separate tool that fetches attempts?

Alternatively we could always return attempts with task, and call it 'attempts' rather than 'attempt notes'. Wdyt?

PR #1151, opened 2025-10-31
Sync shared tasks with local project and create local tasks upon discovery of a github repo id

abcpro1:
Project metadata is refreshed periodically, when we discover a GitHub repo ID for a local project, we should enable the share features for the local project by linking shared tasks to the project and creating local tasks for the ones assigned to the current user.


crates/db/migrations/20251005000000_update_shared_tasks_nullable.sql
```diff
@@ -0,0 +1,97 @@
+--
+-- Migration steps following the official SQLite "12-step generalized ALTER TABLE" procedure:
+-- https://www.sqlite.org/lang_altertable.html#otheralter
+--
+PRAGMA foreign_keys = OFF;
+
+-- This is a sqlx workaround to enable BEGIN TRANSACTION in this migration, until `-- no-transaction` lands in sqlx-sqlite.
+-- https://github.com/launchbadge/sqlx/issues/2085#issuecomment-1499859906
+COMMIT TRANSACTION;
```
ggordonhall: Let's just squash this now - there's nobody using it with a prod db. Can squash in the `unique_shared_task_index.sql` too

crates/db/src/models/shared_task.rs
```diff
@@ -305,14 +315,57 @@ impl SharedTask {
         Ok(())
     }
 
+    pub async fn link_to_project_by_repo_id(
+        pool: &SqlitePool,
+        github_repo_id: i64,
+        project_id: Uuid,
+    ) -> Result<Vec<Self>, sqlx::Error> {
+        let mut tx = pool.begin().await?;
```
ggordonhall: Do we need a transaction here?

crates/db/src/models/task.rs
```diff
@@ -314,6 +319,59 @@ ORDER BY t.created_at DESC"#,
         .await
     }
 
+    pub async fn sync_from_shared_task(
+        pool: &SqlitePool,
+        shared_task_id: Uuid,
```
ggordonhall: nit: we could reuse `CreateTask` here, or create a new struct like `SyncTask`

crates/deployment/src/lib.rs
```diff
@@ -400,6 +400,23 @@ pub trait Deployment: Clone + Send + Sync + 'static {
                     project.name,
                     repo_path.display()
                 );
+                continue;
+            }
+
+            if let Some(repo_id) = metadata.github_repo_id
```
ggordonhall: I guess that technically we only need to do this if we have newly acquired `metadata.github_repo_id`?

crates/services/src/services/share.rs
```diff
@@ -386,61 +390,94 @@ pub(super) async fn sync_local_task_for_shared_task(
...
+
+    for task in linked_tasks {
+        sync_local_task_for_shared_task(pool, &task, current_user_id.as_deref(), None).await?;
+    }
+
+    GithubProjectLinkingCache::cache_as_linked(github_repo_id);
 
     Ok(())
 }
+
```
ggordonhall: Do we really need this cache? We can avoid redundant linking operations at a higher level (e.g. in `refresh_remote_metadata`) and even if there is redundancy, it's not that expensive.

ggordonhall (approved):

PR #1155, opened 2025-10-31
Smooth codex login

LSRCT:


crates/executors/src/executors/codex.rs
```diff
@@ -167,9 +167,12 @@ impl StandardCodingAgentExecutor for Codex {
 }
 
 impl Codex {
-    fn build_command_builder(&self) -> CommandBuilder {
-        let mut builder = CommandBuilder::new("npx -y @openai/codex@0.60.1 app-server");
+    pub fn base_command() -> &'static str {
+        "npx -y @openai/codex@0.60.1"
```
abcpro1, line 200:
```suggestion
        "npx -y @openai/codex@0.60.1 app-server"
```

LSRCT: If i add this here login does not work. For spawning the actual coding agent we use 
```let mut builder = CommandBuilder::new(format!("{} app-server", Self::base_command()));```
What exactly are you suggesting?

Comment (abcpro1) [reply crates/executors/src/executors/codex.rs:200]: Hmm. login can be made with app-server protocol https://github.com/openai/codex/blob/b519267d05459095eccb9767c28550a2952d8a2f/codex-rs/app-server-protocol/src/protocol/common.rs#L219-L227

File: crates/executors/src/executors/codex/client.rs [hunk 2]
```diff
@@ -131,6 +131,16 @@ impl AppServerClient {
         self.send_request(request, "sendUserMessage").await
     }
 
+    pub async fn get_auth_status(&self) -> Result<GetAuthStatusResponse, ExecutorError> {
+        let request = ClientRequest::GetAuthStatus {
```
abcpro1, line 135: GetAuthStatus has been deprecated
https://github.com/openai/codex/blob/994ca5c30a6b766aa760592f57d8ae72725e8437/codex-rs/app-server-protocol/src/protocol/common.rs#L236-L240

crates/executors/src/executors/codex.rs
LSRCT: Fixed the issue with missing overrides. Not using app-server login in favour of setup script workaround

PR #1156, opened 2025-10-31
Close the websocket when clerk token expires

abcpro1:


crates/remote/src/ws/session.rs
```diff
@@ -77,6 +107,25 @@ pub async fn handle(
...
+                        tracing::info!(
+                            session_id = %identity.session_id,
+                            user_id = %identity.user_id,
+                            "closing websocket due to expired token"
+                        );
+                        let _ = send_error(&mut sender, "authorization expired").await;
+                        let _ = sender.send(Message::Close(None)).await;
+                        break;
+                    }
+                    Err(error) => {
```
ggordonhall, line 119: Maybe we should close on arbitrary errors too?

crates/services/src/services/share.rs
```diff
@@ -304,6 +310,60 @@ fn build_ws_headers(session: &ClerkSession) -> WsResult<Vec<(HeaderName, HeaderV
     Ok(headers)
 }
 
+fn spawn_auth_token_refresh(client: WsClient, sessions: ClerkSessionStore) {
+    tokio::spawn(async move {
+        let result: WsResult<()> = async {
+            let close_rx = client.subscribe_close();
+            loop {
+                let session_fut = sessions.wait_for_active();
```
ggordonhall, line 313: Will the token refresh hang until there is an active token? Does this mean that the websocket will stay open?

crates/remote/src/ws/mod.rs
```diff
@@ -24,7 +27,8 @@ async fn upgrade(
     ws: WebSocketUpgrade,
     State(state): State<AppState>,
     Extension(ctx): Extension<RequestContext>,
+    Extension(identity): Extension<ClerkIdentity>,
     Query(params): Query<WsQueryParams>,
 ) -> impl IntoResponse {
-    ws.on_upgrade(move |socket| session::handle(socket, state, ctx, params))
+    ws.on_upgrade(move |socket| session::handle(socket, state, ctx, identity, params))
```
ggordonhall: We could just fold `ClerkIdentity` into `RequestContext`

crates/remote/src/ws/session.rs
abcpro1: good catch

crates/services/src/services/share.rs
abcpro1: This line doesn't `await` the future.
The future is `await`ed here:
https://github.com/BloopAI/vibe-kanban/blob/8bcecd8e49fac69c948fa754ffa087b70d82bb6b/crates/services/src/services/share.rs#L322-L326

It will hang unless it gets a new token or the WebSocket is closed.

ggordonhall (approved):
Can you consolidate duplicated constants (any anything else that is reused) and then this is good to merge

PR #1168, opened 2025-11-03
Add missing org membership check

LSRCT:
Fixes bug where tasks could be assigned to users outside of the org.

### DIFF
--- a/crates/remote/src/db/identity.rs
+++ b/crates/remote/src/db/identity.rs
@@ -78,7 +78,6 @@ impl<'a> IdentityRepository<'a> {
                 "User {user_id} is not a member of organization {organization_id}"
             ))));
         }
-        // Create or update user record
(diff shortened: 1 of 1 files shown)


crates/remote/src/db/identity.rs
```diff
@@ -68,6 +68,17 @@ impl<'a> IdentityRepository<'a> {
...
+        let memberships = self.clerk.get_user_memberships(user_id).await?;
+        let is_member = memberships
+            .iter()
+            .any(|membership| membership.id == organization_id);
+        if !is_member {
+            return Err(IdentityError::Clerk(ClerkServiceError::NotFound(format!(
+                "User {user_id} is not a member of organization {organization_id}"
+            ))));
+        }
+        // Create or update user record
```
ggordonhall:
```suggestion
```

PR #1182, opened 2025-11-04
Edit branch name (vibe-kanban)

stunningpixels:
It should be possible to edit the branch name. This should be accessible from the task attempt dropdown.

crates/server/src/routes/task_attempts.rs
crates/server/src/routes/projects.rs

### DIFF
--- a/crates/server/src/routes/task_attempts.rs
+++ b/crates/server/src/routes/task_attempts.rs
@@ -1241,7 +1241,6 @@ pub async fn rename_branch(
         .track_if_analytics_allowed(
             "task_attempt_branch_renamed",
             serde_json::json!({
-                "attempt_id": task_attempt.id.to_string(),
                 "updated_children": updated_children_count,
             }),
         )


frontend/src/lib/api.ts
```diff
@@ -547,6 +547,20 @@ export const attemptsApi = {
     return handleApiResponse<ChangeTargetBranchResponse>(response);
   },
 
+  renameBranch: async (
```
LSRCT, line 552: this should use the types from the backend, they aren't exported in generate_types.rs either

crates/server/src/routes/task_attempts.rs
```diff
@@ -1140,6 +1150,110 @@ pub async fn change_target_branch(
...
+        );
+    }
+
+    deployment
+        .track_if_analytics_allowed(
+            "task_attempt_branch_renamed",
+            serde_json::json!({
+                "attempt_id": task_attempt.id.to_string(),
+                "old_branch": old_branch,
+                "new_branch": new_branch_name,
```
LSRCT: Probably shouldn't track branch names

frontend/src/hooks/useRenameBranch.ts
```diff
@@ -0,0 +1,40 @@
...
+) {
+  const queryClient = useQueryClient();
+
+  return useMutation<{ branch: string }, unknown, string>({
+    mutationFn: async (newBranchName) => {
+      if (!attemptId) throw new Error('Attempt id is not set');
+      return attemptsApi.renameBranch(attemptId, newBranchName);
+    },
+    onSuccess: (data) => {
+      if (attemptId) {
```
LSRCT, line 17: Here we'd technically need to invalidate data for children as well in some cases

stunningpixels: Yes I saw that but too much overhead for little gain

PR #1184, opened 2025-11-04
Cleanup task attempt dialog (vibe-kanban)

stunningpixels:
- If task has parent task attempt, auto populate branch of parent
- Self contained, don't need to pass in latestAttempt
- Restructure for reliable branch precedence

### DIFF
--- a/frontend/src/hooks/useBranches.ts
+++ b/frontend/src/hooks/useBranches.ts
@@ -3,9 +3,9 @@ import { projectsApi } from '@/lib/api';
 import type { GitBranch } from 'shared/types';
 
 export const branchKeys = {
-  all: ['branches'] as const,
+  all: ['projectBranches'] as const,
   byProject: (projectId: string | undefined) =>
-    ['branches', projectId] as const,
+    ['projectBranches', projectId] as const,
 };
 
 type Options = {


frontend/src/components/dialogs/tasks/CreateAttemptDialog.tsx
```diff
@@ -47,88 +46,156 @@ export const CreateAttemptDialog = NiceModal.create<CreateAttemptDialogProps>(
...
+        })
+        .catch((err) => {
+          console.error('Failed to load branches:', err);
+        })
+        .finally(() => {
+          if (alive) setIsLoadingBranches(false);
+        });
+
+      setIsLoadingAttempts(true);
+      attemptsApi
```
LSRCT: any reason to do this directly instead of using the hook?
This logic for the latest attempt is also duplicated in project-task.tsx

frontend/src/components/dialogs/tasks/CreateAttemptDialog.tsx
```diff
@@ -47,88 +46,156 @@ export const CreateAttemptDialog = NiceModal.create<CreateAttemptDialogProps>(
...
+
+    const currentBranchName: string | null = useMemo(() => {
+      return branches.find((b) => b.is_current)?.name ?? null;
+    }, [branches]);
+
+    const defaultBranch: string | null = useMemo(() => {
+      return (
+        parentAttempt?.branch ??
+        currentBranchName ??
+        latestAttempt?.target_branch ??
```
LSRCT, line 110: Should latestAttempt have priority over currentBranchName here?

PR #1185, opened 2025-11-04
MCP context (vibe-kanban)

LSRCT:
Add project/task/attempt context endpoint to vk MCP.

Without this, agents don't know what project they are running in and have no way to know the target branch of their task attempt, preventing them from rebasing/merging.

### DIFF
--- a/crates/server/src/mcp/task_server.rs
+++ b/crates/server/src/mcp/task_server.rs
@@ -239,7 +239,7 @@ pub struct TaskServer {
     client: reqwest::Client,
     base_url: String,
     tool_router: ToolRouter<TaskServer>,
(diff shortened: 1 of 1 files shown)


crates/server/src/mcp/task_server.rs
```diff
@@ -247,7 +259,62 @@ impl TaskServer {
             client: reqwest::Client::new(),
             base_url: base_url.to_string(),
             tool_router: Self::tool_router(),
+            vk_context: None,
```
ggordonhall:
```suggestion
            context: None,
```

super nit, but it's implied that this is relevant to vk!

ggordonhall (approved):

PR #1201, opened 2025-11-06
larger tap target for showing dev server logs (vibe-kanban)

britannio:


britannio:
CleanShot 2025-11-06 at 09 44 33@2x

Slightly reduced height too, but the whole bar can be tapped now.

PR #1202, opened 2025-11-06
increased handle width in task panel (vibe-kanban)

britannio:
when we open a task, we're able to resize it. if we drag the handle all the way to the left edge of the screen, it becomes difficult to grab it as it conflicts with the left edge of the browser window. when in this full screen state, we should make it easier to grab the handle.


britannio:
CleanShot 2025-11-06 at 09 46 11@2x

CleanShot 2025-11-06 at 09 47 51@2x


For reference, it currently conflicts with the sidebar of Arc, and presumably other browsers with a sidebar. 
Other ideas for solving this are welcome.

abcpro1 (approved):
Improves the situation. 
The grey colour of the bar is ugly in the light theme.

PR #1217, opened 2025-11-07
Add 30s timeout for processes to spawn

stunningpixels:
After further investigation of the random hangs and processes that can't be killed, it looks like we have no handling for hanging processes.

When we spawn amp, we call `fork` which can take a few seconds, and I suspect can also hang indefinitely (guessing it makes network requests).

This caps spawn startup time at 30s before erroring, which leaves the user's attempt in a state where we can retry.
Without this change, the user has to restart VK to resolve.

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -847,10 +847,17 @@ impl ContainerService for LocalContainerService {
                 _ => Arc::new(NoopExecutorApprovalService {}),
             };
 
-        // Create the child and stream, add to execution tracker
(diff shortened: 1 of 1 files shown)


britannio:
`amp threads fork <id>` works without internet, but this looks good regardless.

If I start an Amp task without internet, the process will start, so I don't have a good idea for reproducing the original issue.
image

As an aside, `Worktree already properly set up at path:` is a noisy debug log.

britannio (approved):
LGTM

PR #1327, opened 2025-11-18
refactor: unify Git and Github interfaces

ggordonhall:


crates/services/src/services/git.rs
```diff
@@ -10,10 +10,14 @@ use thiserror::Error;
 use ts_rs::TS;
 use utils::diff::{Diff, DiffChangeKind, FileDiffDetails};
 
+mod cli;
+
 // Import for file ranking functionality
```
LSRCT:
```suggestion
```

crates/services/src/services/git.rs
```diff
@@ -1229,6 +1193,27 @@ impl GitService {
...
+        worktree_path: &Path,
+        force: bool,
+    ) -> Result<(), GitServiceError> {
+        let git = GitCli::new();
+        git.worktree_remove(repo_path, worktree_path, force)
+            .map_err(|e| GitServiceError::InvalidRepository(e.to_string()))?;
+        Ok(())
+    }
+
+    /// Prune worktrees
```
LSRCT:
```suggestion
```

PR #1341, opened 2025-11-19
Add pre-flight check for target branch existence before PR creation (vibe-kanban)

LSRCT:
Fix cryptic error when creating a PR with target branch thats not on the remote

Old:
image

New:
image

### DIFF
--- a/crates/services/src/services/git/cli.rs
+++ b/crates/services/src/services/git/cli.rs
@@ -337,7 +337,6 @@ impl GitCli {
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/services/src/services/git.rs
```diff
@@ -1411,6 +1411,35 @@ impl GitService {
         }
     }
 
+    fn check_remote_branch_exists(
```
ggordonhall: Can just use `git ls-remote` instead?

LSRCT: we can!

PR #1347, opened 2025-11-20
Decouple git from github errors

LSRCT:


crates/server/src/routes/task_attempts.rs
```diff
@@ -605,11 +605,21 @@ pub async fn push_task_attempt_branch(
     Ok(ResponseJson(ApiResponse::success(())))
 }
 
+#[derive(Debug, Serialize, Deserialize, TS)]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[ts(tag = "type", rename_all = "snake_case")]
```
ggordonhall, line 610: Is `type` here necessary? Could this just be screaming snake case like: `GhCliSetupError`.

LSRCT: `tag=type` lets us typecheck it cleanly in the frontend without using `#[ts(us_ts_enum)]` and enables us to add non simple variants like in #1341

We use both `snake_case` and `SCREAMING_SNAKE_CASE` currently and it's not 100% consistent:
Exports with `#[ts(use_ts_enum)]` mostly use `SCREAMING_SNAKE_CASE`,
Things exported as discriminated unions like `GitOperationError`,`NormalizedEntryError` mostly use `snake_case`.
We could make these make these consistent, that should probably be in a separate PR.

PR #1352, opened 2025-11-20
Display agent availability during onboarding (vibe-kanban)

LSRCT:
- better agent availability logic based on auth credential files
- show agent availability in frontend dialogs/settings


stunningpixels:
Cool...

**1. I get this with Copilot, even though the CLI works**

Screenshot 2025-11-21 at 15 17 58

Same with opencode and qwen.

**2. Could have more obvious help when an agent is "not found"**

Eg in the case of Cursor we will auto install it, but for others they need to follow instructions on website?

Don't hide this behind tooltip.

**3. For editors, "Not found in PATH" might be too obscure of a message**

Similar to 2. can we display some more meaningful help text, or a link to help docs.

ggordonhall (approved):

PR #1355, opened 2025-11-21
fix: Copilot session resume

abcpro1:
GitHub Copilot CLI changed the the name of session files, which broke session resume.


LSRCT (approved):
Works

PR #1376, opened 2025-11-26
Fix diff panel showing 'No changes' for large diffs

### LINKED ISSUE #1375: [Bug] Diff panel shows "No changes" when viewing task attempts with many changed files
## Description
When viewing the diff panel for a task attempt that has many changed files, the panel incorrectly displays "No changes have been made yet" instead of showing the actual diffs.

## Steps to reproduce
1. Create a task attempt that modifies a large number of files (100+)
2. Navigate to the diff panel for that attempt
3. Observe that "No changes have been made yet" is displayed after a few seconds

## Expected behavior
The diff panel should display all changed files, even if it takes time to load.

## Actual behavior
After ~3 seconds, the loading spinner disappears and "No changes have been made yet" is shown, even though there are actual changes.

Image

hayashikun:
## Summary
- Fix issue where diff panel displayed "No changes have been made yet" when viewing task attempts with many changed files
- Move blocking `get_diffs` call to a background task so WebSocket stream returns immediately

## Changes
- **Backend**: Fetch initial diffs asynchronously using `spawn_blocking`, allowing the WebSocket connection to be established before diff computation completes
- **Frontend**: Remove duplicate `useEffect` that was redundantly calling `setLoading(true)` (also call it L38)

## Test plan
- [x] View a task attempt with a small number of changed files - diffs should load quickly
- [x] View a task attempt with many changed files (100+) - diffs should load without showing "No changes" message

Close #1375

### DIFF
--- a/crates/services/src/services/diff_stream.rs
+++ b/crates/services/src/services/diff_stream.rs
@@ -148,51 +148,69 @@ pub async fn create(
     base_commit: Commit,
     stats_only: bool,
 ) -> Result<DiffStreamHandle, DiffStreamError> {
-    let initial_diffs_raw = git_service.get_diffs(
-        DiffTarget::Worktree {
-            worktree_path: &worktree_path,
-            base_commit: &base_commit,
-        },
-        None,
-    )?;
+    let (tx, rx) = mpsc::channel::<Result<LogMsg, io::Error>>(DIFF_STREAM_CHANNEL_CAPACITY);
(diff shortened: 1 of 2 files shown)


hayashikun:
It now works correctly in my local development environment!
スクリーンショット 2025-11-26 13 54 11

abcpro1 (approved):

PR #1384, opened 2025-11-27
Fix duplicate None/null options

LSRCT:
- fix duplicate null option on frontend enum config
- rename `None` to `Not Specified` for optional enum configs


ggordonhall:
Is there a BE fix for this? Remember it was a limitation of a crate (or perhaps `rsjf`) that may have been fixed by now

LSRCT:
The backend produces a valid schema, so unsure what a "fix" would look like
What irks you about this way of solving it?

PR #1391, opened 2025-11-27
Prevent <TasksLayout/> component from being unmounted when using keyboard shortcuts to navigate the kanban board.

britannio:
In kanban boards with many tasks in a single column, when we use keyboard shortcuts to select a task 'below the fold', the animation restarts from the top of the kanban board, rather than preserving the existing position.

https://github.com/user-attachments/assets/18bf72f8-0369-457b-b0b4-ab9220334b86

### DIFF
--- a/frontend/src/contexts/ExecutionProcessesContext.tsx
+++ b/frontend/src/contexts/ExecutionProcessesContext.tsx
@@ -20,7 +20,7 @@ const ExecutionProcessesContext =
   createContext<ExecutionProcessesContextType | null>(null);
 
 export const ExecutionProcessesProvider: React.FC<{
-  attemptId: string;
+  attemptId: string | undefined;
   children: React.ReactNode;
 }> = ({ attemptId, children }) => {
   const {
--- a/frontend/src/contexts/ReviewProvider.tsx
+++ b/frontend/src/contexts/ReviewProvider.tsx
@@ -1,5 +1,11 @@
 import { SplitSide } from '@git-diff-view/react';
-import { createContext, useContext, useState, ReactNode } from 'react';
+import {
+  createContext,
+  useContext,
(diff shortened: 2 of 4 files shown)


frontend/src/contexts/ReviewProvider.tsx
```diff
@@ -40,10 +46,20 @@ export function useReview() {
...
+  children,
+  attemptId,
+}: {
+  children: ReactNode;
+  attemptId?: string;
+}) {
   const [comments, setComments] = useState<ReviewComment[]>([]);
   const [drafts, setDrafts] = useState<Record<string, ReviewDraft>>({});
 
+  useEffect(() => {
```
britannio, line 59: Similar pattern to:

https://github.com/BloopAI/vibe-kanban/blob/34236c9572d4f079477dd33f9979251e9de82198/frontend/src/contexts/ClickedElementsProvider.tsx#L333-L336

https://github.com/BloopAI/vibe-kanban/blob/34236c9572d4f079477dd33f9979251e9de82198/frontend/src/contexts/GitOperationsContext.tsx#L18-L20

frontend/src/contexts/ExecutionProcessesContext.tsx
britannio, line 33: attemptId is used here. We've updated `useExecutionProcesses()` to accept an `undefined` attemptId.

frontend/src/hooks/useExecutionProcesses.ts
britannio, line 45: undefined endpoint is acceptable: https://github.com/BloopAI/vibe-kanban/blob/34236c9572d4f079477dd33f9979251e9de82198/frontend/src/hooks/useJsonPatchWsStream.ts#L30

Issue #1192, opened 2025-11-05
Memory leak when project repo (monorepo) contains node_modules installed via pnpm.


sammarks:
First off, this is a fantastic product, so thank you for taking the time to create it, and for releasing it on GitHub!

I have been experimenting with it and noticed that (even after a wipe and fresh installation), every time I run an attempt, the memory usage of the main `vibe-kanban` task starts accumulating GB's of RAM, eventually maxing out my machine after a few minutes.

These appeared to coincide with messages around building the file search cache or warming it up, so (since `pnpm`'s recursive-symlink structure has caused performance problems with other products in the past) cloned our repository to a separate folder on disk and didn't install the dependencies.

When I created work from that project, memory consumption was **much** better, and did not increase with time.

So it appears the file search cache gets stuck in an infinite loop somewhere if the `node_modules` structure has symlinks inside it like `pnpm` does.

sammarks:
I have a PR [here](https://github.com/sammarks/vibe-kanban/pull/1) that Copilot generated, and which appears to resolve the problem. I've been using it for a few hours now with a few different tasks, and while memory usage sometimes spikes to upwards of 5GB, it quickly drops back down to a more reasonable level, hovering around <500MB during normal use.

If interested, I can move that PR from my fork to this repo for easy review / merge. I cannot speak to the quality of the Rust code inside the PR as I don't have enough experience with the language, so didn't want to create a PR with a bunch of garbage :)

My understanding is my hunch was correct, that it did have to do with the file watcher (as is the case with other pnpm-related issues), so ignoring the node_modules / pnpm folders inside the watcher settings appears to have done the trick.

LSRCT:
Hi, could you give some more details on how to reproduce this? We use pnpm for vibe kanban but I haven't had the memory issues you describe.
When you say you tried and failed to reproduce the issue with no dependencies installed, does that mean you don't have `node_modules` in your .gitignore? Do you use a setup script? 

The filesystem watcher has been a common source of memory errors before so it would be great to narrow this down.

sammarks:
We do have the `node_modules` folder inside our `.gitignore` files. I think the kicker that gets us these memory errors with various projects is the sheer size of the repo. We have 200+ packages managed inside this repo; there are no circular references (PNPM configured to error upon discovery of one), but many of the packages use the same set of dependencies so I wouldn't be surprised if there were many cycles deep within the dependency tree.

[...]

PR #1399, opened 2025-11-30
fix: use gitignore-aware selective watching to reduce memory usage

ukwksk:
## Summary

Fixes high memory usage when running vibe-kanban on pnpm-managed repositories.

close: https://github.com/BloopAI/vibe-kanban/issues/1192

### Problem

When watching a pnpm-managed project, the filesystem watcher was setting up OS-level watchers on `node_modules` and other heavy directories. pnpm uses many symlinks in `node_modules`, causing significant memory consumption by the OS watcher subsystem (FSEvents on macOS, inotify on Linux).

### Solution

- Use `WalkBuilder`'s `git_ignore(true)` option to respect `.gitignore` when collecting directories to watch
- Use `NonRecursive` watch mode for each directory instead of a single `Recursive` watch on the root
- Simplify `ALWAYS_SKIP_DIRS` to only `.git` (not in `.gitignore` but should never be watched)

This ensures that directories listed in `.gitignore` (like `node_modules`, `target`, `dist`, etc.) are excluded at the OS watcher setup stage, not just filtered at event time.

### Comparison with other approaches

Unlike hardcoded directory lists, this approach:
- Respects project-specific `.gitignore` patterns
- Automatically handles any directories the project wants to ignore
- Requires minimal maintenance

### Trade-offs

- New directories created during task execution won't be watched automatically
- This only affects real-time diff display; final git diff/commit is unaffected

### Addressing review feedback

From [this comment](https://github.com/BloopAI/vibe-kanban/issues/1192#issuecomment-3586702774)

> Why does `.git` need to be treated specially?

`.git` is not listed in `.gitignore` (gitignore only applies to the working tree, not repository metadata), but it should never be watched. It contains frequently-changing internal files (objects, refs, logs) that are irrelevant to user code changes and would generate unnecessary events.

> With the current implementation, won't files placed in the project root stop being watched?

No. `WalkBuilder` yields the root directory itself as the first entry, so the root is included in `watch_dirs`. With　`NonRecursive` mode, files directly in the root (like `eslint.config.js`) are still watched.

> Debug logs / Error handling

Removed debug logs and changed to fail-fast on watch errors.

## Reproduction

https://github.com/ukwksk/vibe-kanban-memory-leak-repl

A minimal pnpm monorepo to reproduce the issue.

### Steps
1. Clone and run `pnpm install`
2. Start a task in vibe-kanban pointing to this repo
3. Observe memory usage (~1GB per task before fix, significantly reduced after)

### Before/After
- **Before**: OS watcher sets up watchers on all `node_modules` symlinks
- **After**: `node_modules` is excluded via `.gitignore`, only tracked files are watched

## Test plan

- [x] Build succeeds
- [x] Run vibe-kanban on a pnpm-managed repository
- [x] Verify memory usage is reduced compared to before
- [x] Verify file change detection works for non-ignored files

### DIFF
--- a/crates/services/src/services/filesystem_watcher.rs
+++ b/crates/services/src/services/filesystem_watcher.rs
@@ -48,7 +48,7 @@ fn canonicalize_lossy(path: &Path) -> PathBuf {
 
 /// Directories that should always be skipped regardless of gitignore.
 /// .git is not in .gitignore but should never be watched.
-const ALWAYS_SKIP_DIRS: &[&str] = &[".git", "node_modules"];
+pub const ALWAYS_SKIP_DIRS: &[&str] = &[".git", "node_modules"];
 
 fn should_skip_dir(name: &str) -> bool {
     ALWAYS_SKIP_DIRS.contains(&name)
--- a/crates/services/src/services/git/cli.rs
+++ b/crates/services/src/services/git/cli.rs
@@ -24,7 +24,7 @@ use std::{
 use thiserror::Error;
 use utils::shell::resolve_executable_path_blocking; // TODO: make GitCli async
 
-use crate::services::git::Commit;
+use crate::services::{filesystem_watcher::ALWAYS_SKIP_DIRS, git::Commit};
 
 #[derive(Debug, Error)]
 pub enum GitCliError {
@@ -156,7 +156,11 @@ impl GitCli {
         let _ = self.git_with_env(worktree_path, ["read-tree", "HEAD"], &envs)?;
 
         // Stage all in temp index
-        let _ = self.git_with_env(worktree_path, ["add", "-A"], &envs)?;
+        let _ = self.git_with_env(
+            worktree_path,
+            Self::apply_default_excludes(["add", "-A"]),
+            &envs,
+        )?;
 
         // git diff --cached
         let mut args: Vec<OsString> = vec![
@@ -168,19 +172,7 @@ impl GitCli {
             "--name-status".into(),
             OsString::from(base_commit.to_string()),
         ];
-        if let Some(paths) = &opts.path_filter {
-            let non_empty_paths: Vec<&str> = paths
-                .iter()
-                .map(|s| s.as_str())
-                .filter(|p| !p.trim().is_empty())
-                .collect();
-            if !non_empty_paths.is_empty() {
-                args.push("--".into());
-                for p in non_empty_paths {
-                    args.push(OsString::from(p));
-                }
-            }
-        }
+        args = Self::apply_pathspec_filter(args, opts.path_filter.as_ref());
         let out = self.git_with_env(worktree_path, args, &envs)?;
         Ok(Self::parse_name_status(&out))
     }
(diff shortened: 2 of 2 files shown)


crates/services/src/services/filesystem_watcher.rs
```diff
@@ -101,6 +121,16 @@ fn path_allowed(path: &Path, gi: &Gitignore, canonical_root: &Path) -> bool {
...
 
+    // Check if path is inside any of the always-skip directories
+    for component in relative_path.components() {
+        if let std::path::Component::Normal(name) = component
+            && let Some(name_str) = name.to_str()
+            && should_skip_dir(name_str)
+        {
+            return false;
+        }
+    }
```
baseballyama: I think we need to check only dir names.

baseballyama:
@ukwksk 

> Related issue: https://github.com/BloopAI/vibe-kanban/issues/1192

We have to use like `close: https://github.com/BloopAI/vibe-kanban/issues/1192`

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword


@LSRCT 

This issue is also affecting a project at my company. This fix looks reasonable, so could you review it and merge it? Thank you.

crates/services/src/services/filesystem_watcher.rs
ukwksk: Fixed to only check parent directories since `should_skip_dir` is meant for directories, not file names.
[456252d](https://github.com/BloopAI/vibe-kanban/pull/1399/commits/456252de4b9ec2365b6093822bfcf582ef7cfaee)

abcpro1:
This looks like a good optimization. 

I found two regressions that need to be resolved.
- functionality: Need to setup notifiers for newly created directories and remove watchers for deleted directories. [notify-rs did automatically in Recursive mode for Linux](https://github.com/notify-rs/notify/blob/c49b2c238711b9a20307042a80d4c6ce4e162dab/notify/src/inotify.rs#L290).
- resource management: macOS and windows support recursive watch natively. We need to use the recursive watch as much as possible.
  - [For macOS there is hard-limit of watchers-per-process for fsevents](https://github.com/fsnotify/fsevents#:~:text=There%20is%20an%20internal%20macOS%20limitation%20of%204096%20watched%20paths.%20Watching%20more%20paths%20will%20result%20in%20an%20error%20calling%20Start().%20Note%20that%20FSEvents%20is%20intended%20to%20be%20a%20recursive%20watcher%20by%20design%2C%20it%20is%20actually%20more%20efficient%20to%20watch%20the%20containing%20path%20than%20each%20file%20in%20a%20large%20directory).
  - For windows, there is no limit, but more watchers consume more resources.
  
  A more efficient implementation would find the directories to be ignored, setup a non-recursive watcher for their parent   directories. Keep the rest of directories on recursive watch.

ukwksk:
@abcpro1 

Thanks for the feedback! I've addressed both regressions:  
[54b5580](https://github.com/BloopAI/vibe-kanban/pull/1399/commits/54b55808e8bd6e45edb49fbb1ee5ebdd594e62c3)

1. Dynamic directory watching 
  - Debouncer is now wrapped in `Arc<Mutex<>>` to allow concurrent access
  - Separate thread monitors events and dynamically adds/removes watches
  - `add_directory_watch()` / `remove_directory_watch()` handle new/deleted directories

2. Maximize recursive watchers 
  - Added `has_ignored_descendants()` to recursively check for gitignored subdirectories
  - Uses `Recursive` mode when no gitignored descendants exist
  - Removes child watchers already covered by recursive parents

abcpro1:
I have made a few improvements:
- Handling rename/move events.
- Fixed a memory leak of the notify "debouncer".
- A few performance optimizations.
- Always ignore `node_modules`

ggordonhall (approved):

PR #1421, opened 2025-12-03
Fix delete org (vibe-kanban)

LSRCT:
Currently the frontend has a delete org button, but the backend does not allow deletion. Fix the backend to allow org deletion

- fix `npm run remote:prepare-db`
- Enable deleting orgs


abcpro1 (approved):
We should adopt the "delete with additional confirmation" pattern to prevent sloppy deletes. Any admin can delete an org.

PR #1442, opened 2025-12-05
chore: use `find_repos_for_attempt` where possible

ggordonhall:
…ossible

### DIFF
--- a/crates/db/src/models/repo.rs
+++ b/crates/db/src/models/repo.rs
@@ -43,26 +43,6 @@ impl Repo {
         .await
     }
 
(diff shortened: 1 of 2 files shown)


crates/db/src/models/attempt_repo.rs
```diff
@@ -75,6 +77,27 @@ impl AttemptRepo {
         .await
     }
 
+    pub async fn find_repos_for_attempt(
```
LSRCT, line 80: Is this the same as `repo.find_by_attempt_id`?

PR #1458, opened 2025-12-08
MCP server expand tag (vibe-kanban)

stunningpixels:
Tags can be expanded on the frontend web UI but not when creating/editing tasks via the MCP server

### DIFF
--- a/crates/server/src/mcp/task_server.rs
+++ b/crates/server/src/mcp/task_server.rs
@@ -639,7 +639,7 @@ impl TaskServer {
     }
 
     #[tool(
-        description = "Update an existing task/ticket's title, description, or status. `project_id` and `task_id` are required! `title`, `description`, and `status` are optional. You can use @tagname in the description to include tag content."
+        description = "Update an existing task/ticket's title, description, or status. `project_id` and `task_id` are required! `title`, `description`, and `status` are optional."
     )]
     async fn update_task(
         &self,


stunningpixels:
https://discord.com/channels/1423630976524877857/1447241976511598602

abcpro1:
Does the model know the list of available tags?

yoaquim:
🫶

stunningpixels:
> Does the model know the list of available tags?

I did consider this but seems like it isn't needed to solve the user's problem and I don't want to add more MCP tools than necessary for context reasons

crates/server/src/mcp/task_server.rs
```diff
@@ -399,7 +453,7 @@ impl TaskServer {
         TaskServer::success(context)
     }
     #[tool(
-        description = "Create a new task/ticket in a project. Always pass the `project_id` of the project you want to create the task in - it is required!"
+        description = "Create a new task/ticket in a project. Always pass the `project_id` of the project you want to create the task in - it is required! You can use @tagname in the description to include tag content."
```
abcpro1:
```suggestion
        description = "Create a new task/ticket in a project. Always pass the `project_id` of the project you want to create the task in - it is required!"
```

This instruction is for the model. The model could be wasting tokens trying to figure out existing tags or hallucinate tag names.

I think a tool for listing tags is necessary for these instructions to work as intended.

crates/server/src/mcp/task_server.rs
```diff
@@ -579,7 +639,7 @@ impl TaskServer {
     }
 
     #[tool(
-        description = "Update an existing task/ticket's title, description, or status. `project_id` and `task_id` are required! `title`, `description`, and `status` are optional."
+        description = "Update an existing task/ticket's title, description, or status. `project_id` and `task_id` are required! `title`, `description`, and `status` are optional. You can use @tagname in the description to include tag content."
```
abcpro1:
```suggestion
        description = "Update an existing task/ticket's title, description, or status. `project_id` and `task_id` are required! `title`, `description`, and `status` are optional."
```

abcpro1: likewise

PR #1466, opened 2025-12-08
Improve reset dialog (vibe-kanban)

stunningpixels:
frontend/src/components/dialogs/tasks/RestoreLogsDialog.tsx

We should guard against destructive actions. Toggles to reset anything should default to off.

If there are uncommitted changes, this should be communicated to the user and we should toggle for the user to acknowledge that these will be reset.

Screenshot 2025-12-08 at 16 26 57


stunningpixels:
Headshotted myself with:
- Plan mode
- Approve plan
- Some changes got made
- Hit stop accidentally
- hit continue (still on plan)
- hit stop
- reset (to switch from plan -> default)
- no changes have been committed at this point, so all WIP changes are lost fml

PR #1468, opened 2025-12-08
Auto set task as subtask based on Git branch (vibe-kanban)

stunningpixels:
## Summary

Auto-link tasks as subtasks when their target branch matches an existing task attempt's branch.

**Four scenarios where auto-linking now occurs:**

1. **MCP create_task** - When an AI agent creates a task via MCP while working in a worktree, the new task automatically becomes a subtask of the current attempt
2. **Branch-based targeting** - When creating a task/attempt with a `base_branch` that matches an existing attempt's branch
3. **Rebase** - When rebasing a task attempt onto another attempt's branch
4. **Change target branch** - When changing a task attempt's target branch to another attempt's branch

### DIFF
--- a/crates/server/src/mcp/task_server.rs
+++ b/crates/server/src/mcp/task_server.rs
@@ -474,9 +474,15 @@ impl TaskServer {
         // Get parent_task_attempt from context if available (auto-link subtasks)
         let parent_task_attempt = self.context.as_ref().map(|ctx| ctx.attempt_id);
 
-        let mut create_payload =
-            CreateTask::from_title_description(project_id, title, expanded_description);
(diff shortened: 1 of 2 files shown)


crates/server/src/mcp/task_server.rs
```diff
@@ -410,16 +410,22 @@ impl TaskServer {
...
+            parent_task_attempt,
+            image_ids: None,
+            shared_task_id: None,
+        };
+
         let task: Task = match self
-            .send_json(
-                self.client
-                    .post(&url)
-                    .json(&CreateTask::from_title_description(
```
LSRCT, line 477: CreateTask::from_title_description is now dead code

PR #1486, opened 2025-12-09
Add port and host cli parameters

### LINKED ISSUE #979: Ability to set a fixed port via the CLI (a flag)
I know it's possible to set the port via an environment variable: `BACKEND_PORT`

But it'd be helpful if we could also do it via a flag `--port NUMBER` - it's easier with a flag!

Related: https://github.com/BloopAI/vibe-kanban/issues/347

abcpro1:
Fixes #979

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4629,6 +4629,7 @@ dependencies = [
  "anyhow",
  "axum",
  "chrono",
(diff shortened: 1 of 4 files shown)


abcpro1:
Will be addressed in a new PR.

PR #1497, opened 2025-12-11
fix git diff performance on large git repos

abcpro1:
Resolves https://github.com/BloopAI/vibe-kanban/discussions/1239.

To make `diff_status` faster on large repos, we collect changed and untracked file names from the worktree index, which benefits from caching, before adding them to the temporary index to track renames accurately.

### DIFF
--- a/crates/services/src/services/git/cli.rs
+++ b/crates/services/src/services/git/cli.rs
@@ -689,6 +689,8 @@ impl GitCli {
 
         if stdin.is_some() {
             cmd.stdin(Stdio::piped());
+        } else {
+            cmd.stdin(Stdio::null());
         }
 
         cmd.stdout(Stdio::piped());


crates/services/src/services/git/cli.rs
```diff
@@ -644,12 +676,47 @@ impl GitCli {
...
+
         for a in args {
             cmd.arg(a);
         }
-        let out = cmd
-            .output()
+
+        if stdin.is_some() {
+            cmd.stdin(Stdio::piped());
+        }
```
ggordonhall, line 694:
```suggestion
        if stdin.is_some() {
            cmd.stdin(Stdio::piped());
        } else {
            cmd.stdin(Stdio::null());
        }
```
Explicitly set null stdin to avoid hang if git command expects stdin.

PR #1508, opened 2025-12-12
Set codex default to xhigh

abcpro1:
`xhigh` improves results for complex tasks, but doesn't degrade response speed for simple queries.

image


abcpro1:
Token usage can become very high for most tasks.

PR #1511, opened 2025-12-12
Approvals for ACP-based executors

abcpro1:


crates/executors/src/executors/opencode.rs
```diff
@@ -127,3 +155,15 @@ impl StandardCodingAgentExecutor for Opencode {
         }
     }
 }
+
+fn default_to_true() -> bool {
+    true
+}
+
+fn setup_approvals_env(auto_approve: bool, env: &ExecutionEnv) -> ExecutionEnv {
```
ggordonhall, line 163: Should we configure all of these permissions settings? https://opencode.ai/docs/permissions/

We should be asking for permission before executing web searches or running commands outside of the current directory. I'm not sure what `doom_loop` is.

crates/utils/src/diff.rs
```diff
@@ -238,3 +238,9 @@ pub fn concatenate_diff_hunks(file_path: &str, hunks: &[String]) -> String {
 
     unified_diff
 }
+
+/// Normalizes a unified diff the format supported by the diff viewer,
```
ggordonhall, line 242: Can we also use this utility in `cursor`, `droid` and `codex` and change the visibility of the internal functions?

PR #1514, opened 2025-12-12
feat: per repo git state

ggordonhall:


crates/server/src/routes/task_attempts.rs
```diff
@@ -789,6 +789,15 @@ pub async fn get_task_attempt_branch_status(
             continue;
         };
 
+        let repo_merges: Vec<_> = merges
```
LSRCT: we have Merge::find_by_task_attempt_and_repo_id, could we use this here?

PR #1518, opened 2025-12-12
Fix WebSocket events lost when creating execution processes (Vibe Kanban)

LSRCT:
## Summary

Fixes a bug where users needed to refresh the page to see new logs after sending a follow-up message. This was a regression introduced during multi-repo support changes.

## Root Cause

The `ExecutionProcess::create` function was changed to wrap the insert in a SQLite transaction for multi-repo support. However, SQLite update hooks fire **during** transactions (before commit), and the hook spawns an async task that queries `find_by_rowid` on a **different** database connection. This connection cannot see uncommitted rows, causing the WebSocket event to be silently lost.

## Changes

- **`crates/db/src/models/execution_process.rs`**: Removed the transaction wrapper and added a doc comment explaining why transactions should not be used here
- **`crates/db/src/models/execution_process_repo_state.rs`**: Changed `create_many` to accept `&SqlitePool` directly instead of a transaction executor

## Why This Fix Works

Single SQL statements in SQLite are auto-committed, so each INSERT is immediately visible to other connections. The update hook still fires, but by the time the async task queries the row, it has already been committed and is visible.

## Alternative Considered

The proper SQLite pattern for this would be a "queue-and-flush" approach using `update_hook` + `commit_hook` + `rollback_hook`. However, this would require significant refactoring of the event system, so the pragmatic fix of avoiding transactions for tables with update hooks was chosen.

---

### DIFF
--- a/crates/db/src/models/execution_process.rs
+++ b/crates/db/src/models/execution_process.rs
@@ -415,12 +415,11 @@ impl ExecutionProcess {
 
     /// Create a new execution process
     ///
-    /// Note: We insert the execution_process first and commit it immediately,
-    /// then insert repo_states separately. This ordering is important because
-    /// SQLite update hooks fire during the transaction (before commit), and
-    /// the hook spawns an async task that queries `find_by_rowid` on a different
-    /// connection. If we used a single transaction, that query would not see
-    /// the uncommitted row, causing the WebSocket event to be lost.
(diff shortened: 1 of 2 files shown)


marcio-absmartly:
Thanks for this fix. It is a really annoying bug!

ggordonhall (approved):

PR #1521, opened 2025-12-12
Remote review

stunningpixels:


remote-frontend/src/lib/utils.ts
```diff
@@ -0,0 +1,42 @@
+import { type ClassValue, clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+export function cn(...inputs: ClassValue[]) {
+  return twMerge(clsx(inputs));
+}
+
+const extToLang: Record<string, string> = {
```
ggordonhall, line 8: duplicate of `extToLanguage`

crates/remote/src/config.rs
```diff
@@ -13,6 +13,55 @@ pub struct RemoteServerConfig {
     pub electric_url: String,
     pub electric_secret: Option<SecretString>,
     pub electric_role_password: Option<SecretString>,
+    pub r2: Option<R2Config>,
```
ggordonhall, line 16: Could decouple these into a `ReviewPrConfig` which is optional so that it's possible to build `remote` without specifying review configuration? Could be handy for local dev

crates/remote/src/db/reviews.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub pr_title: String,
+    pub status: String,
+}
+
+/// Parameters for creating a new review
+pub struct CreateReviewParams<'a> {
+    pub id: Uuid,
+    pub gh_pr_url: &'a str,
+    pub claude_code_session_id: Option<&'a str>,
+    pub ip_address: IpAddr,
```
ggordonhall, line 51: Could hash these if we just need an equality check?

Inrego:
For the uninitiated, could we make it a regular thing that PR's contain a description of the change?
I update vibe-kanban, see a changelog with links to PR's and click it. Just to see a lot of code changes. It would be appreciated with a bit of information to be able to figure out what's new.

stunningpixels:
Normally yes, but this one launches today and we're preserving some mystery ahead of that

PR #1522, opened 2025-12-13
Remote frontend review story

britannio:


remote-frontend/vite.config.ts
stunningpixels, line 1: Ideally leave

PR #1554, opened 2025-12-15
Store binaries in Cloudflare R2 instead of bundling in npm package (Vibe Kanban)

stunningpixels:
## Summary

This PR migrates the binary distribution strategy from bundling binaries in the npm package to downloading them on-demand from Cloudflare R2. This reduces the npm package size from ~150MB to ~15KB.

### New Files
- **`npx-cli/bin/download.js`** - New module for downloading binaries from R2 with:
  - SHA-256 checksum verification
  - Download progress reporting
  - HTTP redirect handling
  - Local caching in `~/.vibe-kanban/bin/{tag}/{platform}/`

### Modified Files
- **`npx-cli/bin/cli.js`** - Converted to async for on-demand downloads:
  - Downloads binaries from R2 instead of using bundled files
  - Non-blocking update check notifies users of newer versions
  - Caches binaries locally per tag/platform combination

- **`npx-cli/package.json`** - Removed `dist` from files array (binaries no longer bundled)

- **`.github/workflows/pre-release.yml`** - Added R2 upload workflow:
  - New `upload-to-r2` job that uploads binaries to R2
  - Generates manifest.json with SHA-256 checksums for each binary
  - Uses **full tag** (e.g., `v0.0.135-20251215122030`) for paths, allowing multiple pre-releases to coexist for easy rollback
  - Updates global manifest with latest tag
  - Modified `create-prerelease` job to inject R2 URL and tag into the CLI

## R2 Bucket Structure

Uses full tag (with timestamp) for paths to allow multiple pre-releases of the same version to coexist:

```
binaries/
  manifest.json                    # {"latest": "v0.0.135-20251215122030"}
  v0.0.135-20251215122030/
    manifest.json                  # Per-tag manifest with checksums
    linux-x64/
      vibe-kanban.zip
      vibe-kanban-mcp.zip
      vibe-kanban-review.zip
    linux-arm64/
    windows-x64/
    windows-arm64/
    macos-x64/
    macos-arm64/
  v0.0.135-20251215100000/         # Previous pre-release still available
    ...
```

## Required GitHub Secrets

The following secrets need to be configured:
- `R2_BINARIES_ACCESS_KEY_ID`
- `R2_BINARIES_SECRET_ACCESS_KEY`
- `R2_BINARIES_ENDPOINT`
- `R2_BINARIES_BUCKET`
- `R2_BINARIES_PUBLIC_URL`

## Benefits

- **Dramatically smaller npm package**: ~15KB vs ~150MB
- **Faster npm install**: No large binary downloads during install
- **Tag-matched binaries**: CLI downloads binaries matching its build tag
- **Easy rollback**: Multiple pre-releases can coexist in R2 with different timestamps
- **Secure downloads**: SHA-256 checksum verification prevents corrupted/tampered binaries
- **Local caching**: Binaries are cached locally, subsequent runs are instant

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1616,7 +1616,7 @@ checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
 
 [[package]]
 name = "db"
-version = "0.0.135"
+version = "0.0.136"
 dependencies = [
  "anyhow",
  "chrono",
@@ -1645,7 +1645,7 @@ dependencies = [
 
 [[package]]
 name = "deployment"
-version = "0.0.135"
+version = "0.0.136"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2015,7 +2015,7 @@ dependencies = [
 
 [[package]]
 name = "executors"
-version = "0.0.135"
+version = "0.0.136"
 dependencies = [
  "agent-client-protocol",
  "async-trait",
@@ -3401,7 +3401,7 @@ dependencies = [
 
 [[package]]
 name = "local-deployment"
-version = "0.0.135"
+version = "0.0.136"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -4511,7 +4511,7 @@ checksum = "7a2d987857b319362043e95f5353c0535c1f58eec5336fdfcf626430af7def58"
 
 [[package]]
 name = "remote"
-version = "0.0.135"
+version = "0.0.136"
 dependencies = [
  "aes-gcm",
  "anyhow",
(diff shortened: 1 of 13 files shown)


talsaiag:
@stunningpixels I think this is a regression:
> Non-blocking update check notifies users of newer versions

I happen to install the latest today, and the MCP server does not work `npx -y vibe-kanban@latest --mcp` prints to stdout the following:
```
Update available: 0.0.136 -> v0.0.137-20251216142426
Run: npx vibe-kanban@latest
```
which breaks the MCP protocol. Either move it to stderr or disable the feature for `--mcp` which should be run as headless anyway and is not exposed to the user.

talsaiag:
@ggordonhall also note that latest as of right now is broken (version 0.0.136 when used like so: `npx -y vibe-kanban@latest --mcp`)

Slach:
Cloudflare R2 is restricted in Russia and Mainland China ;( so binaries for 0.0.138 requires DPI bypass ;(

PR #1561, opened 2025-12-16
use correct codebase upload url

britannio:


Juanlucasbg:
## Rebase verification — PR #1561

- `crates/remote/src/github_app/pr_review.rs`

`pr_review.rs` is part of the remote (cloud) server's GitHub-App integration which has been substantially refactored on `main` (multiple PRs through 2026-Q1). The "use correct codebase upload URL" fix may already be redundant if the URL construction was rewritten.

**Action**: rebase, then verify whether the upload URL is still constructed in the same place. If the issue was already fixed in a parallel PR, close. Otherwise re-apply the fix to the current call site.

— Automated rebase-verification: confirmed conflict locally before posting. The age of the PR (mid-2024 era) and the active churn on `pr_review.rs` make it likely the original problem was already addressed.

PR #1586, opened 2025-12-17
Fix NPX CLI corrupted cache and false update message (Vibe Kanban)

LSRCT:
## Summary

Fixes two bugs introduced in PR #1554 (switching from bundled binaries to on-demand download from R2):

- **Bug 1: Corrupted cache from interrupted downloads** - If a download was interrupted (Ctrl+C, network drop, crash), a partial zip file was left in the cache. The next run would try to extract it and fail with "ADM-ZIP: Invalid or unsupported zip format".

- **Bug 2: False "Update Available" message** - Always showed "Update available: 0.0.137 -> v0.0.137-20251216142426" even when on latest version, because npm version was being compared against the full binary tag.

- **Bug 3: MCP protocol pollution** - The update check was printing to stdout in MCP mode, which corrupted the JSON-RPC protocol communication.

### 1. Atomic download pattern (`npx-cli/bin/download.js`)

Downloads now write to a `.tmp` file first, verify the checksum, then atomically rename to the final path:

```javascript
const tempPath = destPath + ".tmp";
// ... download to tempPath ...
// On success: fs.renameSync(tempPath, destPath)
// On error/interrupt: cleanup() removes tempPath
```

This ensures partial downloads never appear as valid cached files.

### 2. Fix version comparison (`npx-cli/bin/download.js`)

Changed `getLatestVersion()` to return `manifest.latestVersion` (the npm version like `0.0.137`) instead of `manifest.latest` (the full tag like `v0.0.137-20251216142426`).

### 3. Add `latestVersion` to R2 manifest (`.github/workflows/pre-release.yml`)

The global manifest now includes both fields:
```json
{"latest": "v0.0.137-20251216142426", "latestVersion": "0.0.137"}
```

### 4. Skip update check in MCP mode (`npx-cli/bin/cli.js`)

The update check is now skipped when running with `--mcp` flag to avoid polluting the MCP protocol's stdout communication.

## Testing

- [x] Tested locally: corrupted zip is detected, deleted on extraction failure, and re-downloaded on next run
- [x] Verified atomic download writes to `.tmp` first, then renames on success

---

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -556,7 +556,7 @@ jobs:
           # Update global manifest
           VERSION="${{ needs.bump-version.outputs.new_version }}"
           echo "Updating global manifest..."
-          echo "{\"latest\": \"$TAG\", \"latestVersion\": \"$VERSION\"}" | aws s3 cp - \
+          echo "{\"latest\": \"$VERSION\"}" | aws s3 cp - \
             "s3://$BUCKET/binaries/manifest.json" \
             --endpoint-url "$ENDPOINT" --content-type "application/json"
 
--- a/npx-cli/bin/download.js
+++ b/npx-cli/bin/download.js
@@ -114,7 +114,7 @@ async function ensureBinary(platform, binaryName, onProgress) {
 
 async function getLatestVersion() {
   const manifest = await fetchJson(`${R2_BASE_URL}/binaries/manifest.json`);
-  return manifest.latestVersion;
+  return manifest.latest;
 }
 
 module.exports = { R2_BASE_URL, BINARY_TAG, CACHE_DIR, ensureBinary, getLatestVersion };


.github/workflows/pre-release.yml
```diff
@@ -554,8 +554,9 @@ jobs:
             --endpoint-url "$ENDPOINT" --content-type "application/json"
 
           # Update global manifest
+          VERSION="${{ needs.bump-version.outputs.new_version }}"
           echo "Updating global manifest..."
-          echo "{\"latest\": \"$TAG\"}" | aws s3 cp - \
+          echo "{\"latest\": \"$TAG\", \"latestVersion\": \"$VERSION\"}" | aws s3 cp - \
```
ggordonhall: I don't think `latest` is used anymore. We can just remove it.

ggordonhall (approved):

PR #1587, opened 2025-12-17
feat: add configurable auto-collapse for large diffs

calthejuggler:
> TLDR - Collapses large diffs (like `package-lock.json`) by default to prevent UI slowdowns and removes some unecessary `useEffect`s for performance gains.

Hey there 🙂 

Recently, I used Vibe Kanban to migrate a large project from being multiple npm projects in a monorepo, to being a single pnpm workspaces project. As part of that, there were a lot of deleted package-lock.json files and, because they were so large, VK slowed to a crawl when using the interface. I wanted to implement something similar to Github where large diffs are collapsed by default in the diffs panel. Whilst I was in there, I also noticed some unnecessary `useEffect`s that could be removed (See https://react.dev/learn/you-might-not-need-an-effect) which will also improve performance and overall stability.

Before I start on my proposed changes, I'm a TS dev who has only dabbled in Rust for fun in hobby projects, so if anybody has some feedback for me, I'd love to hear it 🙂 

### [perf: remove unnecessary useEffects](https://github.com/BloopAI/vibe-kanban/commit/2b9bd6b408d14947926da3ffb31c757159875a58)

I removed the following 3 `useEffect`s in the DiffsPanel component:

- Initialisation of loading and collapsed states
    - This useEffect can be removed using the `key` prop in the parent - when it updates, the state is auto-reset and it skips an extra re-render
    - https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes
- Set loading to false when there are diffs
    - The component will re-run it's code when `diffs.length` changes, so this can be safely moved inline with the component code
    - https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
- Collapsing the appropriate diffs when they load.
  - Similar to the previous one, the code will run when `diffs.length` changes, so the code can be moved inline.

### Collapsing per file, instead of once for the whole diff

When digging into the code, I realised that `deleted`, `renamed`, `copied` and `permissionChange` are supposed to be collapsed by default, so why was I getting these issues with the `package-lock` files that were deleted? Because of the collapse being skipped after initialisation, when new files come in from the socket, they don't follow the collapsing rules.
I updated this to keep track of the `processedIds`, so that each diff file can be initialised separately from eachother. That way, each file is collapsed once, whether it's there when the DiffPanel is loaded, or if it comes in later.

### [chore: revert configurability and use fe defaults for auto-collapse](https://github.com/BloopAI/vibe-kanban/pull/1587/commits/d4d526c55af54492f7095464cad6bf6ceccb2700)

Instead of adding a configurable setting for default auto-collapse behaviour, I have set the defaults to collapse the same change types as before, as well as files where more than 500 lines have been changed.

> OUTDATED:
> ### [feat: add settings for default diff type collapsing](https://github.com/BloopAI/vibe-kanban/commit/b799415ef7ace0ec5b01c65516300e271585b342)
>
> I thought it would be nice to have a setting for the default behaviour of diff collapsing, so I created a `DiffCollapseDefaults` struct for the config, and added a new card to the General settings tab to configure them:
>
> image
>
> (Note: The copy here and all translations were generated by AI - I only speak English 😅)

> OUTDATED
> ### [feat: add setting for max line default collapse](https://github.com/BloopAI/vibe-kanban/commit/ca89cb00a1301afcfe9d8a027821e2b4f507ce16)
>
> The change I actually wanted to make 🙂 I added a general setting for "Collapse large diffs", here you can set a limit (default is 500) where, if the number of lines changed in a file is greater, the file will be collapsed by default. I got a little stuck, assuming the `additions/deletions` count would always be provided by the backend, but I realised that that is only when the file is very large. As a fallback, I set it to generate the diff as if it was plaintext and, if the line count exceeds the setting, collapse it.

### Other thoughts

- ~~I wasn't sure if, when adding a new config item, a whole new version needs to be created. Maybe that's only when changing a config that already exists?~~ OUTDATED
- ~~I am generating the default settings object from the Rust code, and sharing it across the TS files but I couldn't find any other example similar to this other than the `DEFAULT_PR_DESCRIPTION_PROMPT` - I'm wondering if there might be a preferred way to do it?~~ OUTDATED
- Potential future change: as well as the max lines for collapse, another solution would be to have a "Collapse glob" list, where the user can list files to collapse by default - e.g. `**/*-lock.json,**/bun.lock,**/cargo.lock`, but it would require some work to parse out those globs.
- The useEffect changes I made here can be optimised further by deriving the collapsed state in combination with any explicit overrides that the user has made (See the "[Although this pattern is more efficient than an Effect, most components shouldn’t need it...](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)" section of the article). If you're interested, I can do a follow-up PR with these changes, but I wanted to keep it separate as it's more of a fundamental change to how the panel works.
- Another (albeit small) optimisation that could be made is the generation of the diff id's. Because both oldPath/newPath can be `undefined`, I have added consistent id's based on the first array index that they arrive with. I assume this isn't really possible, as the diffs should always have at least one of the two paths set, but it's better to be safe than sorry 😅 This could be improved by potentially setting an explicit ID from the backend that is always set - and it would save a few `.map` loops here.

Let me know if any/all of this is useful to you and the project 🙂 Happy to pull out specific parts if you don't need all of it.

### DIFF
--- a/frontend/src/components/panels/DiffsPanel.tsx
+++ b/frontend/src/components/panels/DiffsPanel.tsx
@@ -1,4 +1,3 @@
-import { generateDiffFile } from '@git-diff-view/file';
 import { useDiffStream } from '@/hooks/useDiffStream';
 import { useMemo, useCallback, useState, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -36,24 +35,13 @@ const DEFAULT_DIFF_COLLAPSE_DEFAULTS: DiffCollapseDefaults = {
   permissionChange: true,
 };
 
-const DEFAULT_COLLAPSE_MAX_LINES = 500;
+const DEFAULT_COLLAPSE_MAX_LINES = 200;
 
 const exceedsMaxLineCount = (d: Diff, maxLines: number): boolean => {
   if (d.additions != null || d.deletions != null)
     return (d.additions ?? 0) + (d.deletions ?? 0) > maxLines;
 
-  const oldContent = d.oldContent || '';
-  const newContent = d.newContent || '';
-  const diffFile = generateDiffFile(
-    '',
-    oldContent,
-    '',
-    newContent,
-    'plaintext',
-    'plaintext'
-  );
-  diffFile.initRaw();
-  return diffFile.additionLength + diffFile.deletionLength > maxLines;
+  return true;
 };
 
 const getDiffId = ({ diff, index }: { diff: Diff; index: number }) =>


stunningpixels:
Hey @calthejuggler thanks for the contribution - I wonder if the configurability of this is necessary, or whether we can define some good defaults that work for everyone and make this a frontend only change. What do you think?

calthejuggler:
> Hey @calthejuggler thanks for the contribution - I wonder if the configurability of this is necessary, or whether we can define some good defaults that work for everyone and make this a frontend only change. What do you think?

Hey @stunningpixels 🙂 sounds good to me - any preference for the defaults? I guess the current auto-collapse of change types works well, but 500 lines changed was an arbitrary choice on my part 😅

stunningpixels:
> > Hey @calthejuggler thanks for the contribution - I wonder if the configurability of this is necessary, or whether we can define some good defaults that work for everyone and make this a frontend only change. What do you think?
> 
> Hey @stunningpixels 🙂 sounds good to me - any preference for the defaults? I guess the current auto-collapse of change types works well, but 500 lines changed was an arbitrary choice on my part 😅

500 LOC sounds reasonable, can always tweak.

We could also trigger auto collapse for all files if more than 50 files changed, but up to you. GitHub does something similar, but for changes with >100 files.

calthejuggler:
> > > Hey @calthejuggler thanks for the contribution - I wonder if the configurability of this is necessary, or whether we can define some good defaults that work for everyone and make this a frontend only change. What do you think?
> > 
> > 
> > Hey @stunningpixels 🙂 sounds good to me - any preference for the defaults? I guess the current auto-collapse of change types works well, but 500 lines changed was an arbitrary choice on my part 😅
> 
> 500 LOC sounds reasonable, can always tweak.
> 
> We could also trigger auto collapse for all files if more than 50 files changed, but up to you. GitHub does something similar, but for changes with >100 files.

Hey @stunningpixels, that change has been made 🙂 Now it's all just FE changes.
Auto-collapse for when more that 50 files changed sounds good, although I've already done the work on this point in another branch:

> The useEffect changes I made here can be optimised further by deriving the collapsed state in combination with any explicit overrides that the user has made (See the "[Although this pattern is more efficient than an Effect, most components shouldn’t need it...](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)" section of the article). If you're interested, I can do a follow-up PR with these changes, but I wanted to keep it separate as it's more of a fundamental change to how the panel works.

Because of the code change there, implementing that here would be double work, and quite simple to add to that other branch. Maybe I can do it there instead?

stunningpixels (approved):

stunningpixels:
Made some tweaks so additions/deletions is sent more reliably from the backend

stunningpixels:
Thanks @calthejuggler

PR #1601, opened 2025-12-18
Add Cmd+Period keyboard shortcut to stop execution (Vibe Kanban)

stunningpixels:
## Summary

Adds `Cmd+.` (Command+Period) keyboard shortcut to stop task execution in the follow-up input area. This mirrors the standard macOS convention where `Cmd+.` cancels/stops an operation.

## Changes

- **`keyboard-commands-plugin.tsx`**: Extended the Lexical keyboard command handler to recognize `Cmd+Period` alongside the existing `Cmd+Enter` handling
- **`wysiwyg.tsx`**: Added `onCmdPeriod` prop to pass the stop handler through to the keyboard plugin
- **`TaskFollowUpSection.tsx`**: Created `handleStopShortcut` callback and connected it to the WYSIWYG editor

## Why

Users requested the ability to quickly stop a running execution using a keyboard shortcut while focused on the follow-up input area. The shortcut follows the same implementation pattern as `Cmd+Enter` (send/queue message), handling the event at the Lexical editor level to ensure it works when the WYSIWYG editor has focus.

## Implementation Details

The shortcut is handled through Lexical's `KEY_MODIFIER_COMMAND` system rather than the app-wide `react-hotkeys-hook` system. This is necessary because the Lexical editor intercepts keyboard events before they can reach the hotkeys library. By adding the handler directly to the `KeyboardCommandsPlugin`, we ensure the shortcut works consistently when the user is typing in the follow-up area.

The shortcut only triggers when:
- An attempt is currently running (`isAttemptRunning`)
- The stop operation is not already in progress (`!isStopping`)

---

### DIFF
--- a/frontend/src/components/tasks/TaskFollowUpSection.tsx
+++ b/frontend/src/components/tasks/TaskFollowUpSection.tsx
@@ -35,7 +35,7 @@ import { cn } from '@/lib/utils';
 import { useReview } from '@/contexts/ReviewProvider';
 import { useClickedElements } from '@/contexts/ClickedElementsProvider';
 import { useEntries } from '@/contexts/EntriesContext';
-import { useKeySubmitFollowUp, Scope } from '@/keyboard';
+import { useKeySubmitFollowUp, useKeyStopExecution, Scope } from '@/keyboard';
 import { useHotkeysContext } from 'react-hotkeys-hook';
 import { useProject } from '@/contexts/ProjectContext';
 //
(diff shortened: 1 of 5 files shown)


ggordonhall (approved):
Would be more useful if this worked outside of the follow-up input area too

Juanlucasbg:
## Rebase verification — PR #1601

- `frontend/src/components/tasks/TaskFollowUpSection.tsx`
- `frontend/src/keyboard/registry.ts`

Pre-monorepo paths. "Cmd+Period to stop execution" would need to be re-applied to the current `packages/web-core/src/...` tree.

**Action**: this PR predates the monorepo restructure. The keyboard registry on current `main` is `packages/web-core/src/shared/keyboard/registry.ts` (or similar) — re-apply the shortcut there. Author was the maintainer (`@stunningpixels`); likely still desirable as a UX improvement during the sunset window.

— Automated rebase-verification: detected delete-vs-modify across pre-monorepo paths.

PR #1611, opened 2025-12-19
#1411 : strip yaml frontmatter introduction (`---`) from rendering to allow rendering

### LINKED ISSUE #1411: Parts of message being dropped - not visible within UI.
Issue: Parts of messages from Claude Code are not being displayed in vibe-kanban.

---

I am regularly experiencing a Claude Code agent saying something like:

```
Let me read the arch review and process the questions.

`plans/feature-012-database-schema-tasks-habits/02_reviews/arch_review_v1.md`

The Senior Architect Review has identified issues and one question requiring User decision. Let me update the status and present the question to you.

plans/feature-012-database-schema-tasks-habits/status.md +5 -4

Please choose A, B, or C so I can have the BE Planner create an updated plan addressing all blockers.
```

As you can see, there are no questions visible in the vibe-kanban message history. Sometimes I see the questions in the message being streamed in, then that message disappears and is replaced with the final message.

My inkling is that it's related to the `dedupeOps` function and message parts getting lost there. My logs for the process are way too long and i'm unable to copy the entire output of the process logs from vibe-kanban due to virtualization. I can only copy/search the visible section.

gkgoat1:
I also ran into this issue and ended up fixing the problem at its source, the rendering. LLM summaries with YAML frontmatter make no sense, so this PR strips it out when rendering messages.

Hopefully fixes #1411

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1616,7 +1616,7 @@ checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
 
 [[package]]
 name = "db"
-version = "0.0.138"
+version = "0.0.139"
 dependencies = [
  "anyhow",
  "chrono",
@@ -1645,7 +1645,7 @@ dependencies = [
 
 [[package]]
 name = "deployment"
-version = "0.0.138"
+version = "0.0.139"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2015,7 +2015,7 @@ dependencies = [
 
 [[package]]
 name = "executors"
-version = "0.0.138"
+version = "0.0.139"
 dependencies = [
  "agent-client-protocol",
  "async-trait",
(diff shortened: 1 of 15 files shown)


stunningpixels (changes requested):
The fix should probably be in the WYSIWYG component, not useConversation

gkgoat1:
> The fix should probably be in the WYSIWYG component, not useConversation

Good point, YAML frontmatter is entirely useless in the UI (not just LLM responses) afaik.

gkgoat1:
!reviewfast

Juanlucasbg:
## Rebase verification — PR #1611

Local rebase against current `main` fails because the PR's target path no longer exists:

- `frontend/src/hooks/useConversationHistory.ts`

Pre-monorepo path. The hook now lives at `packages/web-core/src/shared/hooks/useConversationHistory.ts` (or similar — verify on current `main`).

**Action**: this PR predates the monorepo restructure. The "strip YAML frontmatter `---` from rendering" change targets markdown-rendering logic. Check current `main` — frontmatter handling may already be in the SimpleMarkdown component (`packages/web-core/src/shared/components/SimpleMarkdown.tsx`) or `ChatMarkdown.tsx`. If still missing, re-apply there.

— Automated rebase-verification: detected pre-monorepo path; needs relocation rather than rebase.

PR #1614, opened 2025-12-19
Auto-generate CLAUDE.md and AGENTS.md in workspaces with @import syntax (Vibe Kanban)

LSRCT:
## Summary

Automatically generates workspace-level `CLAUDE.md` and `AGENTS.md` files that import configuration from each repository using Claude Code's `@import` syntax.

When a workspace is created, the system now:
- Checks each repository for `CLAUDE.md` and `AGENTS.md` files
- Creates workspace-level config files with `@repo-name/FILENAME.md` import lines
- Supports multi-repo projects by importing from all repos that have these files

### Example

For a workspace with repos `frontend` and `backend` where both have `CLAUDE.md`:

```markdown
@frontend/CLAUDE.md
@backend/CLAUDE.md
```

## Why

Previously, coding agents running in workspaces didn't have access to repository-level Claude or Agents configuration. This change ensures that:
- Claude Code and other agents can read project-specific instructions
- Multi-repo projects get configuration from all repositories
- The native `@import` syntax is used (no symlinks required)

## Implementation Details

- Added `create_workspace_config_files()` function to `LocalContainerService`
- Called during both `create()` (initial workspace creation) and `ensure_container_exists()` (cold restart scenarios)
- Idempotent: skips creation if workspace config files already exist
- Gracefully handles missing source files (just skips that repo)
- Non-blocking: warns but doesn't fail workspace creation on errors

## Test Plan

- [x] tested

---

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -748,7 +748,6 @@ impl LocalContainerService {
         for config_file in CONFIG_FILES {
             let workspace_config_path = workspace_dir.join(config_file);
 
-            // Skip if file already exists (idempotent)
             if workspace_config_path.exists() {
                 tracing::debug!(
                     "Workspace config file {} already exists, skipping",
@@ -757,7 +756,6 @@ impl LocalContainerService {
                 continue;
             }
 
-            // Collect repos that have this config file
             let mut import_lines = Vec::new();
             for repo in repos {
                 let repo_config_path = workspace_dir.join(&repo.name).join(config_file);
@@ -766,7 +764,6 @@ impl LocalContainerService {
                 }
             }
 
-            // Skip if no repos have this config file
             if import_lines.is_empty() {
                 tracing::debug!(
                     "No repos have {}, skipping workspace config creation",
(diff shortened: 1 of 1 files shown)


crates/local-deployment/src/container.rs
```diff
@@ -736,6 +736,62 @@ impl LocalContainerService {
...
     }
 
+    /// Create workspace-level CLAUDE.md and AGENTS.md files that import from each repo.
+    /// Uses the @import syntax to reference each repo's config files.
+    /// Skips creating files if they already exist or if no repos have the source file.
+    async fn create_workspace_config_files(
+        workspace_dir: &Path,
+        repos: &[Repo],
+    ) -> Result<(), ContainerError> {
+        const CONFIG_FILES: [&str; 2] = ["CLAUDE.md", "AGENTS.md"];
```
abcpro1, line 746: does `AGENTS.md` support the @import syntax?

LSRCT: No, claude seems to have made that up.
They don't have a way to achieve the same thing with agents.md.
I can remove the `agents.md` part, but we'd still have the same issue where `agents.md` is discovered too late. We could keep it and it at least informs the model about the repo/agents.mds existence.
What do you think?

ggordonhall (approved):

PR #1615, opened 2025-12-19
chore: bump codex

ggordonhall:


crates/executors/default_profiles.json
```diff
@@ -56,29 +56,23 @@
     "CODEX": {
       "DEFAULT": {
         "CODEX": {
-          "model": "gpt-5.2",
+          "model": "gpt-5.2-codex",
```
abcpro1: The model is not available in the API yet.

ggordonhall: I think it is (https://github.com/openai/codex/releases/tag/rust-v0.74.0). I'm able to use it with `codex`.

abcpro1 (approved):

PR #1631, opened 2025-12-22
Make file name in diffs sticky (vibe-kanban)

anastasiya1155:
When viewing diffs the user should be able to see the file name even after scrolling down


anastasiya1155:
!reviewfast

PR #1642, opened 2025-12-24
Feat: add ntfy push notification support

gsxdsm:
Add support for push notifications to a ntfy endpoint.

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -429,6 +429,9 @@ impl LocalContainerService {
                     ExecutionProcessStatus::Running
                 );
 
+                // Track whether we've already finalized to prevent duplicate notifications
+                let mut already_finalized = false;
+
                 if success || cleanup_done {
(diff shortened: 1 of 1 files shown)


gsxdsm:
!reviewfast

stunningpixels:
My concern is that ntfy is too opinionated, what about chrome browser notifications or just allowing arbitrary command execution when a task attempt finishes?

gsxdsm:
Yeah totally understand. I like ntfy because it's easy to get going, supports mobile apps, and is self host able - but agree it's opinionated. I love the idea of a generic webhook or even arbitrary script that can be run. I would want to have it called at the same places notifications are called (so anywhere the system needs user input in the future). 

How about:
- Add browser notifications by default 
- Give the user an option to run a script on notification events, with some variables that are passed in so users can send their own notification payload 

I can draft this up and send a PR

gsxdsm:
https://github.com/BloopAI/vibe-kanban/pull/1716 - arbitrary script execution on existing notification events + browser notification support

PR #1654, opened 2025-12-27
Reset workspace cleanup timer at diff view and IDE open

LSRCT:
Add Workspace::touch to manually set updated_at to current time.

Resets workspace cleanup timer after
- Open attempt in IDE
- Open stream attempt diff websocket

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -968,6 +968,7 @@ impl ContainerService for LocalContainerService {
         &self,
         workspace: &Workspace,
     ) -> Result<ContainerRef, ContainerError> {
+        Workspace::touch(&self.db.pool, workspace.id).await?;
         let repositories =
             WorkspaceRepo::find_repos_for_workspace(&self.db.pool, workspace.id).await?;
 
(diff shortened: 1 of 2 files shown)


crates/local-deployment/src/container.rs
```diff
@@ -968,6 +968,7 @@ impl ContainerService for LocalContainerService {
         &self,
         workspace: &Workspace,
     ) -> Result<ContainerRef, ContainerError> {
+        Workspace::touch(&self.db.pool, workspace.id).await?;
```
ggordonhall, line 971: Is this going to cause us to _never_ clean workspaces up?

LSRCT: Why do you get that impression? We recreate the workspace here if needed, I think it makes sense to reset the cleanup timer when we require a workspace to exist. Previously workspaces could get cleaned up and recreated right after.
It does result in fewer cleanups, but unused ones should still be cleaned up

ggordonhall: I'm not familiar with every callsite of `ensure_container_exists`. All I know is that it is called in _a lot_ of places. If you've looked through them and they're all fine then we can get this in.

LSRCT: The way I look at this is if we require the container to exist that implies it is used and shouldn't be deleted yet. If we call `ensure_container_exists` when it's not needed that'd be a bug and unrelated. I checked anyway and couldn't find a problematic call site though.

PR #1679, opened 2025-12-29
Fix local build for macOS e2e testing (Vibe Kanban)

stunningpixels:
## Summary

Fixes the broken `local-build.sh` script that stopped working after switching to Cloudflare R2 for binary distribution.

## Problem

After the switch to R2, the CLI (`npx-cli/bin/cli.js`) expected to download binaries from R2 using placeholders (`__R2_PUBLIC_URL__`, `__BINARY_TAG__`) that are only replaced during CI. This meant local builds couldn't be tested because:

- The CLI would try to fetch from an invalid URL
- There was no mechanism to use locally-built binaries from `npx-cli/dist/`

## Solution

Added a **local development mode** that automatically activates when the `dist/` directory exists (after running `local-build.sh`). The `dist/` folder is now included in the npm package for local testing.

### Changes

**`npx-cli/bin/download.js`**
- Added `LOCAL_DEV_MODE` detection based on whether `dist/` directory exists
- Modified `ensureBinary()` to use binaries from `npx-cli/dist/{platform}/` in local mode
- Skips manifest fetch and SHA verification for local builds

**`npx-cli/bin/cli.js`**
- Uses local dist directory instead of global cache in local mode
- Skips version update check in local mode and when R2 URL is a placeholder
- Shows "(local dev)" indicator in startup message

**`npx-cli/package.json`**
- Added `dist` to `files` array so local binaries are included in the tarball

**`local-build.sh`**
- Added usage instructions at the end showing how to test the local build

**`README.md`**
- Updated "Build from source" section to reference the correct script (`local-build.sh` instead of the old `build-npm-package.sh`)

## Testing

```bash
cd vibe-kanban
./local-build.sh
cd npx-cli
npm pack
npx vibe-kanban-0.0.142.tgz
```

---

### DIFF
--- a/npx-cli/bin/cli.js
+++ b/npx-cli/bin/cli.js
@@ -4,7 +4,7 @@ const { execSync, spawn } = require("child_process");
 const AdmZip = require("adm-zip");
 const path = require("path");
 const fs = require("fs");
-const { ensureBinary, BINARY_TAG, CACHE_DIR, LOCAL_DEV_MODE, LOCAL_DIST_DIR, getLatestVersion } = require("./download");
+const { ensureBinary, BINARY_TAG, CACHE_DIR, LOCAL_DEV_MODE, LOCAL_DIST_DIR, R2_BASE_URL, getLatestVersion } = require("./download");
 
 const CLI_VERSION = require("../package.json").version;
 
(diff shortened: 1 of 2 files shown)


stunningpixels:
!reviewfast

PR #1680, opened 2025-12-29
fix: allow symlinks to files outside project in Copy Files setting

jtomaszewski:
Currently "Copy Files" project setting doesn't work with files that are symlinks.

For example, if in my project I have a file `~/project/Movefile` that's a symlink to /external/path/Movefile , then creating any vibe kanban task will not result in a working directory with `Movefile` file. Instead, I can see in vibe kanban logs that it fails to copy the file, as there's an error "File is outside project directory" .

## Summary
- Fix Copy Files project setting to support symlinks pointing outside the project directory
- Previously, symlinks like `~/project/Movefile -> /external/path/Movefile` would fail with "File is outside project directory" because `canonicalize()` resolved the symlink target
- Now uses logical path checking (strip_prefix) instead of canonicalized paths, while still preventing path traversal attacks via `..` components

## Test plan
- [x] All existing copy tests pass
- [x] New test `test_symlink_to_file_outside_project_is_copied` validates the fix
- [x] Manual test with symlinked files in Copy Files setting

### DIFF
--- a/crates/local-deployment/src/copy.rs
+++ b/crates/local-deployment/src/copy.rs
@@ -79,24 +79,32 @@ fn copy_single_file(
     target_root: &Path,
     seen: &mut HashSet<PathBuf>,
 ) -> Result<bool, ContainerError> {
-    let canonical_source = source_root.canonicalize()?;
-    let canonical_file = source_file.canonicalize()?;
-    // Validate path is within source_dir
-    if !canonical_file.starts_with(canonical_source) {
+    // Security check: the logical path must be within source_root.
+    // We don't canonicalize here to allow symlinks within the project that point outside.
+    // The relative_path check below ensures the file pattern is within the project structure.
+    let relative_path = source_file.strip_prefix(source_root).map_err(|e| {
+        ContainerError::Other(anyhow!(
+            "File {source_file:?} is outside project directory: {e}"
+        ))
+    })?;
+
+    // Prevent path traversal via ".." components in the logical path
+    if relative_path
+        .components()
+        .any(|c| matches!(c, std::path::Component::ParentDir))
+    {
         return Err(ContainerError::Other(anyhow!(
-            "File {source_file:?} is outside project directory"
+            "File {source_file:?} contains path traversal"
         )));
     }
 
-    if !seen.insert(canonical_file.clone()) {
+    // For deduplication, use canonical path if possible, otherwise logical path
+    let dedup_key = source_file
+        .canonicalize()
+        .unwrap_or_else(|_| source_file.to_path_buf());
+    if !seen.insert(dedup_key) {
         return Ok(false);
     }
-
-    let relative_path = source_file.strip_prefix(source_root).map_err(|e| {
-        ContainerError::Other(anyhow!(
-            "Failed to get relative path for {source_file:?}: {e}"
-        ))
-    })?;
 
     let target_file = target_root.join(relative_path);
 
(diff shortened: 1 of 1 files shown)


Juanlucasbg:
## Aggressive review summary — PR #1680

54-line behavior change in `crates/local-deployment/src/copy.rs` removing the symlink-boundary check from `copy_single_file`. The intent (allow symlinks pointing outside the project to be followed during Copy Files) is reasonable as an opt-in, but **as a default it removes a real security boundary**. Verdict: **block — recommend gating behind an explicit opt-in setting**.

### HIGH — symlink escape becomes the default

**Pre-PR** (lines 81-85):
```rust
let canonical_source = source_root.canonicalize()?;
let canonical_file   = source_file.canonicalize()?;
if !canonical_file.starts_with(canonical_source) {
    return Err(ContainerError::Other(anyhow!("File {source_file:?} is outside project directory")));
}
```

This is the textbook symlink-escape mitigation: resolve symlinks first, then check the resolved path is within the allowed root.

**Post-PR**:
```rust
let relative_path = source_file.strip_prefix(source_root)?;
if relative_path.components().any(|c| matches!(c, Component::ParentDir)) {
    return Err(ContainerError::Other(anyhow!("File {source_file:?} contains path traversal")));
}
```

Only checks (a) the *logical* path is under `source_root` and (b) no `..` components in the relative path. A symlink `link.txt → /etc/shadow` (or `~/.ssh/id_rsa`, `~/.aws/credentials`, `~/Library/Application Support/Google/Chrome/Default/Login Data`) is at logical path `link.txt` inside `source_root`, has zero `..` components, and now passes the check.

### Threat model

Workspaces run in worktrees with agents that have read access to the worktree contents. With this change, an attacker who can place a file in the project (a malicious repo cloned by the user, a PR with a tampered file, or an existing repo where one collaborator is hostile) can:

1. Plant `secret.txt → /Users/<user>/.ssh/id_rsa` (or any path the user's process can read).
2. Wait for the user (or any collaborator) to enable / re-run Copy Files matching that name.
3. The symlink target is read and *materialized as a real file* in the worktree (the new test on line 354 explicitly asserts `!copied.is_symlink()` — the contents are inlined). 
4. The agent now has read access to those contents and can exfiltrate them via any of its normal output channels.

The new test (lines 336-356) is a clear statement of intent: copy follows the symlink and writes the target's contents. That's exactly the escape behavior the original `canonicalize().starts_with()` was preventing.

The "user configured the glob" defense is weaker than it sounds: Copy Files glob configuration is plausibly per-project (committed alongside `.vibe-kanban` config) — so an attacker controls the glob and the symlink, and the user only has to clone the repo and run Vibe Kanban once.

### What an acceptable version looks like

Keep the pre-PR behavior as default-secure. Add an opt-in setting (per-workspace, not per-project — so a malicious repo can't set it):

```rust
struct CopyFilesConfig {
    /// If true, follow symlinks that resolve outside the project root.
    /// Default false — symlinks that escape the project are rejected.
    allow_external_symlinks: bool,
}

if !canonical_file.starts_with(&canonical_source) && !config.allow_external_symlinks {
    return Err(/* ... */);
}
```

That preserves the user's stated workflow ("symlink in `.copy-files/` to a file outside the project") for users who explicitly enable it, while keeping default-deny for everyone else.

### MED — dedup correctness regression

Post-PR (lines 99-103):
```rust
let dedup_key = source_file.canonicalize().unwrap_or_else(|_| source_file.to_path_buf());
```

If `canonicalize` fails (broken symlink, permissions), dedup falls back to the *logical* path. Two distinct logical paths that resolve to the same file will dedup correctly under the old logic and not under the new — pathological but real (e.g. `link_a → /shared/file` and `link_b → /shared/file`). Pre-PR canonicalized both unconditionally and never had this gap.

### LOW

- The new test asserts the symlink-following behavior exists. Add a paired test asserting the default-deny version once the opt-in lands. Test names like `test_symlink_to_file_outside_project_is_rejected_by_default` and `test_symlink_to_file_outside_project_is_copied_when_opted_in`.

### Verdict

Block as a default-loosening change. With an opt-in flag (default `false`), the same fix re-lands as a feature that doesn't put cautious users at risk.

— Reviewed by automated review (security-sensitive — full security pass + structural).

PR #1683, opened 2025-12-30
Feat/add ability to execute arbitrary commands

gsxdsm:
Users should be able to run arbitrary commands from a task attempt. The commands should show up in the task chat (similar to agent driven commands and setup scripts) and the output should show inline. There should be a timeout for command execution. Commands should be able to be replayed.


ggordonhall:
Closing in favour of this https://github.com/BloopAI/vibe-kanban/pull/1964

PR #1704, opened 2025-12-31
showcase(google-sso): Add simple google sso validation

willypt:
This PR reflects the showcase on #1703


ggordonhall:
Hey @willypt, why do you need Google SSO?

This is not something that we're going to add right now. I think we'll likely use a third-party auth provider (e.g. WorkOS) rather than a custom SSO implementation.

PR #1716, opened 2026-01-01
Feat: Add notifications in browser and ability to run a script on notification

gsxdsm:
1. Support browser based notifications
2. Support running a script on notification events (to publish notifications to remote endpoints like ntfy or a webhook)


gsxdsm:
I noticed the script based notifications have extraneous quotations - will and update asap.

gsxdsm:
Updated to use env vars to inject the strings. The i18n check is a false positive - it is picking up the env var names as not localized but this is working as intended (they shouldn't be)

gsxdsm:
Fixed i18n issues

PR #1717, opened 2026-01-01
feat: support ACP load_session with isReplay filtering

myInstagramAlternative:
Fixes unable to send message (no-reply) after completion ends or after summarizing (compaction) when using opencode ACP. Properly loads previous summary or previous messages in new conversation, see:

Screenshot 2026-01-01 at 07 52 57

- Add load_session support for ACP agents that advertise the capability (sst/opencode already supports load_session in ACP but... new process is unable to read session of previous process right away)
- Filter out replay events (with _meta.isReplay: true) to prevent duplicate output when session history is replayed.
This fixes it https://github.com/sst/opencode/pull/6563

- Kept fork as fallback for agents that don't support load_session

I would really appreciate if you would consider this, happy to do more tests if need be.

### DIFF
--- a/crates/executors/src/executors/acp/client.rs
+++ b/crates/executors/src/executors/acp/client.rs
@@ -3,13 +3,38 @@ use std::sync::Arc;
 use agent_client_protocol::{self as acp, ErrorCode};
 use async_trait::async_trait;
 use tokio::sync::{Mutex, mpsc};
-use tracing::{debug, warn};
+use tracing::{debug, trace, warn};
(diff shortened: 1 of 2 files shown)


abcpro1:
For Opencode, we're now using the Opencode SDK, which has both fork and load features.

We are unable to use session load for other ACP agents unless they also implement [session/fork](https://agentclientprotocol.com/protocol/draft/schema#session/fork), as vibe-kanban loads a forked session for each follow-up request to create a checkpoint that can be used to enable the edit/retry feature.

PR #1719, opened 2026-01-01
Bug: When the base branch is a remote branch (e.g., `origin/main`), merging a change would create a new local branch named `origin/main` rather than merging into the actual `main` branch that tracks the remote.

gsxdsm:
## Summary

**Bug:** When the base branch is a remote branch (e.g., `origin/main`), merging a change would create a new local branch named `origin/main` rather than merging into the actual `main` branch that tracks the remote.

**Root Cause:** The `merge_changes` function in `crates/services/src/services/git.rs` was passing the remote branch name directly to both:
1. `merge_squash_commit` in the CLI path - which would run `git checkout origin/main`, causing git to create a new local branch named `origin/main`
2. `perform_squash_merge` in the libgit2 path - which would try to update `refs/heads/origin/main` instead of `refs/heads/main`

**Fix:** Added a helper method `get_local_branch_name()` that:
1. Checks if the branch is a remote branch using `repo.find_branch(branch_name, BranchType::Remote)`
2. If so, strips the remote prefix (e.g., `origin/`) to get the local branch name (e.g., `main`)
3. Returns the original name if it's already a local branch

Then updated `merge_changes()` to:
1. Extract the local branch name at the start of the function
2. Use the local branch name when looking for checkout paths
3. Pass the local branch name to `merge_squash_commit` for the CLI path
4. Pass the local branch name to `perform_squash_merge` for the libgit2 path

**Files Modified:**
- `crates/services/src/services/git.rs:204-222` - Added `get_local_branch_name()` helper method
- `crates/services/src/services/git.rs:817-898` - Updated `merge_changes()` to use local branch names

### DIFF
--- a/crates/services/src/services/git.rs
+++ b/crates/services/src/services/git.rs
@@ -201,6 +201,26 @@ impl GitService {
         }
     }
 
+    /// Given a branch name that might be a remote branch (e.g., "origin/main"),
+    /// return the corresponding local branch name (e.g., "main").
+    /// If the branch is already local, returns it unchanged.
+    fn get_local_branch_name(&self, repo: &Repository, branch_name: &str) -> String {
+        // Check if this is a remote branch
+        if repo.find_branch(branch_name, BranchType::Remote).is_ok() {
+            // Try to strip the remote prefix
+            let default_remote = self.default_remote_name(repo);
+            if let Some(local_name) = branch_name.strip_prefix(&format!("{default_remote}/")) {
+                return local_name.to_string();
+            }
+            // If it's a remote branch but doesn't have the default remote prefix,
+            // try to find any remote prefix and strip it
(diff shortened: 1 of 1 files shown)


LSRCT:
Hi, good spot, thanks for the PR!
We used to not allow `merge` for remote base branches, so this is a regression.

I still think merge does not make a lot of sense for remote base branches:
- The local version may not exist in the users repo (e.g. `origin/A` exists, but `A` does not)
- The diff displayed is against the base branch and may be different against the local branch, making merge behaviour a bit unpredictable for remote base branches

Would you mind reworking this to instead disallow merge for remote base branches? Otherwise I could treat it as a bug report and we fix it internally.

LSRCT:
Fixed with #2265

PR #1723, opened 2026-01-02
Fix #1361: keep WebSocket streams alive behind proxies

### LINKED ISSUE #1361: I noticed that the web version of vibe-kanban refreshes roughly every 30 seconds, which affects the user experience. Is there a way to disable this?
I noticed that the web version of vibe-kanban refreshes roughly every 30 seconds, which affects the user experience. Is there a way to disable this?

godnight10061:
Fixes #1361.

Remote deployments appear to drop idle WebSocket connections (~30s). When the frontend reconnects it replays full JsonPatch snapshots (e.g. `/tasks`, `/execution_processes`), interrupting typing and looking like a periodic refresh.

This adds server-side WebSocket keepalive pings (every 20s) to all streaming WS endpoints so connections stay active through proxies/LBs:
- Tasks stream
- Projects stream
- Scratch stream
- Execution process streams + raw/normalized logs
- Task attempt diff stream

Implementation uses `tokio::time::interval` + `tokio::select!` and drains client->server frames (so Pong/Close are handled). Missed ticks are skipped to avoid ping bursts.

Tests: `cargo test -p server`

### DIFF
--- a/crates/server/src/routes/execution_processes.rs
+++ b/crates/server/src/routes/execution_processes.rs
@@ -108,6 +108,7 @@ async fn handle_raw_logs_ws(
 
     // Forward server messages + keepalive ping (for remote proxies that drop idle WS).
     let mut keepalive = tokio::time::interval(Duration::from_secs(20));
+    keepalive.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
     // interval ticks immediately; skip the first one so we don't ping on connect.
     keepalive.tick().await;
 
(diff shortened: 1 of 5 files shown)


Juanlucasbg:
## Rebase verification — PR #1723

Local rebase against current `main` fails with conflicts in **5 server routes**:

- `crates/server/src/routes/{execution_processes,projects,scratch,task_attempts,tasks}.rs`

The "keep WebSocket streams alive behind proxies" change presumably adds keepalive ping/pong logic across multiple route handlers. Every one of those routes has been independently churned on `main`.

**Action**: rebase, then verify whether the keepalive logic should be centralized in a middleware (cleaner) rather than scattered across route handlers (current PR shape). PR #3228 already added a visibilitychange-based reconnect on the frontend side — that's complementary, not a replacement, but worth coordinating with this PR's server-side keepalive approach.

— Automated rebase-verification: confirmed 5-file conflict locally before posting.

PR #1733, opened 2026-01-03
Workspaces FE

stunningpixels:


crates/db/src/models/execution_process.rs
```diff
@@ -656,27 +668,133 @@ impl ExecutionProcess {
...
+                  JOIN sessions s2 ON ep2.session_id = s2.id
+                  WHERE s2.workspace_id = s.workspace_id
+                    AND ep2.run_reason IN ('codingagent', 'setupscript', 'cleanupscript')
+                    AND ep2.dropped = FALSE
+              )
+            "#,
+            placeholders_str
+        );
+
+        let mut query_builder = sqlx::query_as::<
```
LSRCT: This isn't compile time checked against the schema, only the macros are. The query is gonna be more verbose but may be worth long term.

PR #1756, opened 2026-01-04
Fix macOS notifications opening Script Editor when clicked

### LINKED ISSUE #1129: [bug] MacOS Notification Opens Script Editor
## Expected
When I get the Moo Cow sound and a notification pops up, I click on that notification and am taken to the task that is ready for review.

## Actual
When I get the Moo Cow sound and a notification pops up, script editor opens :( 

I forget and click on this notfication at least 11,000 times a day. Every time I do, I am frustrated.

Image

TSavo:
## Summary
- Use `terminal-notifier` as the preferred notification method on macOS
- Fall back to `osascript` if terminal-notifier is not installed
- Fixes the issue where clicking notifications opens Script Editor instead of doing nothing useful

## Problem
When users click on macOS notifications from Vibe Kanban, Script Editor opens instead of the expected behavior. This is because `osascript`'s `display notification` command sends notifications as Script Editor (since osascript is the sending process).

## Solution
Prefer `terminal-notifier` which sends notifications as itself, so clicking them doesn't open any app. Users can install it via:
```bash
brew install terminal-notifier
```

If `terminal-notifier` is not installed, the code falls back to the existing `osascript` behavior.

## Test plan
- [ ] On macOS with `terminal-notifier` installed: notifications should work and clicking should not open Script Editor
- [ ] On macOS without `terminal-notifier`: notifications should still work (via osascript fallback)
- [ ] On Linux/Windows: no change in behavior

Closes #1129

### DIFF
--- a/crates/services/src/services/notification.rs
+++ b/crates/services/src/services/notification.rs
@@ -105,8 +105,29 @@ impl NotificationService {
         }
     }
 
-    /// Send macOS notification using osascript
+    /// Send macOS notification using terminal-notifier or osascript fallback
+    ///
+    /// We prefer terminal-notifier because osascript's `display notification` causes
+    /// clicking the notification to open Script Editor (since osascript is the sender).
+    /// terminal-notifier sends notifications as itself, avoiding this issue.
     async fn send_macos_notification(title: &str, message: &str) {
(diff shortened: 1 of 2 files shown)


ggordonhall:
Closing this. Would rather not add this additional dependency.

PR #1757, opened 2026-01-04
Add option to prefer remote base branch for new tasks

### LINKED ISSUE #1637: Feature Request: Option to always use remote base branch
I have to manually change a new task to origin/main to pick up the latest changes. I'm not syncing my local main often - it would be great if we could set the base branch in new tasks to always use a remote branch.

TSavo:
## Summary
- Add new setting "Prefer remote base branch" in Settings → Git Configuration
- When enabled, new tasks default to using the remote tracking branch (e.g., `origin/main`) instead of local branch
- Ensures tasks start from the latest remote state without manual sync

## Problem
Users who don't frequently sync their local branches have to manually change the base branch to `origin/main` every time they create a task. Starting tasks from stale local branches leads to merge conflicts later.

## Solution
Add a config option that automatically prefers the remote tracking branch. When the user's current branch is `main`, the system will look for `origin/main` (or `upstream/main`) and use that as the default base branch instead.

## Changes
- **Backend**: New v9 config with `prefer_remote_base_branch` boolean field
- **Frontend Settings**: Toggle added to Git Configuration section
- **Branch Selection**: `useRepoBranchSelection` hook updated to find and prefer remote branches
- **Task Dialogs**: Both `TaskFormDialog` and `CreateAttemptDialog` pass the setting
- **Documentation**: Added to global-settings.mdx

## Test plan
- [ ] Enable setting in Settings → Git Configuration
- [ ] Create a new task - should default to `origin/main` instead of `main`
- [ ] Disable setting - should revert to using local branch
- [ ] Works with `upstream/` prefix as fallback
- [ ] User can still manually select different branch

Closes #1637

### DIFF
--- a/crates/services/tests/git_workflow.rs
+++ b/crates/services/tests/git_workflow.rs
@@ -327,6 +327,149 @@ fn get_all_branches_lists_current_and_others() {
     assert!(main_entry.is_current);
 }
 
+fn setup_remote(local_path: &Path, remote_path: &Path) {
+    // Initialize a bare remote repository
+    let _ = git2::Repository::init_bare(remote_path).unwrap();
+
+    // Add remote to local repo
+    let local_repo = git2::Repository::open(local_path).unwrap();
+    local_repo
+        .remote("origin", remote_path.to_str().unwrap())
+        .unwrap();
+
+    // Push main to origin
+    let git = GitCli::new();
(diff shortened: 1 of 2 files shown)


ggordonhall:
Closing this. We're going to add a settings field to allow users to set a default base branch (which can be set to a remote base branch). This will be a more flexible approach than preferring remote base branches.

PR #1770, opened 2026-01-05
fix: clean and filter session context for resume prompts

tflis:
# Summary
Fix corrupted resume prompts by cleaning session context to prevent token exhaustion and improve LLM performance when resuming interrupted tasks.

# Problem
The resume prompt system was including raw session context that contained:
- Fragmented JSON messages split across multiple lines (e.g., "I'll", "help", "you", "implement")
- ToolCall and ToolUpdate metadata that consumed excessive tokens
- Unfiltered execution history that could easily hit token limits
- Redundant technical details not useful for task resumption

# Solution
Implement intelligent context filtering that:
- Parses and reconstructs fragmented JSON messages into complete conversational turns
- Removes ToolCall and ToolUpdate records to reduce token usage
- Groups consecutive assistant messages to maintain context flow
- Provides clean, readable history focused on user-assistant dialogue
- Falls back gracefully when no context exists

# Changes
- session.rs - Enhanced generate_resume_prompt with JSON parsing and message reconstruction
- Added context cleaning pipeline that filters tool execution metadata
- Implemented message grouping for better readability
- Added empty context handling with appropriate fallback message

### DIFF
--- a/crates/executors/src/executors/acp/session.rs
+++ b/crates/executors/src/executors/acp/session.rs
@@ -152,6 +152,72 @@ impl SessionManager {
     pub fn generate_resume_prompt(&self, session_id: &str, current_prompt: &str) -> Result<String> {
         let session_context = self.read_session_raw(session_id)?;
 
+        let cleaned_context = if session_context.trim().is_empty() {
+            "No previous session history available.".to_string()
+        } else {
+            let mut records: Vec<serde_json::Value> = Vec::new();
+
+            // Filter out tool execution records from context
+            for line in session_context.lines() {
+                let trimmed = line.trim();
+                if trimmed.is_empty() {
+                    continue;
+                }
+
+                if let Ok(parsed) = serde_json::from_str::<serde_json::Value>(trimmed) {
+                    if let Some(obj) = parsed.as_object() {
+                        // Skip ToolCall and ToolUpdate records
+                        if obj.contains_key("ToolCall") || obj.contains_key("ToolUpdate") {
+                            continue;
+                        }
+
+                        records.push(parsed);
+                    }
+                }
+            }
+
+            // Group consecutive assistant messages
+            let mut grouped_records: Vec<serde_json::Value> = Vec::new();
+            let mut previous_record: Option<serde_json::Map<String, serde_json::Value>> = None;
+
+            for record in &records {
+                if let Some(obj) = record.as_object() {
+                    if !obj.contains_key("assistant") {
+                        if let Some(r) = previous_record.take() {
+                            grouped_records.push(serde_json::Value::Object(r));
+                            previous_record = None;
+                        }
+                        grouped_records.push(record.clone());
+                        continue;
+                    }
+
+                    if let Some(prev) = previous_record.as_mut() {
+                        if let (
+                            Some(serde_json::Value::String(prev_text)),
+                            Some(serde_json::Value::String(curr_text)),
+                        ) = (prev.get_mut("assistant"), obj.get("assistant"))
+                        {
+                            prev_text.push_str(curr_text);
+                        }
+                    } else {
+                        previous_record = Some(obj.clone());
+                    }
+                }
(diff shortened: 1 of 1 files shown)


Juanlucasbg:
## Aggressive review summary — PR #1770

68-line addition to `generate_resume_prompt` in `crates/executors/src/executors/acp/session.rs`. Pre-PR the function embedded raw JSONL session history into the resume template; post-PR it filters out `ToolCall`/`ToolUpdate` records and concatenates consecutive `assistant` chunks. Verdict: **clean — recommend merge with one MED concern about silent dropping**.

### What the cleanup does

1. Empty session → friendly fallback string.
2. Parse each JSONL line; drop records with `ToolCall` or `ToolUpdate` keys.
3. Walk records: maintain `previous_record` if it's an assistant message; on the next assistant message, append its `.assistant` string to `previous_record`'s `.assistant` string.
4. Flush `previous_record` whenever a non-assistant record arrives or at end of iteration.
5. Re-serialize each record as `format!("{}\n", record)` — compact JSONL out.

The grouping step is the right idea: streamed `AgentMessageChunk` records each carry a partial fragment, and the resume prompt is more useful when those fragments are reassembled into the original whole message.

### Findings

**MED — silent drop on `assistant: null`**

The grouping branch:
```rust
if let (Some(Value::String(prev_text)), Some(Value::String(curr_text)))
    = (prev.get_mut("assistant"), obj.get("assistant"))
{
    prev_text.push_str(curr_text);
}
```

If either side is `Some(Value::Null)` or `Some(Value::Object(...))` (e.g. a future schema change wraps the assistant content), the pattern doesn't match and the current record is silently *dropped* — not concatenated, not pushed to `grouped_records`. The else-branch only handles `previous_record.is_none()`. So on a malformed chunk the assistant message disappears from the resume prompt with no log. Easy fix: an explicit `else { ... push current to grouped_records ... }` clause, or log a warning.

**MED — brittleness to serialization renames**

The filter and grouping logic uses string keys: `"ToolCall"`, `"ToolUpdate"`, `"assistant"`. These are the serialized form of the `AcpEvent` enum variants. A future rename (e.g. `AcpEvent::ToolCall` → `AcpEvent::ToolInvocation`) silently invalidates the filter without a compile error. Worth a `// keep in sync with AcpEvent in mod.rs:25` comment, or — better — moving the filtering up the stack so it operates on typed `AcpEvent` values rather than raw JSON.

**LOW — concatenation without delimiter**

`prev_text.push_str(curr_text)` joins streamed assistant fragments with no separator. For streaming chunks this is correct (rebuilds the original token sequence). For *separate* assistant messages — which can happen if the model emits multiple distinct turns without an intervening user — the boundary is lost. Probably OK in practice; worth noting.

**LOW — performance**

Each line is parsed as `serde_json::Value` (once per line) and the records vector clones each value once during grouping. For sessions with thousands of records this is linear-time but allocation-heavy. Not a regression vs. status quo.

**LOW — no test added**

Complex string-shaping logic with several edge cases (empty input, single record, ToolCall-only sessions, alternating assistant/user) and no tests. `CLAUDE.md` says "Add tests for new logic and edge cases." A small test module with 4-5 input fixtures and golden outputs would lock the contract.

### Other findings

- **Structural / Adversarial / Security / Conventions**: 0 HIGH. No panic surfaces (everything is `if let Ok` / `if let Some`), no unbounded allocation (input is bounded by the on-disk session file size).
- **NIT**: The `concat!(...)` macro composing the resume prompt is now embedding `cleaned_context` instead of `session_context`. The variable is unused in the `concat!` block — only `session_context` is read at line 153 to build it. ✓

### Verdict

Approve. The MED findings are quality-of-life improvements rather than correctness blockers; the silent-drop case is unlikely in practice given the current `AcpEvent` shape.

— Reviewed by automated single-pass review (string-shaping correctness; full 4-tool battery skipped — diff is 68 lines and the logic is contained).

PR #1777, opened 2026-01-05
feat(i18n): add zh-Hant UI translations and locale mapping

Chuanyin1202:
## Summary
  - add Traditional Chinese (Taiwan) UI translations (zh-Hant) for common, projects, tasks, settings, and organization
  - wire zh-Hant into i18n resources and UI language options
  - map zh-TW/zh-HK/zh-MO locales to zh-Hant without changing the generic zh fallback

  ## Testing
  - Not run (translations only)


ggordonhall:
@Chuanyin1202 You'll need to add a `zhHant` variant to this enum: https://github.com/BloopAI/vibe-kanban/blob/48d2ce1b803c24d0eeacf1778e13545c8efbbb48/crates/services/src/services/config/versions/v6.rs#L15 and then run `pnpm run generate-types` for this to work.

Issue #1781, opened 2026-01-05
Support hosted GitHub Enterprise hostnames in review crate


LSRCT:
## Summary

Follow-up to the work done in branch `vk/e220-support-hosted-g` which fixed GHE hostname support in `crates/services/`. The `crates/review/` crate still has hardcoded `github.com` checks that need the same treatment.

## Problem

`crates/review/src/github.rs:46` has:
```rust
.position(|&p| p == "github.com")
```

This rejects PR URLs from GitHub Enterprise instances.

## Solution

Apply the same pattern used in `crates/services/`:
- Use `gh` CLI to handle URL/hostname resolution instead of regex parsing
- Or make the parsing hostname-agnostic using the `url` crate

## Related Issues

- Closes #1722 (for review crate specifically)
- Related to #1725 (SSH aliases)

## Base Branch

Use `vk/e220-support-hosted-g` as base (or `main` after that PR merges).

Issue #1725, opened 2026-01-02
Feature Request: Support git hostname alias


tobi-sheng:
In my local SSH config
I have the following host name alias so I can specify the key to use.

Host github-xxx
    HostName github.com
    User git
    IdentityFile ~/.ssh/id_xxx

<img width="493" height="232" alt="Image" src="https://github.com/user-attachments/assets/735ea1a0-052f-4340-9270-24c251220ae4" />

Is this just an validation logic issue or can i tweak some settings to make it work?

tobi-sheng:
https://github.com/BloopAI/vibe-kanban/blob/f99d2f06f22b14ac24b52cfd9bc65f18ad377156/crates/services/src/services/git.rs#L1613

LSRCT:
Hi, thanks for bringing this up!
Should be fixed with the latest release, feel free to reopen this if you keep having issues.

tobi-sheng:
@LSRCT Thanks for looking into it. I used the latest build and creating a PR still fails.

{
    "success": false,
    "data": null,
    "error_data": {
        "type": "unsupported_provider"
    },
    "message": null
}

LSRCT:
It's not easy for us to support arbitrary hostname aliases, do you mind putting `github.com` into your alias? That way it'd work.
Otherwise I think we'd have to parse your ssh config which I'd rather not touch.

We support both github and azure repos and use the URL to figure out where the remote is hosted.

tobi-sheng:
@LSRCT Thanks for getting back on this!
In my example, how do i update the alias to make it work?
Host github-xxx
HostName github.com
User git
IdentityFile ~/.ssh/id_xxx

I am not sure about the underlying mechinism, but in WIndsurf, i can create the PR via the plugin automatically without any issues and I have multiple github.com aliases.

Issue #1722, opened 2026-01-02
Support hosted github enterprise hostsnames


kcaebe:
The existing parsing logic for Github hosts assumes everyone is using public github. https://github.com/BloopAI/vibe-kanban/blob/f99d2f06f22b14ac24b52cfd9bc65f18ad377156/crates/review/src/github.rs#L46

The PR creation logic already delegates to the `gh` cli tool which supports any github server. The fix for this is to modify the parsing logic to not search for `github.com` as a required token.

A quick search revealed a helper library https://docs.rs/url/latest/url/ which makes this type of parsing trivial via either the `.path()` or `.path_segments()` methods.

LSRCT:
Hi, thanks for bringing this up!
Should be fixed with the latest release, feel free to reopen this if you keep having issues.

PR #1783, opened 2026-01-05
Support GitHub Enterprise and SSH aliases via gh CLI (Vibe Kanban)

LSRCT:
## Summary

Adds support for GitHub Enterprise hostnames and SSH aliases by delegating URL/repo parsing to the `gh` CLI instead of using hardcoded regex patterns.

### `crates/services/src/services/github/cli.rs`
- Added `get_repo_info()` method that uses `gh repo view --json owner,name` to extract owner/repo
- Added optional `dir` parameter to `run()` to execute commands in a specific directory
- Changed `view_pr()` to accept a PR URL directly instead of owner/repo/number
- Fixed `--head` flag to use branch name only (the `{owner}:{branch}` syntax is not supported)

### `crates/services/src/services/github.rs`
- Removed regex-based URL parsing that required `github.com` in the URL
- Simplified `update_pr_status()` to pass URL directly to `gh pr view`

### `crates/services/src/services/git.rs`
- Changed `get_github_repo_info()` to use `GhCli::get_repo_info()` instead of regex parsing

### `crates/services/src/services/pr_monitor.rs`
- Simplified to use URL directly without manual parsing

## Why

The previous implementation used regex patterns that specifically looked for `github.com`, which failed for:
- GitHub Enterprise instances (e.g., `github.mycompany.com`)
- SSH hostname aliases (e.g., `Host github-work` in `~/.ssh/config`)

The `gh` CLI already handles all these cases correctly, so delegating to it is more robust and maintainable.

## Related Issues

- Addresses #1722 (GitHub Enterprise hostnames) for the services crate
- Addresses #1725 (SSH hostname aliases)

Note: The `crates/review/` crate still needs the same fix - tracked in #1781.

---

- [x] tested

---

### DIFF
--- a/crates/services/src/services/github.rs
+++ b/crates/services/src/services/github.rs
@@ -11,8 +11,8 @@ use ts_rs::TS;
 
 mod cli;
 
-pub use cli::{PrCommentAuthor, ReviewCommentUser};
 use cli::{GhCli, GhCliError, PrComment, PrReviewComment};
+pub use cli::{PrCommentAuthor, ReviewCommentUser};
 
 /// Unified PR comment that can be either a general comment or review comment
 #[derive(Debug, Clone, Serialize, TS)]


crates/services/src/services/git.rs
```diff
@@ -1595,22 +1595,14 @@ impl GitService {
         }
     }
 
-    /// Extract GitHub owner and repo name from git repo path
+    /// Extract GitHub owner and repo name from git repo path.
     pub fn get_github_repo_info(
```
ggordonhall: Don't think this facade is necessary anymore. The callers (all of whom are in `task_attempts/pr.rs`) can just use `GhCli` directly.

ggordonhall (approved):

PR #1804, opened 2026-01-06
Speedup pre-release CI builds

abcpro1:


.github/workflows/pre-release.yml
```diff
@@ -230,26 +288,39 @@ jobs:
...
 
-      - name: Build backend (non-Linux)
-        if: runner.os != 'Linux'
+      - name: Build backend (Windows)
+        if: runner.os == 'Linux' && contains(matrix.target, 'windows')
         run: |
-          cargo build --release --target ${{ matrix.target }} -p server
-          cargo build --release --target ${{ matrix.target }} --bin mcp_task_server
-          cargo build --release --target ${{ matrix.target }} -p review
+          cargo xwin build --release --target ${{ matrix.target }} -p server -p review --bin server --bin mcp_task_server --bin review
```
ggordonhall, line 316: We don't need to build the `review` binary at all here. We should avoid building it on any platform. It has a completely separate release cycle and does not depend on any other crates in this repo.

PR #1825, opened 2026-01-07
Show error when remote branch status is unavailable (Vibe Kanban)

LSRCT:
## Summary

When the branch status API fails, the git operations UI (Merge, PR/Push, Rebase buttons) was previously completely hidden without any indication to the user. This PR:

- Shows an error message when branch status cannot be fetched: "Unable to fetch branch status. You can still change the target branch."
- Keeps the "Change target branch" button enabled during errors, allowing users to recover from deadlock situations
- Provides visibility into what went wrong instead of silently failing

### Core Changes
- **`useBranchStatus.ts`**: Extended hook to expose `error`, `isError`, `isLoading`, and `refetch` properties from the React Query result
- **`GitOperations.tsx`**: 
  - Added `branchStatusError` prop to handle and display error state
  - Removed conflict-based disabling from "Change target branch" button to allow recovery
  - Shows error message with warning icon when branch status fails

### Prop Threading
- **`GitActionsDialog.tsx`**: Passes `branchStatusError` to `GitOperations`
- **`ProjectTasks.tsx`**: Extracts error from `useBranchStatus` and passes it through `DiffsPanelContainer`

### Internationalization
- Added `git.errors.branchStatusUnavailable` translation to all 6 supported locales (en, es, ja, ko, zh-Hans, zh-Hant)

## Why These Changes

Users were experiencing situations where the git operations section would disappear without explanation when:
- Network issues prevented fetching branch status
- API errors occurred
- The remote repository was temporarily unavailable

This left users unable to perform any git operations or even change the target branch to recover. Now users see a clear error message and can still change the target branch to work around the issue.

---

- [x] tested

---

### DIFF
--- a/frontend/src/components/tasks/Toolbar/GitOperations.tsx
+++ b/frontend/src/components/tasks/Toolbar/GitOperations.tsx
@@ -408,7 +408,7 @@ function GitOperations({
                 variant="ghost"
                 size="xs"
                 onClick={handleChangeTargetBranchDialogOpen}
-                disabled={isAttemptRunning}
+                disabled={isAttemptRunning || hasConflictsCalculated}
                 className={settingsBtnClasses}
                 aria-label={t('branches.changeTarget.dialog.title')}
               >


frontend/src/hooks/useBranchStatus.ts
```diff
@@ -2,10 +2,17 @@ import { useQuery } from '@tanstack/react-query';
...
 
 export function useBranchStatus(attemptId?: string) {
-  return useQuery({
+  const result = useQuery({
     queryKey: ['branchStatus', attemptId],
     queryFn: () => attemptsApi.getBranchStatus(attemptId!),
     enabled: !!attemptId,
     refetchInterval: 5000,
   });
+  return {
```
ggordonhall: Is this necessary?

LSRCT:
The original task mentioned allowing users to "get out of a deadlock" by keeping the change target branch button available. I interpreted this as removing both the conflict check and ensuring it works during errors.

However, I may have over-corrected. The conflict check (`hasConflictsCalculated`) was there to prevent changing the target branch when there are active git conflicts, which could leave the repo in an inconsistent state.

Options:
1. **Restore the conflict check** - safer, only enable during branch status errors
2. **Keep it removed** - more flexible for recovery, but riskier during conflicts
3. **Middle ground** - disable during conflicts UNLESS there's a branch status error

Which approach do you prefer?

PR #1826, opened 2026-01-07
feat(executors): add model_provider field to Droid executor

wuzhiguocarter:
Add support for specifying model provider in Droid executor configuration.

Changes:
- Add model_provider field to Droid struct (for UI display purposes)
- Update TypeScript types and JSON schema
- Note: model_provider is not passed to droid CLI as it doesn't support --model-provider flag
- The model ID itself contains all necessary information

This allows users to configure custom models like GLM-4.7 in the UI.


abcpro1:
This field is not needed?
> The model ID itself contains all necessary information

PR #1827, opened 2026-01-07
fix: update font-feature-settings in diff-style-overrides.css

paixaop:
Changed the font-feature-settings property for the diff-tailwindcss-wrapper to disable ligatures by setting it to "liga" off, ensuring consistent text rendering across different components.

All words with `fi`, the `f` and `i` were on top of each other.

This is a known issue stemming from the font's "fi" ligature, which combines the two characters into a single connected glyph. However, since Chivo Mono is designed to be monospaced (fixed-width for each character), the ligature disrupts this uniform spacing—causing the combined "fi" to take up only one character's width instead of two, which makes subsequent characters shift and appear overlapped or misplaced.

### DIFF
--- a/frontend/src/styles/diff-style-overrides.css
+++ b/frontend/src/styles/diff-style-overrides.css
@@ -446,7 +446,7 @@
 .diff-tailwindcss-wrapper .diff-add-widget,
 .diff-tailwindcss-wrapper .diff-widget-tooltip {
   font-family: inherit;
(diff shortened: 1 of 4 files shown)


ggordonhall:
@paixaop Please run `pnpm run format` and push the formatted changes.

paixaop:
ran formatter, thanks.

PR #1887, opened 2026-01-09
feat: auto-trigger agent when cleanup script fails

benmarten:
## Summary
- Automatically triggers the coding agent when a cleanup script fails
- Agent receives prompt to fix the cleanup script issues
- Implements retry limit (max 3 attempts) to prevent infinite loops
- Falls back to task finalization if agent cannot be started

## Test plan
- [x] Verified agent is triggered on cleanup script failure
- [x] Verified retry limit is respected
- [x] Verified task finalizes after max retries exceeded


Juanlucasbg:
## Rebase verification — PR #1887

- `crates/local-deployment/src/container.rs`
- `packages/bun.lock` (!)

The `packages/bun.lock` conflict is suspicious — vibe-kanban uses `pnpm`, not `bun` (per `CLAUDE.md` and the workspace setup). A stray `bun.lock` in a PR diff suggests either:
1. The contributor's local environment uses bun and accidentally committed the bun lockfile, or
2. The lockfile is stale debris that should be deleted.

Either way it's not a file the project tracks.

**Action**: rebase, **delete `packages/bun.lock` from the PR branch** (it shouldn't be there), then re-apply the `container.rs` change for "auto-trigger agent when cleanup script fails." Same `container.rs` surface as PRs #3058, #2103, others — verify the cleanup-script-failure path still exists in current code; the executor-lifecycle was reorganized.

— Automated rebase-verification: confirmed conflict; flagging the stray bun.lock as not-tracked-here.

PR #1895, opened 2026-01-09
Add PostHog analytics event for new UI access (Vibe Kanban)

stunningpixels:
## Summary

Adds PostHog analytics tracking to capture when users access the new UI (`/workspaces` routes).

## Changes

- Added `ui_new_accessed` event tracking in `NewDesignScope.tsx`
- Event fires once per component mount when users visit any `/workspaces` route
- Uses a `useRef` to prevent duplicate events from re-renders

## Implementation Details

The tracking is implemented in `NewDesignScope`, which wraps all new UI routes. This ensures the event is captured regardless of which specific `/workspaces` route the user lands on.

Key behaviors:
- **Once per mount**: The event fires once when the component mounts
- **Respects opt-out**: PostHog SDK automatically respects the user's analytics preferences
- **Minimal overhead**: Uses existing PostHog infrastructure already configured in the app

---

### DIFF
--- a/frontend/src/components/ui-new/scope/NewDesignScope.tsx
+++ b/frontend/src/components/ui-new/scope/NewDesignScope.tsx
@@ -6,8 +6,6 @@ import { ActionsProvider } from '@/contexts/ActionsContext';
 import NiceModal from '@ebay/nice-modal-react';
 import '@/styles/new/index.css';
 
-const UI_NEW_ACCESSED_KEY = 'ui_new_accessed';
-
 interface NewDesignScopeProps {
   children: ReactNode;
 }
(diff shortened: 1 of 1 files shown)


frontend/src/components/ui-new/scope/NewDesignScope.tsx
```diff
@@ -11,6 +12,16 @@ interface NewDesignScopeProps {
 
 export function NewDesignScope({ children }: NewDesignScopeProps) {
   const ref = useRef<HTMLDivElement>(null);
+  const posthog = usePostHog();
+  const hasTrackedSession = useRef(false);
```
LSRCT: what does this guard against? Not sure if it actually prevents any duplicates

PR #1898, opened 2026-01-09
fix: fix infinite loading on the projects page

Keith-CY:
Fix the infinite loading on the projects page when in dev mode by `setting isInitialized properly`

https://github.com/user-attachments/assets/f6a2e770-30da-4deb-857b-6b9e466ad334


ggordonhall:
This whole page has been reworked.

PR #1907, opened 2026-01-10
fix: copy_files not copying .env and other files to worktrees

nicholaslocascio:
## Summary

- Fix copy_files not copying .env and other files to worktrees due to incorrect path handling
- File search results include repo name prefix (e.g., repo-name/.env) for display, but this was incorrectly stored and used during copy operations, causing paths like /repo/repo-name/.env instead of /repo/.env

## Test plan

- [x] Verify existing copy tests pass (`cargo test -p local-deployment copy`)
- [ ] Add .env to copy_files in Project Settings via autocomplete
- [ ] Confirm the path is stored without repo name prefix (just .env)
- [ ] Create a new worktree and verify .env is copied from original repo
- [ ] Test with existing entries that have the old format (should auto-normalize)

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -704,7 +704,24 @@ impl LocalContainerService {
                 && !copy_files.trim().is_empty()
             {
                 let worktree_path = workspace_dir.join(&repo.name);
-                self.copy_project_files(&repo.path, &worktree_path, copy_files)
+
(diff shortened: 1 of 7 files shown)


ggordonhall:
I've added auto-complete to the copy file input box which should address the issue here: https://github.com/BloopAI/vibe-kanban/pull/2004.

PR #1931, opened 2026-01-11
Display project name in Navbar when available

Pdzly:
Added for clarity in what project you are in, the project name ( if available ) to the header.


ggordonhall:
We're re-working the UI now so not looking to merge changes like this!

Issue #1813, opened 2026-01-07
Files listed in Project settings copy files not get copied


msaidf:
I put .env file to the list of copy files in Project settings, but it still do not get copied over to the worktrees. My app cannot connect to external services because the key is in .env that do not get copied

nicholaslocascio:
Opened up a PR https://github.com/BloopAI/vibe-kanban/pull/1907 that fixes this

PR #2004, opened 2026-01-13
feat: copy-file autocomplete

ggordonhall:
Add file autocomplete to the copy input box. 

https://github.com/BloopAI/vibe-kanban/issues/1813


frontend/src/lib/api.ts
```diff
@@ -895,6 +895,20 @@ export const repoApi = {
     });
     return handleApiResponse<OpenEditorResponse>(response);
   },
+
+  searchFiles: async (
+    repoId: string,
+    query: string,
+    mode?: string,
```
LSRCT: could we export the mode type and use it here instead of a string?

PR #2005, opened 2026-01-13
bump claude code to v2.1

abcpro1:


abcpro1:
this is the stable though

abcpro1:
should we use stable or latest?

ggordonhall:
> should we use stable or latest?

Ah! Didn't know that they made that distinction. Let's use `stable`.

PR #2018, opened 2026-01-13
Fix command builder whitespace handling

abcpro1:
Add a separate command for parameters that need to be split on whitespace boundaries, and default to not splitting.

The previous behaviour always split parameters, which caused copilot CLI to fail when a project had whitespace in the repository directory name; `error: too many arguments. Expected 0 arguments but got 1`.

Even though it only happened in Copilot because of an implementation detail where it received a directory name that also happened to contain the project name, the problem was not unique to Copilot.

### DIFF
--- a/crates/executors/src/command.rs
+++ b/crates/executors/src/command.rs
@@ -92,7 +92,7 @@ impl CommandBuilder {
         self
     }
 
-    pub fn extend_shell_params<I>(mut self, more: I) -> Result<Self, CommandBuildError>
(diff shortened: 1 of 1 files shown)


ggordonhall:
!reviewfast

crates/executors/src/command.rs
```diff
@@ -90,6 +92,27 @@ impl CommandBuilder {
         self
     }
 
+    pub fn extend_shell_params<I>(mut self, more: I) -> Result<Self, CommandBuildError>
```
LSRCT: private if possible?

PR #2032, opened 2026-01-14
fixed flickering

t3dotgg:
Spent a lot of time tracing this. Would love to know why that key was on that provider (generally speaking you should never have to key a context provider)

### DIFF
--- a/frontend/src/components/ui-new/ConversationList.tsx
+++ b/frontend/src/components/ui-new/ConversationList.tsx
@@ -97,10 +97,10 @@ export function ConversationList({ attempt, task }: ConversationListProps) {
     loading: boolean;
   } | null>(null);
   const debounceTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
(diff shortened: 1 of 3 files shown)


Juanlucasbg:
## Rebase verification — PR #2032

- `frontend/src/components/ui-new/ConversationList.tsx`
- `frontend/src/components/ui-new/views/WorkspacesMain.tsx`
- `frontend/src/hooks/useConversationHistory.ts`

Pre-monorepo paths. "Fixed flickering" needs to be re-applied to the current tree (likely `packages/ui/src/components/ConversationList.tsx`, `packages/web-core/...`, etc.).

**Action**: this PR predates the monorepo restructure. The flicker fix may already be present on current `main` (several conversation-history / scroll-stabilization PRs have landed since). Recommend close as likely superseded; if the flicker still reproduces, open a fresh PR against the current paths.

— Automated rebase-verification: detected delete-vs-modify across multiple pre-monorepo paths.

PR #2102, opened 2026-01-16
Feature/add all project board

neoAimeow:
1. Add a Makefile for easier project startup.
2. Add a global project board for managing multiple projects (entry in the top-right corner).

image


laraben:
oh tis would be nice

neoAimeow:
> oh tis would be nice

but no one review

Ton-Git:
This would be really useful

PR #2103, opened 2026-01-16
fix: Process queued messages when cleanup script is skipped

jtomaszewski:
## Summary
- Fixes bug where queued follow-up messages were not processed when a cleanup script is configured but skipped (because the coding agent made no changes)
- Adds queued message handling to the "no changes" path in the exit monitor

## Problem
When a cleanup script is configured but the coding agent makes no changes:
1. `should_start_next = false` (no changes made)
2. `finalize_task()` was called directly
3. But `should_finalize()` returns false (because `next_action` exists - the cleanup script)
4. So the queued message processing block was never entered

## Solution
Added queued message processing directly in the "no changes" path where we skip the cleanup script. This ensures follow-up messages are properly consumed and executed.

## Test plan
- [x] Create a task that makes changes with cleanup script configured
- [x] Queue a follow-up message while task is running
- [x] Verify the queued message is processed after task completion

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -461,8 +461,43 @@ impl LocalContainerService {
                             ctx.workspace.id
                         );
 
-                        // Manually finalize task since we're bypassing normal execution flow
-                        container.finalize_task(publisher.as_ref().ok(), &ctx).await;
+                        // When bypassing cleanup, we need to process queued messages here
+                        // since should_finalize() will return false (next_action exists)
+                        if let Some(queued_msg) =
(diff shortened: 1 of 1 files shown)


Juanlucasbg:
## Rebase verification — PR #2103

- `crates/local-deployment/src/container.rs`

Same file as #3058 (zombie-process fix); both touch lifecycle code in `container.rs` that has been refactored on `main`.

**Action**: rebase, then verify whether "process queued messages when cleanup script is skipped" still has the same code path. The cleanup-script handling may have been reorganized — the fix's call site may have moved, or the logic that skipped the cleanup script may have been removed.

PR #2112, opened 2026-01-17
chore(deps): add npx-cli to pnpm-workspace.yaml

ctsstc:
This allows the `pnpm install` command to also install dependencies and lock them for the npx-cli.

Otherwise following the `README.md` does not work without additional manual legwork.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -313,6 +313,12 @@ importers:
         specifier: ^5.0.8
         version: 5.4.19(@types/node@24.10.1)
 
+  npx-cli:
+    dependencies:
+      adm-zip:
+        specifier: ^0.5.16
+        version: 0.5.16
+
   remote-frontend:
     dependencies:
       '@git-diff-view/file':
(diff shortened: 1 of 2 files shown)


anastasiya1155 (commented):
Adding npx-cli to pnpm-workspace would install dependencies that are only needed for packaging and publishing the app, most contributors don't need that. It is better to keep dependencies lean and installation faster. I think a better approach would be to update our README.md to include dependency installation step for building from source

ctsstc:
I believe the only dependency installed by including `npx-cli` is `adm-zip` as seen in the changes to the lock file and in its [`package.json`](https://github.com/BloopAI/vibe-kanban/blob/aecb2824bebb4d8c7e77d1949a90e875297cda4f/npx-cli/package.json#L13-L15). It brings in 121kb according to: <https://www.npmjs.com/package/adm-zip>

This is fairly incomparable to the weight of the other dependencies in the top level package and especially the Rust dependencies coming in around 15GB. You might even get it for "free" due to how pnpm works if you already have it on your system.

It may not be the most efficient for everyone, but it's a good DX quality of life improvement for little cost.

Just wanted to share my thoughts on this.

Otherwise I agree that there should at least be some documentation about this.

PR #2147, opened 2026-01-19
fix: correct console.warn! to console.warn syntax

jtomaszewski:
## Summary
- Fixed invalid JavaScript syntax where `console.warn!` (Rust macro syntax) was used instead of `console.warn()` in `useConversationHistory.ts`

## Test plan
- [x] TypeScript compilation passes (`pnpm run check`)
- [x] Frontend lint passes (`pnpm run frontend:lint`)


stunningpixels:
nice catch, bad claude

PR #2149, opened 2026-01-19
feat(utils): add VIBE_KANBAN_ASSET_DIR env var override

jtomaszewski:
## Summary
- Adds `VIBE_KANBAN_ASSET_DIR` environment variable override for the asset directory
- Allows running dev mode against the global database instead of the local `dev_assets` copy - **this can simplify testing any bugfixes to vibe-kanban**

## Test plan
- [x] Set `VIBE_KANBAN_ASSET_DIR=/tmp/test-assets` and run `pnpm run dev`, verify it uses the override path
- [x] Run without the env var, verify default behavior is unchanged


ggordonhall:
Hi @jtomaszewski, thanks but this isn't something we're looking at add right now.

PR #2151, opened 2026-01-19
fix: improve OpenCode installation detection for jsonc config and additional paths

vdmkotai:
- Support opencode.jsonc config file (not just opencode.json)
- Add detection for OpenCode Desktop app (~/.../ai.opencode.desktop)
- Add detection for OpenCode CLI home directory (~/.opencode)

### DIFF
--- a/crates/executors/src/mcp_config.rs
+++ b/crates/executors/src/mcp_config.rs
@@ -5,7 +5,10 @@
 
 use std::{collections::HashMap, path::Path, sync::LazyLock};
 
-use jsonc_parser::{ParseOptions, cst::CstRootNode};
+use jsonc_parser::{
+    ParseOptions,
+    cst::{CstObject, CstRootNode},
+};
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
 use tokio::fs;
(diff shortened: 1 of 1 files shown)


crates/executors/src/executors/opencode.rs
```diff
@@ -221,13 +221,28 @@ impl StandardCodingAgentExecutor for Opencode {
...
+        // Try multiple config file names (.json and .jsonc) in XDG/platform config dirs
         #[cfg(unix)]
         {
-            xdg::BaseDirectories::with_prefix("opencode").get_config_file("opencode.json")
+            let base_dirs = xdg::BaseDirectories::with_prefix("opencode");
+            // First try opencode.json, then opencode.jsonc
+            base_dirs
+                .get_config_file("opencode.json")
+                .filter(|p| p.exists())
+                .or_else(|| base_dirs.get_config_file("opencode.jsonc"))
```
LSRCT, line 266: we have built-in MCP config editor in the frontend settings (`/settings/mcp`) which I think does not handle `.jsonc`.
Opencode supports .jsonc, maybe there is a way for us to support it, too?

vdmkotai: updated

LSRCT: I tried it and while it doesn't throw an error anymore, we remove all nested comments on save:
```
alexnetsch@Alexs-MacBook-Pro ~ % cat .config/opencode/opencode.jsonc
{
  "mcp": {
    "vibe_kanban": {...}
  },
  "$schema": "https://opencode.ai/config.json",
  "tui": {
    // test
    "scroll_speed": 3,
  }
}
```

Change nothing, press save in vk:
```
alexnetsch@Alexs-MacBook-Pro ~ % cat .config/opencode/opencode.jsonc
{
  "mcp": {
    "vibe_kanban": {...}
  },
  "$schema": "https://opencode.ai/config.json",
  "tui": {
    "scroll_speed": 3,
  }
}
```

IMO there are two options here:
- Fix the `write_jsonc_preserving_comments` function to actually preserve comments where we can
- Just drop comments on save completely

Given the complexity of this I'd say we just treat `.jsonc` as `.json` and nuke the comments on write.
(`.TOML` has the same issue, we just drop the comments.)

What do you think?

vdmkotai: I was thinking about the same but decided to try 
I'll take another look and if no - ok then

vdmkotai: @LSRCT should be good now (comments are preserved)
can you give it a try?

vdmkotai:
!reviewfast

LSRCT (approved):

stunningpixels (approved):

PR #2156, opened 2026-01-19
fix(frontend): filter token_usage_info entries to fix VirtuosoMessageList rendering

jtomaszewski:
## Summary
- Filter out `token_usage_info` entries in `flattenEntriesForEmit` because they render as `null` in `DisplayConversationEntry`
- This fixes VirtuosoMessageList calculating zero height for these entries, causing the list to get stuck when starting from the last item
- Token usage continues to be displayed in the header gauge via the separate `useTokenUsage` hook

## Test plan
- [ ] Run the app and start a conversation
- [ ] Verify VirtuosoMessageList scrolls correctly to the bottom
- [ ] Verify token usage is still displayed in the header gauge
- [ ] Verify no stuck/frozen list behavior when new messages arrive

### DIFF
--- a/frontend/src/hooks/useConversationHistory.ts
+++ b/frontend/src/hooks/useConversationHistory.ts
@@ -251,11 +251,14 @@ export const useConversationHistory = ({
             );
             entries.push(userPatchTypeWithKey);
 
-            // Remove all coding agent added user messages, replace with our custom one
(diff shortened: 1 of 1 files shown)


stunningpixels:
Closing in favour of https://github.com/BloopAI/vibe-kanban/pull/2172

Same fix but forked useConversationHistory so we don't break token counts in the new UI

Issue #2164, opened 2026-01-20
Bug: Git repository scanner returns duplicate results on case-insensitive filesystems


G-XD:
## Problem

On case-insensitive filesystems (macOS APFS default, Windows), the git repository scanner returns **duplicate repository entries** when the same repository is scanned via paths that differ only in casing.

<img width="1154" height="992" alt="Image" src="https://github.com/user-attachments/assets/d95a3617-9012-4694-bb46-d4817f590bff" />

### Example

If you have a repository at `/Users/me/Projects/my-repo`:

**Current behavior:**

```
[
  { name: "my-repo", path: "/Users/me/Projects/my-repo" },
  { name: "my-repo", path: "/Users/me/projects/my-repo" }  // Duplicate!
]
```

**Expected behavior:**
```
[
  { name: "my-repo", path: "/Users/me/Projects/my-repo" }  // Only one entry
]
```

stefanofa:
@stunningpixels is this something only happening to me and @G-XD? because I don't see it being fixed and also I don't see anyone else that acknowledges this

PR #2165, opened 2026-01-20
fix: dedupe repo scan paths via normalized+canonicalized keys

G-XD:
- Use canonicalized paths for deduplication in list_git_repos_inner
  - Deduplicate scan paths in list_common_git_repos to avoid redundant I/O

Fix: #2164

### DIFF
--- a/crates/services/src/services/filesystem.rs
+++ b/crates/services/src/services/filesystem.rs
@@ -52,6 +52,17 @@ impl FilesystemService {
     }
 
     #[cfg(not(feature = "qa-mode"))]
+    fn normalize_path_for_dedupe(path: &Path) -> PathBuf {
+        // 1) Normalize known macOS aliases (/private/var -> /var, /private/tmp -> /tmp)
+        // 2) Canonicalize when possible to resolve symlinks/.. and produce a stable key for
+        //    dedupe (especially on case-insensitive filesystems common on macOS).
+        let normalized = utils::path::normalize_macos_private_alias(path);
+        // If canonicalize fails (e.g., path doesn't exist), use the normalized path
+        // rather than the original to ensure consistent /private/var handling
+        dunce::canonicalize(&normalized).unwrap_or(normalized)
+    }
+
+    #[cfg(not(feature = "qa-mode"))]
     fn get_directories_to_skip() -> HashSet<String> {
         let mut skip_dirs = HashSet::from(
             [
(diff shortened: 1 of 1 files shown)


Juanlucasbg:
## Aggressive review summary — PR #2165

24-line fix to `crates/services/src/services/filesystem.rs` adding `normalize_path_for_dedupe()` and applying it (a) to input scan roots before launching the walker and (b) to output entries inside the walker filter. Verdict: **clean — recommend merge**.

### Why the fix is correct

Two concrete dedup gaps existed:

1. **Input roots**: A user passes `["/Users/foo", "/users/foo"]` (case differs on case-insensitive APFS) or `["/var/x", "/private/var/x"]` (macOS alias). Both list the same tree; pre-PR each one launches a separate walker and the work is duplicated.
2. **Walker output**: When multiple roots overlap (added at lines 270+ via `walker_builder.add`), the same directory shows up under different in-memory `Path` representations and is yielded twice. The pre-PR `seen_dirs.contains(entry.path())` compares raw paths so dedup misses.

The PR's key:
```rust
let normalized = utils::path::normalize_macos_private_alias(path);
dunce::canonicalize(&normalized).unwrap_or(normalized)
```

This (a) collapses `/private/var → /var` (re-using existing helper from `crates/utils/src/path.rs`), (b) resolves symlinks via `dunce::canonicalize` (Windows-safe variant — already used in `crates/services/src/services/filesystem_watcher.rs:47` with the same `unwrap_or` fallback pattern), (c) gracefully degrades to the normalized path when canonicalize fails (path doesn't exist yet, permission denied, etc.).

### Findings

- **Structural**: 0 HIGH. Dep already in `Cargo.toml:48` (`dunce = "1.0"`); no new dep. `#[cfg(not(feature = "qa-mode"))]` matches the surrounding gating.
- **Adversarial**: 0 HIGH. Failure mode = canonicalize fails → falls back to normalized path. Not worse than pre-PR which had no normalization at all.
- **Security**: 0 HIGH. Symlink resolution via canonicalize is the desired behavior (lets the picker honor the actual on-disk identity).
- **Conventions**: PASS — mirrors `filesystem_watcher.rs:47` exactly.
- **Performance NIT**: Calling `dunce::canonicalize` on every walker entry is a syscall per directory. On a tree with 100k dirs that adds a meaningful syscall overhead even when no dedup happens. Most users won't notice; consider gating the canonicalize behind "the same path appears > 1 time in `paths`" if performance shows up.

### NITs (non-blocking)

- The added comment on lines 274-275 is good — keep it.
- Consider documenting that the dedup key is "normalize-then-canonicalize-with-graceful-fallback" so future readers don't re-derive the rationale.

### Verdict

Approve.

— Reviewed by automated single-pass review (path-dedup convention triage; full 4-tool battery skipped — diff is 24 lines and the canonicalize pattern matches existing usage at `filesystem_watcher.rs:47`).

Issue #2093, opened 2026-01-16
unable to create file xxx: Filename too long


arcadia-ai:
After creating the task, it remains stuck in the "loading history" state.
<img width="2459" height="892" alt="Image" src="https://github.com/user-attachments/assets/4c9d5155-52f0-48b6-ae12-feea3f0dee41" />
<img width="1479" height="799" alt="Image" src="https://github.com/user-attachments/assets/145b44a1-b969-454b-a5f3-f01d1224adb7" />

ggordonhall:
You can now manually override the default worktree directory. If you set it to something short you might get round the system file path length limits that you're running into.

Issue #1830, opened 2026-01-07
Feature Request: Configurable base directory for Git worktrees (per-project + global default)


irvingoujAtDevolution:
**Problem**
Vibe Kanban creates a git worktree per task attempt, but there’s no documented way to choose *where* those worktrees are stored (e.g., move them off the system drive / to a fast scratch disk).

**Proposed change**
- Add a **Worktree Base Directory** setting (global default)
- Optional **per-project override**
- Optional env var for headless deployments: `VIBE_WORKTREE_BASE_DIR=/path/to/worktrees`

**Acceptance**
When the base dir is set, new task attempts create their worktrees under that directory; if unset, behavior stays the same.

Maxsky5:
+1 that would be awesome to be able to configure the worktree directory!

waylonwang:
+1

ggordonhall:
A global override has been implemented in the latest release. We're overhauling per-project settings. Will follow on in time.

Issue #1620, opened 2025-12-20
Feature Request: Custom worktrees path


cncolder:
Hi team thanks for the awesome tools.

Is there a way to custom the git worktrees path instead of tmp path? e.g. 

When my repo is `~/codes/my-project` and I want to create the git worktrees to `~/codes/my-project.worktrees/abc` like what the vscode copilot chat background agent doing.

The pros:
- Custom env by [direnv](https://github.com/direnv/direnv)
- Custom gitconfig by path blob
- Easily manage git worktrees by the other tools.

EvanZhouDev:
Would love this feature as well. Just generally makes it easier to monitor what's happening with the project.

Also since I already am using worktrees, there's a bit of confusion re: where the source of truths for the worktrees are when Vibe Kanban is doing it's own thing.

waynebrantley:
@stunningpixels would you accept a pr for this?

waylonwang:
+1

ggordonhall:
A global worktree default directory override has been added in the latest release.

Issue #1244, opened 2025-11-12
[Bug] Worktrees not working when project is on another drive


Inrego:
It's creating the worktree just fine. But when it's reading/writing files, it works with the original folder.
I suspect the issue might be that my project code is on another drive than where vk is installed. I get the following in logs:

`WARN utils::path: Failed to make canonical path relative: '\\?\E:\Projects\Visual Studio\InstallerPortal\InstallerPortal\Services\InstallerPortalService.cs' relative to '\\?\C:\Users\Username\AppData\Local\Temp\vibe-kanban\worktrees\4e64-table-data-wrong', error: prefix not found, returning original`

ggordonhall:
You might be able to get this working by configuring a custom worktree directory. This is implemented in the latest release.

PR #2169, opened 2026-01-20
feat: override default worktree directory

ggordonhall:
Adds a global override for the default worktree directory.

Addresses:
- https://github.com/BloopAI/vibe-kanban/issues/2093
- https://github.com/BloopAI/vibe-kanban/issues/1830
- https://github.com/BloopAI/vibe-kanban/issues/1620
- https://github.com/BloopAI/vibe-kanban/issues/1244

### DIFF
--- a/frontend/src/i18n/locales/en/settings.json
+++ b/frontend/src/i18n/locales/en/settings.json
@@ -135,6 +135,14 @@
             "invalidChars": "Contains invalid characters.",
             "controlChars": "Contains control characters."
           }
+        },
+        "workspaceDir": {
+          "label": "Workspace Directory",
+          "placeholder": "~/",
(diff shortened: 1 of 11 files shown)


frontend/src/i18n/locales/en/settings.json
```diff
@@ -135,6 +135,14 @@
             "invalidChars": "Contains invalid characters.",
             "controlChars": "Contains control characters."
           }
+        },
+        "workspaceDir": {
+          "label": "Workspace Directory",
+          "placeholder": "~/",
```
LSRCT, line 141: Minor, but maybe this is misleading. The default is the temp directory which can be cleaned up by the OS.

LSRCT (approved):
Works!

PR #2175, opened 2026-01-20
fix(file_search): store debouncer and cleanup watchers on cache eviction

jtomaszewski:
## Summary

- Store debouncer in watchers DashMap (was created but never stored, causing watcher to be dropped immediately)
- Add eviction listener to cache builder to cleanup watchers when entries expire  
- Update watchers type to `Arc<DashMap<...>>` for sharing with eviction listener

## Context

Lately I've been getting a lot of errors like below after VibeKanban runs for a while. Restarting helps temporarily. This PR is an attempt to fix the root cause - file watchers being created but never cleaned up.

```
2026-01-20T12:50:13.588224Z ERROR services::services::diff_stream: Diff stream manager failed: IO error: /private/var/folders/.../participants: IO error for operation on /private/var/folders/.../participants: Too many open files (os error 24)
```

## Test plan

- [ ] Run VibeKanban for extended period and monitor for "Too many open files" errors
- [ ] Verify file watchers are properly cleaned up when cache entries expire

### DIFF
--- a/crates/services/src/services/file_search.rs
+++ b/crates/services/src/services/file_search.rs
@@ -12,8 +12,8 @@ use db::models::{
 use fst::{Map, MapBuilder};
 use ignore::WalkBuilder;
 use moka::future::Cache;
-use notify::{RecommendedWatcher, RecursiveMode};
-use notify_debouncer_full::{DebounceEventResult, new_debouncer};
+use notify::RecursiveMode;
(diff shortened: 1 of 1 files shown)


jtomaszewski:
Closing this, as it hasn't resolved the main issue.

I still get "Too many open files (os error 24)" though if I keep vibe kanban open long enough. As soon as I restart it, everything goes back to normal.

PR #2176, opened 2026-01-20
feat: add VIBE_KANBAN_PORT env var to avoid conflicts with child processes

jtomaszewski:
## Problem

Many apps depend on the `PORT` env var. When running Vibe Kanban with `PORT=3000`, any child processes (e.g., Node.js dev servers spawned by task attempts) inherit this variable and try to bind to the same port, causing conflicts.

## Solution

Add `VIBE_KANBAN_PORT` as an alternative that takes precedence over `PORT`. This allows setting a dedicated port for Vibe Kanban without affecting child processes that read `PORT`.

## Changes
- `crates/server/src/main.rs` - read `VIBE_KANBAN_PORT` before `PORT`
- `crates/server/src/bin/mcp_task_server.rs` - read `VIBE_KANBAN_PORT` before `PORT`
- `scripts/setup-dev-environment.js` - check `VIBE_KANBAN_PORT` before `PORT`
- `README.md` - document the new env var

## Test plan
- [x] `cargo check --workspace` passes
- [x] `pnpm run check` passes
- [x] Dev build: `VIBE_KANBAN_PORT=4444` correctly allocates ports 4444/4445
- [x] Dev build: `PORT=3000 VIBE_KANBAN_PORT=4444` uses 4444 (precedence verified)
- [x] Release build: `VIBE_KANBAN_PORT=4444` starts server on :4444
- [x] Release build: `PORT=3000 VIBE_KANBAN_PORT=5555` starts server on :5555 (precedence verified)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ The following environment variables can be configured at build time or runtime:
 |----------|------|---------|-------------|
 | `POSTHOG_API_KEY` | Build-time | Empty | PostHog analytics API key (disables analytics if empty) |
 | `POSTHOG_API_ENDPOINT` | Build-time | Empty | PostHog analytics endpoint (disables analytics if empty) |
+| `VIBE_KANBAN_PORT` | Runtime | Auto-assign | Same as `PORT` but takes precedence; avoids conflicts with child processes that read `PORT` |
 | `PORT` | Runtime | Auto-assign | **Production**: Server port. **Dev**: Frontend port (backend uses PORT+1) |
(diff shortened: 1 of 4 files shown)


LSRCT:
Did you check if the `BACKEND_PORT` env var solves your issue?
If it does not ideally we'd simplify the environment set up rather than adding one more variable if that's possible.

jtomaszewski:
Seems like it does, thanks.

Although I'd still suggest either making the port available as a command argument `--port 1234` or as a prefixed env var so that it doesn't conflict with any other env vars that may be commonly used by any of user's projects.

PR #2184, opened 2026-01-20
feat: add unified Entity SDK with TanStack DB optimistic mutations (Vibe Kanban)

stunningpixels:
## Summary

This PR introduces a unified Entity SDK that combines Electric SQL shapes with TanStack DB's optimistic mutation support, providing a complete real-time sync + mutation solution for the frontend.

### Key Changes

**Backend (Rust)**
- Added `define_entity!` macro that unifies shape definitions with mutation type generation
- Replaced `IssueDependency` with more flexible `IssueRelationship` model supporting multiple relationship types (blocking, related, has_duplicate)
- Added `id` field to join tables (IssueAssignee, IssueFollower, IssueTag) for proper mutation support
- Generated Create/Update request types for all entities

**Shared Types (`remote-types.ts`)**
- Added `EntityDefinition<TRow, TCreate, TUpdate>` interface with phantom types for type-safe SDK generation
- Generated entity definitions for all 12 entities with mutation URLs and scope metadata
- Cleaned up phantom type pattern using object casts instead of null coercion

**Frontend Electric SDK**
- Added `createEntityCollection()` in `collections.ts` with `onInsert`, `onUpdate`, `onDelete` handlers that leverage TanStack DB's built-in optimistic updates
- Created unified `useEntity()` hook returning `{ data, isLoading, error, retry, insert, update, remove }`
- Deleted deprecated `useElectricCollection` hook and barrel export file
- Deleted unused `config.ts`

**Electric Test Page**
- Migrated from shapes to entities throughout
- Added `MutationPanel` component for testing CRUD operations
- Wired up Tags and Comments lists with full mutation support (create, update, delete)

### Why

The previous setup required separate implementations for reading (Electric shapes) and writing (API calls). This unified approach:
1. Uses TanStack DB's built-in optimistic update support for instant UI feedback
2. Provides type-safe mutations derived from Rust types
3. Simplifies the developer experience with a single `useEntity` hook for both sync and mutations
4. Electric automatically reconciles optimistic state when real data syncs from Postgres

### Implementation Details

- Mutations are optimistic: changes appear immediately in the UI while API requests happen in background
- If an API call fails, TanStack DB automatically rolls back the optimistic state
- Entity definitions include scope metadata (Organization, Project, Issue, Comment) for proper URL construction
- The `useEntity` hook creates collections lazily and caches them per entity/params combination

---

### DIFF
--- a/crates/remote/src/db/issues.rs
+++ b/crates/remote/src/db/issues.rs
@@ -185,8 +185,13 @@ impl IssueRepository {
         Ok(MutationResponse { data, txid })
     }
 
-    /// Update an issue with partial fields. Uses COALESCE to preserve existing values
-    /// when None is provided.
+    /// Update an issue with partial fields.
+    ///
+    /// For non-nullable fields, uses COALESCE to preserve existing values when None is provided.
+    /// For nullable fields (Option<Option<T>>), uses CASE to distinguish between:
+    /// - None: don't update the field
+    /// - Some(None): set the field to NULL
+    /// - Some(Some(value)): set the field to the value
     #[allow(clippy::too_many_arguments)]
     pub async fn update(
         pool: &PgPool,
@@ -204,6 +209,19 @@ impl IssueRepository {
     ) -> Result<MutationResponse<Issue>, IssueError> {
         let mut tx = pool.begin().await?;
 
+        // For nullable fields, extract boolean flags and flattened values
+        // This preserves the distinction between "don't update" and "set to NULL"
+        let update_description = description.is_some();
+        let description_value = description.flatten();
+        let update_start_date = start_date.is_some();
+        let start_date_value = start_date.flatten();
+        let update_target_date = target_date.is_some();
+        let target_date_value = target_date.flatten();
+        let update_completed_at = completed_at.is_some();
+        let completed_at_value = completed_at.flatten();
+        let update_parent_issue_id = parent_issue_id.is_some();
+        let parent_issue_id_value = parent_issue_id.flatten();
+
         let data = sqlx::query_as!(
             Issue,
             r#"
@@ -211,16 +229,16 @@ impl IssueRepository {
             SET
                 status_id = COALESCE($1, status_id),
                 title = COALESCE($2, title),
-                description = COALESCE($3, description),
-                priority = COALESCE($4, priority),
-                start_date = COALESCE($5, start_date),
-                target_date = COALESCE($6, target_date),
-                completed_at = COALESCE($7, completed_at),
-                sort_order = COALESCE($8, sort_order),
-                parent_issue_id = COALESCE($9, parent_issue_id),
-                extension_metadata = COALESCE($10, extension_metadata),
+                description = CASE WHEN $3 THEN $4 ELSE description END,
+                priority = COALESCE($5, priority),
+                start_date = CASE WHEN $6 THEN $7 ELSE start_date END,
+                target_date = CASE WHEN $8 THEN $9 ELSE target_date END,
+                completed_at = CASE WHEN $10 THEN $11 ELSE completed_at END,
+                sort_order = COALESCE($12, sort_order),
+                parent_issue_id = CASE WHEN $13 THEN $14 ELSE parent_issue_id END,
+                extension_metadata = COALESCE($15, extension_metadata),
                 updated_at = NOW()
-            WHERE id = $11
+            WHERE id = $16
             RETURNING
                 id                  AS "id!: Uuid",
                 project_id          AS "project_id!: Uuid",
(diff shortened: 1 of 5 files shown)


ggordonhall:
!reviewfast

crates/remote/src/shapes.rs
```diff
@@ -1,22 +1,24 @@
+//! Shape definitions for realtime streaming.
```
ggordonhall, line 1: Delete this file

crates/remote/src/mutation_types.rs
```diff
@@ -0,0 +1,192 @@
...
+///
+/// // Handlers must be defined:
+/// async fn list_tags(...) -> Result<Json<ListTagsResponse>, ErrorResponse> { ... }
+/// async fn get_tag(...) -> Result<Json<Tag>, ErrorResponse> { ... }
+/// async fn create_tag(...) -> Result<Json<Tag>, ErrorResponse> { ... }
+/// async fn update_tag(...) -> Result<Json<Tag>, ErrorResponse> { ... }
+/// async fn delete_tag(...) -> Result<StatusCode, ErrorResponse> { ... }
+/// ```
+#[macro_export]
+macro_rules! define_mutation_router {
```
ggordonhall, line 172: I don't love this. Makes it harder to follow what's going on under the hood. Would rather routes were defined explicitly.

crates/remote/src/entities.rs
```diff
@@ -0,0 +1,231 @@
...
+crate::define_entity!(
+    Notification,
+    table: "notifications",
+    mutation_scope: Organization,
+    shape: {
+        where_clause: r#""organization_id" = $1 AND "user_id" = $2"#,
+        params: ["organization_id", "user_id"],
+        url: "/shape/notifications",
+    },
+    fields: [seen: bool],
```
ggordonhall, line 57: It's no longer possible to updated the `dismissed_at` field. If unused, we should remove that field from the DB completely (and references to it).

crates/remote/src/routes/issue_tags.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    Ok(Json(issue_tag))
+}
+
+#[instrument(
+    name = "issue_tags.update_issue_tag",
+    skip(_state, _ctx, _payload),
+    fields(issue_tag_id = %_issue_tag_id)
+)]
+async fn update_issue_tag(
```
ggordonhall, line 103: Router macro generates non-existent endpoints like this one. We should remove these. This is the case on other join tables.

crates/remote/src/db/issues.rs
```diff
@@ -90,4 +90,185 @@ impl IssueRepository {
...
+                completed_at        AS "completed_at?: DateTime<Utc>",
+                sort_order          AS "sort_order!",
+                parent_issue_id     AS "parent_issue_id?: Uuid",
+                extension_metadata  AS "extension_metadata!: Value",
+                created_at          AS "created_at!: DateTime<Utc>",
+                updated_at          AS "updated_at!: DateTime<Utc>"
+            "#,
+            status_id,
+            title,
+            description.flatten(),
```
ggordonhall: I think calling `.flatten()` here collapses the distinction between `Option<T>` and `Option<Option<T>>` so we lose the distinction between _not_ updating a value, and updating it to `null`.

ggordonhall (commented):
Would be nice to collect macro definitions in a `remote/src/macros/` directory .

crates/remote/src/shapes.rs
stunningpixels: It's still needed the deprecation notice was erroneous and has been removed

crates/remote/src/mutation_types.rs
stunningpixels: It enforces that all CRUD operations exist, the current pattern is:
- Force that these methods exist in the router
- If a method doesn't support a CRUD operation, add it but return error

Probably better ways we can do this with further refactoring in future

crates/remote/src/entities.rs
stunningpixels: Updated, this gets updated when seen is changed

crates/remote/src/routes/issue_tags.rs
stunningpixels: See above, known limitation of approach, let's ship and refactor later

crates/remote/src/db/issues.rs
stunningpixels: Fixed

crates/remote/src/routes/issue_tags.rs
ggordonhall: My refactor would be to remove the macro altogether. Not a fan of this approach.

ApocalypseYun:
It seems that this PR changed the return format of the remote service create_project from Project to MutationResponse<Project>, but forgot to update the local client RemoteClient::create_project to parse the new format, resulting in an exception when creating and linking remote projects?

ggordonhall:
> It seems that this PR changed the return format of the remote service create_project from Project to MutationResponse, but forgot to update the local client RemoteClient::create_project to parse the new format, resulting in an exception when creating and linking remote projects?

Hi @ApocalypseYun, this feature has been removed in preparation for improved team collaboration functionality.

ApocalypseYun:
@ggordonhall Thank you for your reply, but I see that there are still some reservations on the interface. May I inquire about the plans regarding improvements to team collaboration features?

PR #2193, opened 2026-01-21
feat: slash commands

### LINKED ISSUE #642: Slash commands
Determine available slash commands (`~/.claude/commands` and current projects's `.claude/commands`) and show a menu when `/` is typed with support of what is in command's frontmatter, reading description and parameters. Read more at https://docs.anthropic.com/en/docs/claude-code/slash-commands#custom-slash-commands

abcpro1:
Fixes #642 

Adds slash command auto-completion in the session chat box for Claude Code, Codex, and Opencode with a typehead popup similar to one existing for tags auto-completion.

Screenshot 2026-01-21 at 12 29 34

- [x] tested

### DIFF
--- a/frontend/src/hooks/useSlashCommands.ts
+++ b/frontend/src/hooks/useSlashCommands.ts
@@ -13,10 +13,11 @@ export function useSlashCommands(
   agent: BaseCodingAgent | null | undefined,
   opts?: { workspaceId?: string; repoId?: string }
 ) {
+  const { workspaceId, repoId } = opts ?? {};
   const endpoint = useMemo(() => {
     if (!agent) return undefined;
-    return agentsApi.getSlashCommandsStreamUrl(agent, opts);
-  }, [agent, opts?.repoId, opts?.workspaceId]);
+    return agentsApi.getSlashCommandsStreamUrl(agent, { workspaceId, repoId });
+  }, [agent, workspaceId, repoId]);
 
   const initialData = useCallback(
     (): SlashCommandsStreamState => ({


crates/executors/src/executors/codex/slash_commands.rs
```diff
@@ -0,0 +1,591 @@
...
+    pub async fn spawn_slash_command(
+        &self,
+        current_dir: &Path,
+        prompt: &str,
+        session_id: Option<&str>,
+        env: &ExecutionEnv,
+    ) -> Result<SpawnedChild, ExecutorError> {
+        if let Some(command) = CodexSlashCommand::parse(prompt) {
+            return match command {
+                CodexSlashCommand::Init => {
```
LSRCT, line 71: Is this the same behaviour as in the CLI? It's not using any built in functionality right, just a prompt?

crates/executors/src/executors/codex/slash_commands.rs
```diff
@@ -0,0 +1,591 @@
...
+            "compact" => Some(Self::Compact {
+                instructions: if cmd.arguments.is_empty() {
+                    None
+                } else {
+                    Some(cmd.arguments.to_string())
+                },
+            }),
+            "status" => Some(Self::Status),
+            "mcp" => Some(Self::Mcp),
+            "logout" => Some(Self::Logout),
```
LSRCT: should we support this? I think the login helper may not work with the new UI. IMO logout shouldnt be one of the 5 commands we support, especially if there is no way to log back in

crates/executors/src/executors/codex/slash_commands.rs
```diff
@@ -0,0 +1,591 @@
...
+                }
+            };
+            let _ = exit_signal_tx.send(exit_result);
+        });
+
+        spawned.exit_signal = Some(exit_signal_rx);
+        Ok(spawned)
+    }
+
+    async fn compact_inner(
```
LSRCT, line 181: this is very different to how we usually start a new codex session (v2 api?), is there a reason for that?

crates/executors/src/executors/opencode/slash_commands.rs
```diff
@@ -0,0 +1,614 @@
...
+            reorder_slash_commands,
+        },
+    },
+};
+
+/// OpenCode slash command with known variants and custom fallback.
+#[derive(Debug, Clone)]
+pub enum OpencodeSlashCommand {
+    Compact,
+    Share,
```
LSRCT: Does this work? I get a link and `Error | Uncaught Client Exception` when opened, also don't see how it would work

crates/executors/src/executors/utils.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Executors can use this key to cache expensive slash command retrievals.
+pub struct SlashCommandCache {
+    cache: Mutex<LruCache<SlashCommandCacheKey, CachedEntry>>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct SlashCommandCacheKey {
+    path: PathBuf,
+    executor_id: &'static str,
```
LSRCT: is this a BaseCodingAgent? Should we type it?

crates/server/src/routes/config.rs
```diff
@@ -485,3 +496,77 @@ async fn check_agent_availability(
 
     ResponseJson(ApiResponse::success(info))
 }
+
+#[derive(Debug, Deserialize)]
+pub struct AgentSlashCommandsStreamQuery {
+    executor: BaseCodingAgent,
+    #[serde(default)]
+    task_attempt_id: Option<Uuid>,
```
LSRCT: is this a workspace_id?

crates/services/src/services/container.rs
```diff
@@ -96,6 +97,59 @@ pub trait ContainerService {
 
     fn workspace_to_current_dir(&self, workspace: &Workspace) -> PathBuf;
 
+    async fn available_agent_slash_commands(
+        &self,
+        executor_profile_id: ExecutorProfileId,
+        workspace_id: Option<Uuid>,
+        repo_id: Option<Uuid>,
```
LSRCT, line 104: is this ever used?

crates/services/src/services/container.rs
```diff
@@ -96,6 +97,59 @@ pub trait ContainerService {
 
     fn workspace_to_current_dir(&self, workspace: &Workspace) -> PathBuf;
 
+    async fn available_agent_slash_commands(
+        &self,
+        executor_profile_id: ExecutorProfileId,
+        workspace_id: Option<Uuid>,
```
LSRCT, line 103: this seems to not be set for anything involving the new ui

crates/executors/src/executors/codex/slash_commands.rs
abcpro1: This is identical to the tui implementation

crates/executors/src/executors/codex/slash_commands.rs
abcpro1: fair point, we can add back later with a login pair

crates/executors/src/executors/codex/slash_commands.rs
abcpro1: there is no straightforward way for implementing compact so you get what you get

crates/executors/src/executors/opencode/slash_commands.rs
abcpro1: it should work. I'll look into the bug

crates/executors/src/executors/utils.rs
abcpro1: no exactly. just an internal slug, we could derive it from BaseCodingAgent with a constructor

crates/server/src/routes/config.rs
abcpro1: yes should be renamed

crates/services/src/services/container.rs
abcpro1: yes. for deriving repo-level /commands when there is no worktree yet

crates/services/src/services/container.rs
abcpro1: Interesting, what do you mean?

LSRCT: `workspace_id: Option<Uuid>` does not seem to be set ever from the new ui

crates/services/src/services/container.rs
LSRCT: `repo_id: Option<Uuid>` does not seem to be set ever, maybe i missed it? I think we just fall back to home.

crates/services/src/services/container.rs
abcpro1: might have been a rebase regression will take a look

crates/executors/src/executors/opencode/slash_commands.rs
abcpro1: removed. share and unshare rely on a constant session id, they don't really work with our automatic forking

crates/services/src/services/container.rs
abcpro1: You are correct. fixed.

frontend/src/hooks/useSlashCommands.ts
```diff
@@ -0,0 +1,57 @@
...
+  commands: SlashCommandDescription[];
+  discovering: boolean;
+  error: string | null;
+};
+
+export function useSlashCommands(
+  agent: BaseCodingAgent | null | undefined,
+  opts?: { workspaceId?: string; repoId?: string }
+) {
+  const endpoint = useMemo(() => {
```
LSRCT, line 17: should this be in api.ts

frontend/src/components/ui-new/containers/CreateChatBoxContainer.tsx
```diff
@@ -104,6 +104,8 @@ export function CreateChatBoxContainer() {
   // Get project ID from context
   const projectId = selectedProjectId;
 
+  const repoId = repos[0]?.id;
```
LSRCT: for a multi repo workspace the agent wouldn't run in repos[0] and might not have access to slash commands defined there. Not really sure what the alternative is either, at this point we don't know what directory the agent is going to run in

abcpro1: the initial idea was to select repo if there is only one, I guess let's do that.

frontend/src/hooks/useSlashCommands.ts
abcpro1: moved

PR #2196, opened 2026-01-21
fix: stop button not working during approval mode in workspace conversations

vdmkotai:
## Problem
When viewing a pending approval in workspace conversations, clicking the Stop button did nothing. The execution continued and the UI remained stuck in approval mode. Additionally, logs sometimes failed to display after stopping.

## Root Causes Identified

1. **Frontend guard condition too strict**: `stopExecution` had `!isAttemptRunning` check that blocked the call during approval mode due to timing issues with process status detection.

2. **Approval entry not updated on stop**: When killing a process, the backend never updated the pending approval's tool status, so the UI still showed `pending_approval` and remained in approval mode.

3. **Executor creating 'timed out' message**: When we cancelled approvals by sending `ApprovalStatus::TimedOut`, the executor received this and created a misleading 'Approval timed out' entry before being killed.

4. **Slow stop (4-6 seconds)**: `kill_process_group` sent SIGINT → 2s wait → SIGTERM → 2s wait → SIGKILL. During approval, the executor is blocked on I/O and doesn't respond to graceful signals.

5. **Logs not reloading**: When a process transitioned from 'running' to 'killed', the frontend didn't reload entries from the database, leaving potentially incomplete streamed data.

### Frontend Changes

**useAttemptExecution.ts**: Removed `!isAttemptRunning` guard. User intent to stop should always be honored; backend handles no-op gracefully.

**useConversationHistory.ts**: Added status transition detection that reloads entries when a process changes from 'running' to another status, ensuring complete logs are displayed after stop.

### Backend Changes (Rust)

**approvals.rs**: Added two methods:
- `has_pending_for_execution_process()`: Checks if pending approvals exist
- `cancel_for_execution_process()`: Updates tool status to `Failed` in MsgStore without sending approval response (prevents executor from creating 'timed out' message)

**container.rs**: Bifurcated stop logic:
- With pending approvals: Kill immediately with SIGKILL → then cancel approvals (process is dead, can't create spurious messages)
- Without pending approvals: Graceful shutdown (SIGINT → SIGTERM → SIGKILL with 5s timeout) for clean process termination

**command.rs**: Added `kill_process_group_force()` that skips graceful signals and goes straight to SIGKILL.

## Why This Approach

1. **Kill before cancel**: We must kill the process BEFORE cancelling approvals. Otherwise, dropping the approval channel causes the executor to receive `TimedOut` and create a misleading message.

2. **SIGKILL for approval stops**: The executor is blocked waiting for approval (I/O wait) and won't respond to SIGINT/SIGTERM. Going straight to SIGKILL makes stop instant instead of 4-6 seconds.

3. **ToolStatus::Failed vs TimedOut**: We set `Failed` (not `TimedOut`) in MsgStore because the user explicitly stopped - it wasn't a timeout.

4. **Entry reloading**: WebSocket streaming during execution may have incomplete data. Reloading from database after stop ensures complete logs are shown.


stunningpixels:
!reviewfast

frontend/src/components/ui-new/hooks/useConversationHistory.ts
```diff
@@ -41,6 +41,9 @@ export const useConversationHistory = ({
   const loadedInitialEntries = useRef(false);
   const streamingProcessIdsRef = useRef<Set<string>>(new Set());
   const onEntriesUpdatedRef = useRef<OnEntriesUpdated | null>(null);
+  const previousStatusMapRef = useRef<Map<string, ExecutionProcessStatus>>(
```
stunningpixels, line 47: Is this different from displayedExecutionProcesses ?

stunningpixels:
https://github.com/user-attachments/assets/a36f3b1f-bd62-498f-b7aa-df2fc0fc4292

I don't see the logs when I stop a plan

frontend/src/components/ui-new/hooks/useConversationHistory.ts
vdmkotai: Yes, they are different and both are needed.
`displayedExecutionProcesses` stores ExecutionProcessState which contains ExecutionProcessStaticInfo. Looking at the type definition in types.ts:
```
export type ExecutionProcessStaticInfo = {
  id: string;
  created_at: string;
  updated_at: string;
  executor_action: ExecutorAction;
};
```
It does not include status. This is intentional - it only stores immutable/static data about the process.
previousStatusMapRef tracks ExecutionProcessStatus (running, killed, completed, failed) for each process ID, which we need to detect when a process transitions from running to another status so we can reload its entries from the database.

We can't derive the previous status from displayedExecutionProcesses because it doesn't store status information.

stunningpixels: Got it, should we extend ExecutionProcessStaticInfo with "last status" ?

This is the one file I'm keen to try and reduce complexity as much as possible.

vdmkotai: I considered this carefully, and I believe keeping previousStatusMapRef separate is actually the cleaner approach.

### 1. Different Data Sources and Lifecycles
`displayedExecutionProcesses` stores what we've loaded and are displaying—it's our "view state." The status we're tracking in `previousStatusMapRef` comes from `executionProcessesRaw` (live WebSocket data), not from displayed data. These have different update triggers:

* `displayedExecutionProcesses` updates when we load entries from DB
* Status updates on every WebSocket message

### 2. Semantic Mismatch
`ExecutionProcessStaticInfo` is intentionally named "Static" because it only contains immutable data (`id`, `created_at`, `updated_at`, `executor_action`). Adding a mutable `lastStatus` field would:

* Make the name misleading
* Mix concerns (immutable identity vs. dynamic state)

### 3. Adding `lastStatus` Would Increase Complexity
If we added status to `ExecutionProcessStaticInfo`, we'd need additional sync logic to keep it updated from `executionProcessesRaw` on every change. Currently, the effect at lines 640-686 is self-contained—it reads from `executionProcessesRaw`, compares with `previousStatusMapRef`, and triggers reloads. No sync needed.

### 4. What `previousStatusMapRef` Actually Tracks
It's not "last displayed status" - it's "last known status from the live data source." This distinction matters because:

* A process might be in `displayedExecutionProcesses` with stale/no status
* We specifically need to detect transitions in the live data

abcpro1:
!reviewfast

PR #2202, opened 2026-01-21
Require executor profiles for follow-ups and queue (Vibe Kanban)

LSRCT:
## Summary
- Require executor profile IDs for follow-up sends and queued messages across API + frontend flows.
- Persist follow-up drafts with executor profile metadata and derive defaults from latest execution processes.
- Enforce executor consistency against latest process or session.executor for follow-ups and queued consumption.

## Why
Follow-up execution now needs a concrete executor profile to avoid ambiguous or mismatched runs and to keep session metadata consistent when no prior execution exists.

## Implementation details
- `CreateFollowUpAttempt` and queue payloads now carry `executor_profile_id`, and queued follow-ups store that profile for later execution.
- Server and local container validate executor mismatches and initialize `session.executor` on first follow-up.
- Frontend send/queue paths pass executor + variant and block actions when no executor is available.

- [x] tested

### DIFF
--- a/crates/local-deployment/src/container.rs
+++ b/crates/local-deployment/src/container.rs
@@ -823,25 +823,17 @@ impl LocalContainerService {
         let executor_profile_id = queued_data.executor_profile_id.clone();
 
         // Validate executor matches session if session has prior executions
-        if let Some(existing_profile) =
+        let expected_executor: Option<String> =
             ExecutionProcess::latest_executor_profile_for_session(&self.db.pool, ctx.session.id)
                 .await?
-        {
-            if existing_profile.executor != executor_profile_id.executor {
-                return Err(SessionError::ExecutorMismatch {
-                    expected: existing_profile.executor.to_string(),
-                    actual: executor_profile_id.executor.to_string(),
-                }
-                .into());
+                .map(|profile| profile.executor.to_string())
+                .or_else(|| ctx.session.executor.clone());
+
+        if let Some(expected) = expected_executor {
+            let actual = executor_profile_id.executor.to_string();
+            if expected != actual {
+                return Err(SessionError::ExecutorMismatch { expected, actual }.into());
             }
-        } else if let Some(session_executor) = ctx.session.executor.as_ref()
-            && session_executor != &executor_profile_id.executor.to_string()
-        {
-            return Err(SessionError::ExecutorMismatch {
-                expected: session_executor.to_string(),
-                actual: executor_profile_id.executor.to_string(),
-            }
-            .into());
         }
 
         if ctx.session.executor.is_none() {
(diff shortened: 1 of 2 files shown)


LSRCT:
!reviewfast

crates/local-deployment/src/container.rs
```diff
@@ -821,33 +820,38 @@ impl LocalContainerService {
...
-        let base_executor = match ExecutionProcess::latest_executor_profile_for_session(
-            &self.db.pool,
-            ctx.session.id,
-        )
-        .await
-        .map_err(|e| ContainerError::Other(anyhow!("Failed to get executor profile: {e}")))?
+        let executor_profile_id = queued_data.executor_profile_id.clone();
+
+        // Validate executor matches session if session has prior executions
+        if let Some(existing_profile) =
```
ggordonhall: Can we simplify this condition like so:

```
let expected_executor: Option<String> = ExecutionProcess::latest_executor_profile_for_session(
     &self.db.pool, ctx.session.id
 )
     .await?
     .map(|profile| profile.executor.to_string())
     .or_else(|| ctx.session.executor.clone());

 if let Some(expected) = expected_executor {
     let actual = executor_profile_id.executor.to_string();
     if expected != actual {
         return Err(SessionError::ExecutorMismatch { expected, actual }.into());
     }
 }
```

crates/server/src/routes/sessions/mod.rs
```diff
@@ -118,31 +115,31 @@ pub async fn follow_up(
...
-        };
-
-    let executor_profile_id = ExecutorProfileId {
-        executor: base_executor,
-        variant: payload.variant,
-    };
+    let executor_profile_id = payload.executor_profile_id;
+
+    // Validate executor matches session if session has prior executions
+    if let Some(existing_profile) =
```
ggordonhall: Similarly:

```
let expected_executor: Option<String> = ExecutionProcess::latest_executor_profile_for_session(pool, session.id)
      .await?
      .map(|profile| profile.executor.to_string())
      .or_else(|| session.executor.clone());

  if let Some(expected) = expected_executor {
      let actual = executor_profile_id.executor.to_string();
      if expected != actual {
          return Err(ApiError::Session(SessionError::ExecutorMismatch { expected, actual }));
      }
  }
```

crates/server/src/routes/task_attempts.rs
```diff
@@ -1406,9 +1406,7 @@ pub async fn run_setup_script(
         None => {
             Session::create(
                 pool,
-                &CreateSession {
-                    executor: Some("setup-script".to_string()),
```
ggordonhall, line 1410: 😮

PR #2217, opened 2026-01-22
Enhance default opencode's MCP config path retrieval logic

kerlw:
Added logic to check for configuration file in both current and alternative paths.

In windows, use bun to install opencode, the mcp config path would be ~/.config/opencode/opencode.json


abcpro1:
Superseded by #2379

PR #2247, opened 2026-01-22
feat: remote pr endpoint

ggordonhall:
Adds:
- `pull_requests` table and associated shape (no mutations)
- Create and update routes for `PullRequests` that will be called from the client backend. These can trigger changes to `Issue` status
- `organisation_members` shape for real time sync


stunningpixels:
Do we want to cache repo name too on PRs

ggordonhall:
> Do we want to cache repo name too on PRs

Yes, I though about that. Is repo name in the designs? Really we'd want some `Repo` entity in the remote for this to be meaningful.

PR #2270, opened 2026-01-23
feat: remote routes for workspaces

ggordonhall:
- Create remote workspace (from scratch, or from existing local workspace)
- Delete remote workspace by `local_workspace_id`
- Unlink remote workspace (delete by remote `workspace_id`)
- Update remote workspace by `local_workspace_id`
- Amend `CreatePullRequest` to create by `local_workspace_id`
- Automatically move associated `issues` into `In progress` when the first associated workspace is created


stunningpixels:
How/should we enforce that each local workspace can only be linked to a single remote workspace?

ggordonhall:
> How/should we enforce that each local workspace can only be linked to a single remote workspace?

There's a `UNIQUE` constraint on `local_workspace_id` in the DB.

PR #2289, opened 2026-01-23
Fix PR creation for Github enterprise repos

ranmocy:
When the repo is an enterprise repo, we need to specify the hostname to create the PR correctly.
Otherwise it will raise error of "Can't find the repo"

### DIFF
--- a/crates/services/src/services/git_host/github/cli.rs
+++ b/crates/services/src/services/git_host/github/cli.rs
@@ -31,8 +31,6 @@ pub struct GitHubRepoInfo {
 }
 
 impl GitHubRepoInfo {
-    /// Returns the repository specification in the format expected by gh CLI.
-    /// For GitHub Enterprise: "hostname/owner/repo", otherwise "owner/repo".
     pub fn repo_spec(&self) -> String {
         match &self.hostname {
             Some(host) => format!("{}/{}/{}", host, self.owner, self.repo_name),


crates/services/src/services/git_host/github/cli.rs
```diff
@@ -123,6 +125,45 @@ impl GhCli {
...
 
+    /// Extract hostname from a git remote URL.
+    ///
+    /// Supports:
+    /// - HTTPS: `https://host/owner/repo` or `https://host/owner/repo.git`
+    /// - SSH: `git@host:owner/repo.git`
+    /// - SSH protocol: `ssh://git@host/owner/repo.git`
+    ///
+    /// Returns `None` for `github.com` (the default host).
+    fn extract_hostname(remote_url: &str) -> Option<String> {
```
LSRCT: Can you change this to
- Use the url crate to parse it
- Return github.com - no need to throw that away

crates/services/src/services/git_host/github/cli.rs
```diff
@@ -171,21 +212,25 @@ impl GhCli {
         remote_url: &str,
         repo_path: &Path,
     ) -> Result<GitHubRepoInfo, GhCliError> {
-        let raw = self.run(
-            ["repo", "view", remote_url, "--json", "owner,name"],
-            Some(repo_path),
-        )?;
-        Self::parse_repo_info_response(&raw)
+        // gh repo view accepts the remote URL directly and resolves the host
+        let args = vec!["repo", "view", remote_url, "--json", "owner,name"];
```
LSRCT:
```suggestion
        let args = vec!["repo", "view", remote_url, "--json", "owner,name,url"];
```

crates/services/src/services/git_host/github/cli.rs
```diff
@@ -171,21 +212,25 @@ impl GhCli {
...
+        let raw = self.run(args, Some(repo_path))?;
+        let hostname = Self::extract_hostname(remote_url);
+        Self::parse_repo_info_response(&raw, hostname)
```
LSRCT: use url returned by cli to parse, it's always HTTPs so parsing the hostname is simpler

LSRCT (approved):

PR #2310, opened 2026-01-25
git diff refactoring

vdmkotai:


anastasiya1155:
!reviewfast

vdmkotai:
> Do we really need to sync file tree and diffs scrolling? I imagine it is a complex computation and is not working for me anyway. I think we shouldn't pursue it or change the approach e.g. use onMouseOver to make it more precise

Can you please check again? It should be better.

vdmkotai:
!reviewfast

PR #2324, opened 2026-01-26
feat: forgejo/gitea

zexa:
Adds support for Forgejo/Gitea git hosting providers, enabling users to create PRs, list PRs, fetch PR comments, and attach existing PRs for repositories hosted on Forgejo/Gitea instances (including Codeberg).

Previously, Vibe Kanban only supported GitHub and Azure DevOps for git hosting operations. Users with repositories on Forgejo/Gitea instances (self-hosted or public like Codeberg) couldn't use PR-related features. This PR adds first-class support for the Forgejo/Gitea API.

  - ForgejoProvider that implements all GitHostProvider trait methods via HTTP API
  - Automatic detection of well-known hosts: codeberg.org, gitea.com
  - Support for self-hosted instances via user configuration
  - Full PR lifecycle support: create, list, get status, fetch comments

  - New gitHosts config section for mapping custom domains to providers
  - Users can configure self-hosted Forgejo/Gitea instances with API tokens
  - Example config:
  
  ```json
  {
    "gitHosts": {
      "hosts": {
        "git.mycompany.com": {
          "provider": "forgejo",
          "token": "your-api-token"
        }
      }
    }
  }
  ```
  
  - Removed centralized detection.rs in favor of distributed detection logic

### DIFF
--- a/crates/services/src/services/git_host/forgejo/http.rs
+++ b/crates/services/src/services/git_host/forgejo/http.rs
@@ -98,25 +98,30 @@ pub fn extract_host(url: &str) -> Result<String, GitHostError> {
 /// Extract owner and repo from a git remote URL.
 /// Supports both HTTPS and SSH formats.
 pub fn parse_owner_repo(url: &str) -> Result<(String, String), GitHostError> {
+    let extract_from_path = |path: &str| -> Option<(String, String)> {
+        let mut parts = path
+            .trim_start_matches('/')
+            .trim_end_matches(".git")
+            .split('/');
+        let owner = parts.next()?;
+        let repo = parts.next()?;
+        if owner.is_empty() || repo.is_empty() {
+            return None;
+        }
+        Some((owner.to_string(), repo.to_string()))
+    };
+
     // Try to parse as URL first
     if let Ok(parsed) = url::Url::parse(url) {
-        let path = parsed
-            .path()
-            .trim_start_matches('/')
-            .trim_end_matches(".git");
-        let parts: Vec<&str> = path.split('/').collect();
-        if parts.len() >= 2 {
-            return Ok((parts[0].to_string(), parts[1].to_string()));
+        if let Some(result) = extract_from_path(parsed.path()) {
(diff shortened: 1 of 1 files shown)


LSRCT:
Thanks for the contribution.
We want to support more git host providers, to make that smooth in the future we should make same change to this.
My concern is mostly with the way configs are handled.

Change config structure from HashMap with struct to HashMap with tagged enum:
```rust
#[serde(tag = "type", rename_all = "snake_case")]
pub enum GitHostConfig {
    Forgejo(ForgejoConfig),
    GitHub(GitHubConfig),
    ...
}
pub struct ForgejoConfig { pub token: String }
pub struct GitHubConfig {}  // empty, uses CLI
// in the config:
pub git_hosts: HashMap<String, GitHostConfig>
```


Change provider construction from provider holding config to provider holding just what it needs:
```rust
pub struct ForgejoProvider {
    host: String,
    token: String,
}

impl ForgejoProvider {
    pub fn new(host: String, config: &ForgejoConfig) -> Self {
        Self { host, token: config.token.clone() }
    }
}
```


Change detection logic from distributed detection across providers to centralized detection:
```rust
pub fn for_url(url: &str, git_hosts: &HashMap<String, GitHostConfig>) -> Result<Self, GitHostError> {
    let host = extract_host(url)?;    
    // 1. Explicit config takes priority
    if let Some(config) = git_hosts.get(&host) {
        return Self::from_config(&host, config);
    }
    // 2. Fall back to URL detection (CLI providers only)
    return Self::from_url(url);
}
```


Ideally we'd also use the `gix_url` crate rather than manually parsing the git URLs.

Not sure if we need a config migration for this since we are just adding things, could probably keep v8.

In terms of the frontend ideally we'd have a new settings page for this, the general settings are already quite big.

Would be great if you could implement this, I'm of course open to help since it's quite a big change.
Thanks!

zexa:
MR stale, plus I made my own forgejo agent so no longer need this. Someone can take over.

PR #2327, opened 2026-01-26
feat(frontend): load project repos in workspace

Karrq:
This PR concerns the new workspaces UI.

In particular, when a new workspace is created, I noticed that the list of repositories wouldn't update when I switched projects, and I had to add repositories manually each time I wanted to switch to a new "set".

This change makes it so that the repositories associated with a project are loaded when the project is selected, so users (me included) can use a project as a "set of repositories" and easily switch betwen them when creating a workspace


Juanlucasbg:
## Rebase verification — PR #2327

- `frontend/src/components/ui-new/containers/CreateModeProjectSectionContainer.tsx`
- `frontend/src/contexts/CreateModeContext.tsx`
- `frontend/src/hooks/useCreateModeState.ts`

Pre-monorepo paths. The "load project repos in workspace" feature would need to be re-applied against the current tree (likely `packages/web-core/src/...` and possibly `packages/local-web/src/...`).

**Action**: this PR predates the monorepo restructure. Recommend re-doing as a fresh PR — the original 3-file change won't apply mechanically.

— Automated rebase-verification: detected delete-vs-modify across multiple paths before posting.

PR #2329, opened 2026-01-26
feat: use --resume-session-at for Claude reset operations (Vibe Kanban)

LSRCT:
## Summary

- Use Claude's `--resume-session-at` with message UUIDs for reset operations, keeping the same session ID (required for plan mode slug consistency)
- Track message UUIDs: user messages are always valid; assistant messages only after `Result` confirms no incomplete tool calls
- Add `--replay-user-messages` flag to capture UUIDs from cancel/interrupt scenarios
- Skip replayed messages in UI via `isReplay` field
- Rename `message_uuid` to `reset_to_message_uuid` to clarify intent
- Add `CancellationToken` for graceful approval cancellation

## Key Changes

- **Reset**: `--resume <session_id> --resume-session-at <uuid>` - truncates history
- **Normal follow-up**: `--resume <session_id>` only - continues conversation
- Message UUIDs stored in `coding_agent_turns.agent_message_uuid`

---


LSRCT:
!reviewfast

LSRCT:
!reviewfast

crates/utils/src/log_msg.rs
```diff
@@ -15,6 +16,7 @@ pub enum LogMsg {
     Stderr(String),
     JsonPatch(Patch),
     SessionId(String),
+    MessageUuid(String),
```
abcpro1: Rename to `MessageId` instead of `MessageUuid`.
Unique IDs only need to be a string and be unique per message, they don't need to respect the UUID format. Hard coding implementation details in interfaces make it hard to extend to other vendors in the future.

crates/executors/src/logs/mod.rs
```diff
@@ -152,6 +152,7 @@ pub enum ToolStatus {
         timeout_at: DateTime<Utc>,
     },
     TimedOut,
+    Cancelled,
```
abcpro1: sneaky enum variant? doesn't seem to be used.

crates/utils/src/log_msg.rs
LSRCT: Changed

crates/executors/src/logs/mod.rs
LSRCT: You're right, removed it.

crates/executors/src/executors/claude.rs
```diff
@@ -121,6 +122,7 @@ impl ClaudeCode {
             "--output-format=stream-json",
             "--input-format=stream-json",
             "--include-partial-messages",
+            "--replay-user-messages",
```
abcpro1, line 125: is this flag needed for this feature?

LSRCT: It is. For the good case where claude finishes normally we don't need it, but if claude gets killed by something there may be no resumable assistant message. This flag will make sure we can resume to an appropriate message

abcpro1: Does this mean that if Claude is killed all messages from the turn are lost?

LSRCT: Maybe this needs more context. Not every message id can be used for resuming, only ids from user messages and some assistant messages can be used. That's not documented, but empirically the case.

Assistant message ids work if they are at the end of a turn (right before a user message or right before a result message). We very likely wont have a valid one if the process gets killed.

Adjacent topic in the cc sdk docs:
https://platform.claude.com/docs/en/agent-sdk/file-checkpointing
It's not quite the same, it resets files rather than messages, but it's the only place they talk about using message ids at all. There they recommend enabling that flag.

PR #2353, opened 2026-01-27
Move git service to the new crate (vibe-kanban)

LSRCT:
Moves GitService to a new crate.

Reasoning:
- crate:Services cannot be used by executors crate (circular import) but need it, currently we have crates/utils/src/git.rs as a workaroundGit
- GitService does not rely on db access or similar that would justify being in services

### DIFF
--- a/crates/deployment/src/lib.rs
+++ b/crates/deployment/src/lib.rs
@@ -13,6 +13,7 @@ use db::{
 };
 use executors::executors::ExecutorError;
 use futures::{StreamExt, TryStreamExt};
+use git::{GitService, GitServiceError};
 use git2::Error as Git2Error;
 use serde_json::Value;
 use services::services::{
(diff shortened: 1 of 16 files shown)


crates/deployment/src/lib.rs
```diff
@@ -37,6 +36,7 @@ use sqlx::Error as SqlxError;
 use thiserror::Error;
 use tokio::sync::RwLock;
 use utils::sentry as sentry_utils;
+use workspace_git::{GitService, GitServiceError};
```
ggordonhall: Why is it called `workspace_git` everywhere instead of just `git`?

LSRCT: To avoid confusion with git2. This way it's easy to check nothing but our git crate is used when reviewing things.
Should I rename it?

ggordonhall: yes please

PR #2366, opened 2026-01-28
feat: add default_working_dir setting to repositories for monorepo support (Vibe Kanban)

LSRCT:
## Summary

Re-adds the ability to configure a default working directory for repositories, improving monorepo support. This setting was previously removed in PR #1874 when it was replaced with auto-calculation.

- Adds `default_working_dir` field to the Repo model
- Users can specify a subdirectory (relative to repo root) where the coding agent should run
- Useful for monorepos where the agent should operate in a specific package/app directory

### Backend
- **Database migration**: Added `default_working_dir` column to `repos` table
- **Repo model**: Updated `Repo` struct and `UpdateRepo` to include the new field
- **SQL queries**: Updated all queries in `repo.rs`, `project_repo.rs`, `workspace_repo.rs`
- **Workspace creation**: Modified `tasks.rs` and `task_attempts.rs` to use `PathBuf::join()` for proper path construction:
  - Single-repo with `default_working_dir` set → `repo_name/configured_subdir`
  - Single-repo without setting → `repo_name` (unchanged from before)
  - Multi-repo → workspace root (unchanged)

### Frontend
- Added settings field in the new UI (`ReposSettingsSection.tsx`)
- Added i18n translations for all 7 locales (en, es, fr, ja, ko, zh-Hans, zh-Hant)

## Implementation Details

The path is stored relative to the repo root, then combined with the repo name at workspace creation time using `PathBuf::join()` to ensure proper cross-platform path handling. This means if a user sets `packages/frontend`, the agent will run in `workspace_root/repo_name/packages/frontend`.

The setting only applies to **single-repo workspaces**. Multi-repo workspaces continue to use the workspace root.

- [x] tested

---

### DIFF
--- a/frontend/src/components/ui-new/dialogs/settings/ReposSettingsSection.tsx
+++ b/frontend/src/components/ui-new/dialogs/settings/ReposSettingsSection.tsx
@@ -306,7 +306,9 @@ export function ReposSettingsSection() {
             >
               <SettingsInput
                 value={draft.default_working_dir}
-                onChange={(value) => updateDraft({ default_working_dir: value })}
+                onChange={(value) =>
+                  updateDraft({ default_working_dir: value })
+                }
                 placeholder={t(
                   'settings.repos.general.defaultWorkingDir.placeholder'
                 )}


lee-b:
I'm curious: is this intended to limit vibe-kanban to ONLY operating in the subdirectory? Isn't the point of a monorepo that you can modify code across the system-of-systems in a single commit, where there are cross-cutting changes needed?

LSRCT:
It's so that coding agent starts off in the right directory. You might be a frontend dev on "project9861" in a large repo, you can set it up so claude runs in `repo/project9861/frontend` and doesn't get confused by other things.

The agent can still make changes in other places, this PR just adds a setting for the agent working directory

PR #2373, opened 2026-01-28
fix: make comment author nullable

ggordonhall:
Tweak the schema making comment author a nullable field. This avoids a `CASCADE` that would delete comments from the database when a user was deleted.

This edits a migration file. We'll have to wipe the remote dev DB.


ggordonhall:
!reviewfast

PR #2374, opened 2026-01-28
feat(git): add merge strategy setting to allow diverged branches (Vibe Kanban)

jtomaszewski:
## Summary

Adds a new global setting "Merge Strategy" that controls how merges behave when the base branch has advanced since the task branch was created.

**Options:**
- **Strict** (default): Requires manual rebase before merge if base branch is ahead
- **Allow diverged**: Allows merge to proceed if no actual conflicts exist

## Why

Previously, merging would fail with "branches have diverged" error whenever the base branch had new commits, even if those commits didn't conflict with the task's changes. This forced users to manually rebase before every merge, which can be tedious for non-conflicting changes.

With the new "Allow diverged" strategy, the merge will succeed as long as there are no actual file conflicts - the existing atomic 3-way merge handles this correctly.

### Backend
- **`crates/git/src/lib.rs`**: Added `MergeStrategy` enum and updated `merge_changes()` to conditionally skip the early divergence check based on strategy
- **`crates/services/src/services/config/versions/v9.rs`**: New config version with `merge_strategy` field
- **`crates/server/src/routes/task_attempts.rs`**: Reads config and passes strategy to merge operation

### Frontend
- **`frontend/src/pages/settings/GeneralSettings.tsx`**: Added Merge Strategy dropdown in the Git settings section
- **`frontend/src/i18n/locales/en/settings.json`**: Added translations

### Tests
- Updated all `merge_changes()` call sites in test files to include the new strategy parameter

## Implementation Details

The existing `perform_squash_merge()` already does an atomic in-memory 3-way merge using `repo.merge_commits()`. The divergence check was an early bailout that prevented reaching this code. With `AllowDiverged`, we simply skip the early check and let the atomic merge proceed - if there are actual conflicts, the merge fails atomically without any partial state.

---


Juanlucasbg:
## Rebase verification — PR #2374

- `crates/git/tests/git_workflow.rs`
- `crates/server/src/routes/task_attempts.rs`
- `frontend/src/pages/settings/GeneralSettings.tsx` (pre-monorepo path)
- `shared/types.ts` (generated)

Mixed conflict: pre-monorepo path + active server/test file + generated types. The git_workflow.rs test file was also touched by PR #3335 (rebase + merge-commit strategies).

**Action**: rebase. The `git_workflow.rs` conflict is on test fixtures — easy to merge mechanically. The pre-monorepo `GeneralSettings.tsx` needs relocation to `packages/web-core/src/...`. Regenerate `shared/types.ts` rather than hand-editing.

— Automated rebase-verification: detected mixed-path + generated-file conflict before posting.

Issue #2372, opened 2026-01-28
feat: Can the cow sound be changed from default?


albinantoab:

lee-b:
Just checking, that you know it can be changed in the settings. You probably mean that the default should be something else, even without going into the settings. I'd tend to agree.

albinantoab:
@lee-b yeah, just suggesting to keep something else as the default option. I'm aware I can change it in settings.

ggordonhall:
It's now possible to choose the notification sound in the onboarding screen. The cow is no longer the default.

Issue #2292, opened 2026-01-23
Remove the default Cow as the default sound.


TheKingYouNeed:
The default cow is not funny, it's just annoying and very unprofessional. Just keep it but don't make it the default. And it's very loud, it just makes problems.

marmoure:
@TheKingYouNeed They have many options, I personally like the cow.

Raymond-Hear:
Totally agree! The first time I used it I was really shocked by that sound effect, and anyone wearing headphones would be immediately knocked out.

alabhyajindal:
Same! Abstract Sound 1 sounds much nicer.

bash-bandicoot:
Cow is scary even! It came out of nowhere (first-time user) and caught me off guard!

LSRCT:
🐮

ggordonhall:
It's no longer the default sound. You can choose your sound during the onboarding.

Issue #1862, opened 2026-01-09
feat: would like to disable/configure the 🐮 MOO on task completion


eseidel:
It's driving my wife crazy. 🤣

maktouch:
<img width="1517" height="952" alt="Image" src="https://github.com/user-attachments/assets/bf74c52c-61f3-4261-a06f-4ea910abdc3a" />

LSRCT:
🐮

eseidel:
Thanks.  Apologies for the noise.

PR #2397, opened 2026-01-29
Fix cow default sound

LSRCT:
Request here:
https://github.com/BloopAI/vibe-kanban/issues/2372
https://github.com/BloopAI/vibe-kanban/issues/2292
https://github.com/BloopAI/vibe-kanban/issues/1862

### DIFF
--- a/crates/services/src/services/config/versions/v2.rs
+++ b/crates/services/src/services/config/versions/v2.rs
@@ -173,7 +173,7 @@ impl Default for NotificationConfig {
         Self {
             sound_enabled: true,
             push_enabled: true,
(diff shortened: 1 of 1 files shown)


crates/services/src/services/config/versions/v2.rs
```diff
@@ -173,7 +173,7 @@ impl Default for NotificationConfig {
         Self {
             sound_enabled: true,
             push_enabled: true,
-            sound_file: SoundFile::CowMooing,
+            sound_file: SoundFile::Rooster,
```
albinantoab, line 176: Can this be vibrate?

PR #2406, opened 2026-01-29
Trigger manual path load on Enter (Vibe Kanban)

LordMike:
## Summary
- Add Enter key handling to the manual path input in the folder picker dialog.
- Keep behavior aligned with the existing "Go" button action.

## Why
On Windows, pressing Enter in the "Enter path manually" input did nothing, which made it seem like manual path entry was unsupported. The task is to ensure Enter triggers the same directory load as clicking Go.

## Implementation details
- Added an `onKeyDown` handler to the manual path input in `frontend/src/components/dialogs/shared/FolderPickerDialog.tsx`.
- When the key is Enter, it prevents the default and calls the existing `handleManualPathSubmit`, which already invokes `loadDirectory(manualPath)`.
- No changes to the underlying file system API or directory listing logic.

### DIFF
--- a/frontend/src/components/dialogs/shared/FolderPickerDialog.tsx
+++ b/frontend/src/components/dialogs/shared/FolderPickerDialog.tsx
@@ -111,13 +111,22 @@ const FolderPickerDialogImpl = NiceModal.create<FolderPickerDialogProps>(
       // Don't set manual path here since home directory path varies by system
     };
 
-    const handleManualPathChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-      setManualPath(e.target.value);
-    };
+  const handleManualPathChange = (e: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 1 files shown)


LordMike:
Code on this project must move so fast if everyone is using vibe-kanban to do work on vibe-kanban.. :|

Juanlucasbg:
## Rebase verification — PR #2406 — **rebases cleanly**

Local verification:
- `git rebase main` — clean (no conflicts).
- `pnpm --filter @vibe/web-core run check` — same TS errors reproduce on `main`, so they are environmental (likely a missing dev-only dep in my local install) rather than PR-caused.

**Action**: push the rebased branch (`git rebase main && git push --force-with-lease`). CI is the authoritative compile signal — local TS errors here are not introduced by this PR.

— Automated rebase-verification: confirmed clean rebase locally; reproduced the same TS errors on `main` to rule out PR causation.

PR #2420, opened 2026-01-29
partial fix: Handle replace ops as upserts in patch streams

abcpro1:
Fixed conversation history tearing issues.

- [x] tested

Before

before

After

after


stunningpixels:
Is this a bandaid, should the backend conform to json patch?

stunningpixels (approved):

abcpro1:
> Is this a bandaid, should the backend conform to json patch?

Yes, it's harder to track in the backend. If json-patch had a native upsert I would prefer to is as it is safer.

There are other sources of this tearing issue, which I will look into next.

PR #2445, opened 2026-01-30
fix: use ScriptContext to determine run_reason for chained scripts

jtomaszewski:
## Summary
- Fix multi-repo project tasks getting stuck in 'in progress' status instead of transitioning to 'in review' after prompt completion
- When cleanup scripts chain to other cleanup scripts, the run_reason is now correctly set based on `ScriptContext` instead of always using `SetupScript`

## Test plan
- [x] Verified manually with multi-repo project having cleanup scripts
- [x] All existing tests pass

### DIFF
--- a/crates/services/src/services/container.rs
+++ b/crates/services/src/services/container.rs
@@ -1209,8 +1209,13 @@ pub trait ContainerService {
 
         // Determine the run reason of the next action
         let next_run_reason = match (action.typ(), next_action.typ()) {
-            (ExecutorActionType::ScriptRequest(_), ExecutorActionType::ScriptRequest(_)) => {
-                ExecutionProcessRunReason::SetupScript
+            (ExecutorActionType::ScriptRequest(_), ExecutorActionType::ScriptRequest(next_script)) => {
+                match next_script.context {
+                    ScriptContext::CleanupScript => ExecutionProcessRunReason::CleanupScript,
+                    ScriptContext::SetupScript => ExecutionProcessRunReason::SetupScript,
+                    ScriptContext::DevServer => ExecutionProcessRunReason::DevServer,
+                    ScriptContext::ToolInstallScript => ExecutionProcessRunReason::SetupScript,
+                }
             }
             (
                 ExecutorActionType::CodingAgentInitialRequest(_)


Juanlucasbg:
## Aggressive review summary — PR #2445

The fix premise is correct (using `ScriptContext` to drive `ExecutionProcessRunReason` for chained scripts is the right call), but **the PR will not compile after merging into current `main` — needs a rebase plus one missing arm**.

### HIGH — semantic merge break

PR branch has `crates/executors/src/actions/script.rs` with **4 variants**:
```rust
pub enum ScriptContext {
    SetupScript, CleanupScript, DevServer, ToolInstallScript,
}
```

Current `main` has **5 variants** — `ArchiveScript` was added in [#2438](https://github.com/BloopAI/vibe-kanban/pull/2438) ("feat: add archive script that runs when workspace is archived"):
```rust
pub enum ScriptContext {
    SetupScript, CleanupScript, ArchiveScript, DevServer, ToolInstallScript,
}
```

GitHub reports `mergeable: MERGEABLE` because this PR doesn't touch `script.rs` — no textual conflict. But the post-merge code lands a 4-arm exhaustive `match` on a 5-variant enum (no `_` arm), which fails the `non_exhaustive_patterns` lint and breaks the build.

I verified locally: on the PR branch as-is, `cargo check -p services` passes (exit 0) because the PR branch's `script.rs` still has the old 4 variants. After rebase onto main, that compile passes again only if the new arm is added.

### Required fix

Rebase onto `main`, then add the missing arm. The `ExecutionProcessRunReason` enum in `crates/db/src/models/execution_process.rs:53` already has `ArchiveScript`, so the natural mapping is symmetric:

```rust
match next_script.context {
    ScriptContext::CleanupScript => ExecutionProcessRunReason::CleanupScript,
    ScriptContext::SetupScript   => ExecutionProcessRunReason::SetupScript,
    ScriptContext::ArchiveScript => ExecutionProcessRunReason::ArchiveScript,  // <- add
    ScriptContext::DevServer     => ExecutionProcessRunReason::DevServer,
    ScriptContext::ToolInstallScript => ExecutionProcessRunReason::SetupScript,
}
```

### MED — semantic question

`ToolInstallScript → SetupScript`. Tool install ≈ setup phase, so the bucketing is reasonable. Worth confirming the maintainer wants tool installs telemetry'd as setup-script runs rather than getting their own bucket. Non-blocking.

### Verdict

Approve in principle, **block merge until rebased + ArchiveScript arm added**. With those two changes, this is a clean improvement over the current hardcoded `SetupScript`.

— Reviewed by automated single-pass review (caught semantic-merge break via local rebase simulation).

PR #1661, opened 2025-12-28
Add option to delete git branches when deleting a task (Vibe Kanban)

stunningpixels:
## Summary

Adds a checkbox to the delete task confirmation dialog that allows users to optionally delete the associated git branches when deleting a task. The checkbox defaults to unchecked to prevent accidental branch deletion.

### Frontend
- **DeleteTaskConfirmationDialog.tsx**: Added a checkbox with label "Also delete associated git branches" that controls whether branches should be deleted along with the task
- **api.ts**: Updated `tasksApi.delete()` to accept an optional `deleteBranches` parameter, sent as `?delete_branches=true` query param
- **i18n**: Added translations for the delete task dialog in all 5 supported languages (English, Spanish, Japanese, Korean, Simplified Chinese)

### Backend
- **git/cli.rs**: Added `delete_branch()` method that runs `git branch -D <branch_name>`
- **git.rs**: Added wrapper `delete_branch()` method in GitService
- **workspace_manager.rs**: Added `delete_branches()` helper method to encapsulate branch deletion logic in the service layer
- **tasks.rs**: Added `DeleteTaskQuery` struct with `delete_branches` field; calls `WorkspaceManager::delete_branches()` in background cleanup task

## Implementation Details

- Branch deletion happens **after** worktree cleanup in a background task (branches can't be deleted while worktrees are using them)
- Uses `git branch -D` (force delete) since task branches may not be fully merged
- Each branch is deleted from ALL repos (supports multi-repo projects)
- Branch deletion failures are logged as warnings but don't fail the overall task deletion
- The checkbox defaults to unchecked to ensure backwards-compatible behavior

---

This PR was written using [Vibe Kanban](https://vibekanban.com)

Screenshot 2025-12-28 at 15 14 45

PR #2519, opened 2026-02-03
feat: Add option to delete branch when deleting workspace (Vibe Kanban)

ggordonhall:
## Summary

Adds an optional checkbox in the workspace deletion dialog that allows users to delete the associated git branch(es) when deleting a workspace.

### Backend
- **`crates/git/src/cli.rs`**: Added `delete_branch()` method to GitCli for force-deleting local branches
- **`crates/server/src/routes/task_attempts.rs`**: 
  - Extended `DeleteWorkspaceQuery` with `delete_branches` parameter
  - Modified `delete_workspace()` to delete branches from all repos after worktree cleanup

### Frontend
- **`DeleteWorkspaceDialog.tsx`** (new): Custom dialog with a checkbox to optionally delete the branch
  - Checkbox is disabled when there's an open PR (with explanation text)
  - Uses existing `useBranchStatus` hook to check PR status
- **`api.ts`**: Updated `attemptsApi.delete()` to accept optional `deleteBranches` parameter
- **`actions/index.ts`**: Updated `DeleteWorkspace` action to use the new dialog

## Why

Users requested the ability to clean up branches when deleting workspaces. This is particularly useful after merging PRs or when branches are no longer needed. The feature is scoped to explicit user deletion only (not automatic cleanup/expiration).

## Implementation Details

- Branch deletion happens in the background cleanup task, **after** worktree cleanup (git requirement)
- Uses force delete (`-D`) since the user has explicitly chosen to delete
- Branch deletion failures are logged as warnings (non-fatal) since the branch might not exist
- The checkbox is automatically disabled when an open PR exists to prevent accidental deletion of active branches

---

Supersedes https://github.com/BloopAI/vibe-kanban/pull/1661

### DIFF
--- a/crates/server/src/routes/task_attempts.rs
+++ b/crates/server/src/routes/task_attempts.rs
@@ -40,7 +40,7 @@ use executors::{
     executors::{CodingAgent, ExecutorError},
     profile::{ExecutorConfigs, ExecutorProfileId},
 };
-use git::{ConflictOp, GitCli, GitCliError, GitServiceError};
+use git::{ConflictOp, GitCliError, GitService, GitServiceError};
 use git2::BranchType;
 use serde::{Deserialize, Serialize};
 use services::services::{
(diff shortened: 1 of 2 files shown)


crates/server/src/routes/task_attempts.rs
```diff
@@ -41,7 +41,7 @@ use executors::{
     executors::{CodingAgent, ExecutorError},
     profile::{ExecutorConfigs, ExecutorProfileId},
 };
-use git::{ConflictOp, GitCliError, GitServiceError};
+use git::{ConflictOp, GitCli, GitCliError, GitServiceError};
```
LSRCT: GitCli is only be used by GitService currently. It's public because of the tests.
Can we wrap it in delete_branch in GitService?

LSRCT (commented):
Works!

PR #2548, opened 2026-02-05
chore: bump agent versions

ggordonhall:
- Use ACP for Github Copilot


Stoneeu:
@ggordonhall  @abcpro1  ACP for Github Copilot Cli is only on preview, it don't support many paramater like : 
npx -y @github/copilot@0.0.403 --allow-all-tools --model claude-sonnet-4.5 --acp --agent=custom-agent --yolo

Now vibe_kanban new version copilot cli not support --agent --yolo and other custom paramaters ,  could you please roll back, and waiting  Github Copilot cli to  support full ACP , thanks

PR #2588, opened 2026-02-06
perf: replace structuredClone with immer produce and rAF batching in WS stream (Vibe Kanban)

stunningpixels:
## Summary

Profiling revealed that the frontend's WebSocket message handler in `streamJsonPatchEntries.ts` was consuming **88.6% of main thread budget** when streaming normalized logs. The root cause was `structuredClone(snapshot)` being called on every WS message (~2940 msg/sec), deep-copying the entire entries array each time (86% of processing time).

This PR applies two complementary optimizations that reduce main thread usage from **88.6% → 3.5%**:

### Changes

**Frontend — `streamJsonPatchEntries.ts`:**
- **Replace `structuredClone` with immer `produce()`** — Uses structural sharing to only clone modified branches instead of the entire snapshot. This pattern already exists in `useJsonPatchWsStream.ts`.
- **Batch WS messages per `requestAnimationFrame`** — Instead of processing each message individually (2940/sec), incoming JSON Patch ops are accumulated and flushed once per animation frame (~60/sec). Cross-message deduplication collapses redundant ops on the same path.

**Backend — Historical log deduplication (`container.rs`, `patch.rs`):**
- Add streaming dedup adapter that buffers consecutive patches targeting the same path, emitting only the final state per entry during historical log replay.
- Fix collect-then-emit approach for historical normalized logs to avoid premature stream termination.

### Profiling Results

| Metric | Before | After |
|--------|--------|-------|
| Clone/produce calls | 5853 in 2s | 34 in 0.3s |
| Total clone/produce time | 1523ms | 7ms |
| Main thread budget | **88.6%** | **3.5%** |

### Implementation Details

- `flush()` is called via `requestAnimationFrame` for normal streaming, but called **synchronously** when a `finished` message arrives to ensure no ops are dropped.
- Pending rAF is cancelled on WebSocket close and on controller `close()` to prevent stale callbacks.
- The initial snapshot clone (one-time setup) still uses `structuredClone` since it's not on the hot path.


Juanlucasbg:
## Rebase verification — PR #2588

- `crates/executors/src/logs/utils/patch.rs`
- `crates/services/src/services/container.rs`

Both files have been actively churned on `main` (log-stream restructuring, container lifecycle changes). The "perf: replace structuredClone with immer + rAF batching" change targets WS-stream hot-paths that may have been refactored.

**Action**: rebase, then verify the original `structuredClone` usage still exists in the current code. If the WS-stream pipeline was reshaped to use a different state mechanism (Zustand stores, etc.), this PR's premise may be partially obsolete.

Note: this is a `@stunningpixels` (maintainer) PR. Worth a maintainer self-review on whether the perf optimization still applies.

PR #2647, opened 2026-02-10
Handle 401 from /token endpoint by showing login dialog (Vibe Kanban)

anastasiya1155:
## Summary

When the `/api/auth/token` endpoint returns a 401 (unauthorized), the frontend now automatically transitions to the logged-out state and shows the OAuth login dialog, prompting the user to re-authenticate.

Previously, a 401 from the token endpoint would silently return `null` with no user-facing feedback — the app would remain in a broken authenticated state where API calls would keep failing.

## Changes

- Added a `handleUnauthorized()` method to `TokenManager` that:
  1. Invalidates the `user-system` React Query cache, causing a re-fetch of login status from the backend (which returns `loggedout`), updating `useAuth().isSignedIn` to `false`
  2. Shows the `OAuthDialog` so the user can immediately re-authenticate
- Called `handleUnauthorized()` in both `getToken()` (initial/cached token fetch) and `doRefresh()` (401-triggered refresh) when no valid token is returned
- Uses dynamic imports to avoid circular dependencies between the token manager and dialog/query modules

## Test plan

- [ ] Log in, then call `POST /api/auth/logout` via curl to clear server-side credentials
- [ ] Observe that the next token fetch triggers the login dialog automatically
- [ ] Verify that after re-authenticating via the dialog, the app resumes normal operation
- [ ] Verify that multiple concurrent 401s don't open multiple login dialogs (deduplication via `refreshPromise`)

### DIFF
--- a/frontend/src/lib/auth/tokenManager.ts
+++ b/frontend/src/lib/auth/tokenManager.ts
@@ -1,10 +1,14 @@
-import { oauthApi } from '../api';
+import { ApiError, oauthApi } from '../api';
 
 const TOKEN_QUERY_KEY = ['auth', 'token'] as const;
 const TOKEN_STALE_TIME = 125 * 1000; // 125 seconds (slightly longer than the BE stale time)
 
 type RefreshStateCallback = (isRefreshing: boolean) => void;
 type PauseableShape = { pause: () => void; resume: () => void };
+
+function isUnauthorizedError(error: unknown): boolean {
+  return error instanceof ApiError && error.statusCode === 401;
+}
 
 class TokenManager {
   private isRefreshing = false;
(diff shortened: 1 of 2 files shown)


frontend/src/lib/auth/tokenManager.ts
```diff
@@ -32,7 +32,11 @@ class TokenManager {
         queryFn: () => oauthApi.getToken(),
         staleTime: TOKEN_STALE_TIME,
       });
-      return data?.access_token ?? null;
+      const token = data?.access_token ?? null;
```
ggordonhall: This doesn't appear to explicitly check for a 401 status code from the endpoint. Does this mean we'll be displaying the log in screen when users are not actually logged out? E.g. a 500 from the server.

anastasiya1155: good catch, fixed

ggordonhall (approved):

PR #2650, opened 2026-02-10
fix(frontend): improve attempts list visibility & add back navigation

harun-yardimci:
## Changes
1. **Fix Attempts List Visibility:** Removed manual height calculation in TaskPanel to allow the attempts list to be visible and scroll properly.
2. **Add Back Navigation:** Added a 'Back to projects' button in the Workspaces sidebar. When filtered by a project, it navigates back to that project's board; otherwise, to the main list.

### DIFF
--- a/frontend/src/components/panels/TaskPanel.tsx
+++ b/frontend/src/components/panels/TaskPanel.tsx
@@ -102,7 +102,7 @@ const TaskPanel = ({ task }: TaskPanelProps) => {
   return (
     <>
       <NewCardContent>
-        <div className="p-6 flex flex-col h-full max-h-[calc(100vh-8rem)]">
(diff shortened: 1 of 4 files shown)


Juanlucasbg:
## Rebase verification — PR #2650

- `frontend/src/components/panels/TaskPanel.tsx`
- `frontend/src/components/ui/new-card.tsx`

Pre-monorepo paths. The "improve attempts list visibility & add back navigation" change targets components that have moved to `packages/ui/` and `packages/web-core/`.

**Action**: this PR predates the monorepo restructure. Recommend re-doing as a fresh PR against the current `packages/...` tree. Given the sunset context, also verify whether the back-navigation UX still applies after #3387's project-routes-to-export-only change.

— Automated rebase-verification: detected pre-monorepo paths; needs full relocation.

PR #2653, opened 2026-02-10
feat(ui): add "Go Back to Kanban" button

kldzj:
As the Workspace view isn't fully ready yet and doesn't include any kind of Kanban view, in the meantime a button for going back to the associated project view could be added.

If this PR is considered at all, I'd probably need some guidance on how to continue, as this implementation currently only redirects to the local projects route, but I guess we should check if the project is remote or local?


kldzj:
Irrelevant due to Cloud enforcement.

PR #2677, opened 2026-02-10
fix: Refresh token fixes

abcpro1:
- Make timeout errors non-retryable to avoid reusing the same refresh token twice, which is not permitted by the remote backend.
- Reimplement frontend Access token refresh properly by parsing the JWT token and preemptively refreshing the token before expiry.
- Reimplement remote-frontend token management similarly with retry logic, timeout protection, and cross-tab locking to prevent multiple from refreshing the same token twice.

### DIFF
--- a/remote-frontend/src/tokenManager.ts
+++ b/remote-frontend/src/tokenManager.ts
@@ -8,7 +8,7 @@ import {
 import { shouldRefreshAccessToken } from "shared/jwt";
 import { refreshTokens } from "./api";
 
-const TOKEN_REFRESH_TIMEOUT_MS = 300_000;
+const TOKEN_REFRESH_TIMEOUT_MS = 80_000;
 const TOKEN_REFRESH_MAX_ATTEMPTS = 3;
 
 async function refreshWithRetry(refreshToken: string) {


frontend/src/lib/auth/tokenManager.ts
```diff
@@ -40,6 +37,14 @@ class TokenManager {
       return null;
     }
 
+    const cachedData = queryClient.getQueryData<{
+      access_token?: string;
+    }>(TOKEN_QUERY_KEY);
+    const cachedToken = cachedData?.access_token;
+    if (cachedToken === undefined || shouldRefreshAccessToken(cachedToken)) {
```
anastasiya1155: Maybe just use !cachedToken in case it is null or empty

remote-frontend/src/tokenManager.ts
```diff
@@ -0,0 +1,106 @@
...
+  getAccessToken,
+  getRefreshToken,
+  storeTokens,
+  clearAccessToken,
+  clearTokens,
+} from "./auth";
+import { shouldRefreshAccessToken } from "shared/jwt";
+import { refreshTokens } from "./api";
+
+const TOKEN_REFRESH_TIMEOUT_MS = 300_000;
```
anastasiya1155: are we sure about 5 minutes timeout here? It seems too long for UX, can we reduce it to say 1 - 2 minutes? BE timeout an stay the same if needed

abcpro1: After the timeout the user will be forcefully logged out. Do you think 1 - 2 minutes would cover occasional slowdowns, either server-side or user's slow connection side?

abcpro1: done

PR #2682, opened 2026-02-10
Fix: handle multi-byte UTF-8 characters split across chunk boundaries

### LINKED ISSUE #1795: vibe-kanban v0.0.134: Garbled Text Error When Interacting with Large Model Proxy via WebSocket
The version of vibe-kanban in use is 0.0.134. When vibe-kanban interacts with the large model proxy, checking via the browser console shows that the requested interface is:
ws://127.0.0.1:11168/api/execution-processes/198d80db-a6ae-49c0-8674-6fc61df91e68/normalized-logs/ws

However, the returned message is:

{"JsonPatch":[{"op":"add","path":"/entries/0","value":{"type":"NORMALIZED_ENTRY","content":{"timestamp":null,"entry_type":{"type":"error_message", "error_type":{"type":"other"}},"content":"Garbled text here","metadata":null}}}]}

What could be the cause of this issue? The large model proxy works normally when accessed directly locally.

hansonkim:
## Summary
- Adds `Utf8Chunker` struct that buffers incomplete UTF-8 trailing bytes across chunk boundaries, preventing garbled CJK (Korean/Chinese/Japanese) characters during streaming output
- Replaces all `String::from_utf8_lossy` calls in stdout/stderr stream processing with the new chunker for correct multi-byte handling
- Adds `utf8_safe_stream()` helper for converting byte-oriented `ReaderStream` into UTF-8 safe `String` streams

Fixes #1795

## Test plan
- [x] Verify Korean/Chinese/Japanese text renders correctly during Claude streaming output
- [x] Verify stdout/stderr streams work correctly with ASCII-only content
- [x] Run `cargo test --workspace` to ensure no regressions

### DIFF
--- a/crates/executors/src/stdout_dup.rs
+++ b/crates/executors/src/stdout_dup.rs
@@ -10,11 +10,118 @@ use std::os::windows::io::{FromRawHandle, IntoRawHandle, OwnedHandle};
 
 use command_group::{AsyncCommandGroup, AsyncGroupChild};
 use futures::{StreamExt, stream::BoxStream};
-use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::io::ReaderStream;
 
 use crate::executors::{ExecutorError, SpawnedChild};
+
+// =========================================
+// UTF-8 chunk boundary helper
+// =========================================
+
+/// Buffers incomplete UTF-8 trailing bytes across chunk boundaries.
+///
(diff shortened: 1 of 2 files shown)


Juanlucasbg:
## Rebase verification — PR #2682

- `crates/executors/src/stdout_dup.rs`

`stdout_dup.rs` (executor stdout capture / duplication) has been actively touched on `main`. The "handle multi-byte UTF-8 split across chunk boundaries" fix targets the chunk-decoding path, which may have been refactored.

**Action**: rebase, verify the chunk-decoder still has the same UTF-8-boundary bug. If the decoder was switched to a streaming-safe variant (`encoding_rs::Decoder` or similar), the fix may already be implicit. Otherwise re-apply the boundary-aware decoding logic.

PR #2688, opened 2026-02-11
Show PR button instantly using cached summary data (Vibe Kanban)

anastasiya1155:
## Summary

When clicking on a workspace that already has a PR attached, the "Open Pull Request" button took several seconds to appear. This happened because the Git panel waited for the expensive `/branch-status` endpoint (which makes git system calls for each repo) before showing any PR information. By the time it loaded, the user may have already navigated to a different workspace.

This PR fixes the delay by using the already-cached workspace summary data as an immediate fallback while the full branch status loads in the background.

### Backend
- **`crates/db/src/models/merge.rs`**: Added a `PrSummary` struct and updated `get_latest_pr_status_for_workspaces` to return `pr_number` and `pr_url` alongside `pr_status` (all from a single cheap DB query, no git calls).
- **`crates/server/src/routes/task_attempts/workspace_summary.rs`**: Extended `WorkspaceSummary` with `pr_number: Option<i64>` and `pr_url: Option<String>`, populated from the new `PrSummary`.

### Frontend
- **`frontend/src/components/ui-new/hooks/useWorkspaces.ts`**: Added `prNumber` and `prUrl` to `SidebarWorkspace`, mapped from the summary response.
- **`frontend/src/components/ui-new/containers/GitPanelContainer.tsx`**: Looks up the selected workspace's PR info from the already-cached summary data and uses it as a fallback when `branchStatus` hasn't loaded yet. Once `branchStatus` arrives, it takes precedence as before.

### Generated
- **`shared/types.ts`**: Regenerated to include the new `pr_number` and `pr_url` fields on `WorkspaceSummary`.

## How it works

The workspace summary endpoint (`POST /api/task-attempts/summary`) was already fetched every 15 seconds and cached on the frontend. It already included `pr_status` but not the PR number or URL. This PR extends it to also return those fields from the database (no git operations needed), then the Git panel uses this data to render the PR button immediately while the full `branch-status` response loads in the background.

---

### DIFF
--- a/crates/remote/Cargo.lock
+++ b/crates/remote/Cargo.lock
@@ -2322,9 +2322,9 @@ dependencies = [
 
 [[package]]
 name = "native-tls"
-version = "0.2.14"
+version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87de3442987e9dbec73158d5c715e7ad9072fda936bb03d19d7fa10e00520f0e"
+checksum = "6cdede44f9a69cab2899a2049e2c3bd49bf911a157f6a3353d4a91c61abbce44"
 dependencies = [
  "libc",
  "log",


crates/db/src/models/merge.rs
```diff
@@ -316,24 +324,29 @@ impl Merge {
...
-    /// Get the latest PR status for each workspace (for workspace summaries)
-    /// Returns a map of workspace_id -> MergeStatus for workspaces that have PRs
+    /// Summary of the latest PR for a workspace
+    /// Get the latest PR info for each workspace (for workspace summaries)
+    /// Returns a map of workspace_id -> PrSummary for workspaces that have PRs
     pub async fn get_latest_pr_status_for_workspaces(
         pool: &SqlitePool,
         archived: bool,
-    ) -> Result<HashMap<Uuid, MergeStatus>, sqlx::Error> {
+    ) -> Result<HashMap<Uuid, PrSummary>, sqlx::Error> {
```
LSRCT: could this return a `HashMap<Uuid, PrMerge>` and avoid the new `PrSummary` struct?

crates/db/src/models/merge.rs
```diff
@@ -316,24 +324,29 @@ impl Merge {
...
         struct PrStatusRow {
             workspace_id: Uuid,
             pr_status: Option<MergeStatus>,
+            pr_number: Option<i64>,
+            pr_url: Option<String>,
         }
 
         // Get the latest PR for each workspace by using a subquery to find the max created_at
         // Only consider PR merges (not direct merges)
         let rows = sqlx::query_as::<_, PrStatusRow>(
```
LSRCT: This was here before your PR, but ideally we'd use a macro `sqlx::query_as!` here. The non macro version is not typechecked at compile time, we avoid it. Maybe easy to change it here given `Merge` implements `From<MergeRow>`.

LSRCT (approved):

PR #2711, opened 2026-02-12
Add no-SSH Docker build path for remote-server

hazem3500:
## Summary

Makes the remote-server Docker image buildable without an SSH agent so it works on Dokploy and other environments where `SSH_AUTH_SOCK` is not set.

### Dockerfile (`crates/remote/Dockerfile`)

- **`builder-no-ssh`** — New stage with the same setup as `builder` (Rust base, build deps, WORKDIR, COPY of crates/shared/assets, and the RUN that strips the private billing dependency from `Cargo.toml` when `FEATURES` is empty). The final RUN runs the same `cargo build` and `cp`, but **without** `--mount=type=ssh`; only the cache mounts (registry, git, target) are used.
- **`runtime-no-ssh`** — New runtime stage with the same structure as `runtime` (Debian base, ca-certificates, wget, git, appuser, HEALTHCHECK, ENTRYPOINT), but it copies the binary from `builder-no-ssh` and static files from `fe-builder`.

The existing `builder` and `runtime` stages are unchanged so builds that need the private billing crate can still use them with SSH.

### Docker Compose (`crates/remote/docker-compose.yml`)

- **`target: ${REMOTE_SERVER_BUILD_TARGET:-runtime-no-ssh}`** — Default build target is now `runtime-no-ssh`, so `docker compose build` does not require an SSH agent.
- **Removed `ssh: - default`** — The default build no longer forwards the host SSH agent into the build.

## Why

The original Dockerfile used `--mount=type=ssh` so Cargo could clone the private billing crate during the build. That requires an SSH agent (e.g. `ssh-agent` with keys, or Docker BuildKit SSH forwarding). On platforms like Dokploy or in CI where no SSH agent is configured, the build fails. By adding a parallel no-SSH path and making it the default, the image builds out of the box in those environments. When `FEATURES` is empty, the billing dep is already stripped in both builder stages, so the no-SSH builder never needs to access private git.

## Usage

- **Default (no SSH):** `docker compose build` builds the `runtime-no-ssh` image. Use this for self-hosted / Dokploy / environments without SSH.
- **With billing (SSH):** Set `REMOTE_SERVER_BUILD_TARGET=runtime` and build with SSH (e.g. `docker compose build --ssh default`) when you need the private billing dependency.

### DIFF
--- a/crates/remote/Dockerfile
+++ b/crates/remote/Dockerfile
@@ -61,6 +61,43 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     cargo build ${FEATURES:+--locked} --release --manifest-path crates/remote/Cargo.toml ${FEATURES:+--features ${FEATURES}} \
  && cp crates/remote/target/release/${APP_NAME} /app/bin/${APP_NAME}
 
+FROM rust:1.93-slim-bookworm AS builder-no-ssh
+ARG APP_NAME
+ARG FEATURES
+ARG POSTHOG_API_KEY=""
+ARG POSTHOG_API_ENDPOINT=""
+
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+ENV POSTHOG_API_KEY=${POSTHOG_API_KEY}
+ENV POSTHOG_API_ENDPOINT=${POSTHOG_API_ENDPOINT}
+
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends pkg-config libssl-dev ca-certificates git openssh-client \
+  && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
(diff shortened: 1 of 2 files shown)


Juanlucasbg:
## Rebase verification — PR #2711

- `crates/remote/docker-compose.yml`

`docker-compose.yml` has been touched on `main` since this PR was opened (multiple env-var additions: `ALLOWED_USERS`, `GITHUB_BASE_URL`, `SMTP_*`, etc.). The "no-SSH Docker build path" change probably overlaps with one of those service definitions.

**Action**: rebase the `docker-compose.yml` change manually — accept `main`'s version as base, then re-add the no-SSH service alongside the current set. The 67-LOC PR also touches the Dockerfile presumably; verify those don't conflict separately.

PR #2715, opened 2026-02-12
Skip unnecessary CI checks on PRs using path-based filtering (Vibe Kanban)

anastasiya1155:
## Summary

Adds path-based filtering to the CI `Test` workflow so that PRs only run checks relevant to the files that changed. This avoids spinning up expensive 16-vCPU runners for jobs that have no chance of being affected.

## What changed

A new lightweight `changes` job (runs on `ubuntu-latest`) uses [`step-security/paths-filter@v3`](https://github.com/step-security/paths-filter) to detect which areas of the codebase a PR touches. Each downstream job is gated on the relevant filter output:

| Filter | Paths | Gates jobs |
|--------|-------|------------|
| **frontend** | `frontend/**`, `shared/**`, `scripts/check-i18n.sh`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `package.json`, `.npmrc` | `frontend-checks` |
| **backend** | `crates/{api-types,db,deployment,executors,git,local-deployment,review,server,services,utils}/**`, `Cargo.toml`, `Cargo.lock`, `shared/**`, `scripts/prepare-db.js`, `rustfmt.toml`, `rust-toolchain.toml`, `.cargo/**`, `pnpm-lock.yaml`, `package.json` | `backend-schema-checks`, `backend-clippy`, `backend-test` |
| **backend-remote** | `crates/remote/**`, `crates/api-types/**`, `crates/utils/**`, `Cargo.toml`, `Cargo.lock`, `shared/**`, `rust-toolchain.toml`, `.cargo/**`, `pnpm-lock.yaml`, `package.json` | `backend-remote-checks` |

## Why

- **Frontend-only PRs** (CSS, component, i18n changes) no longer wait for 4 Rust compilation jobs that can't possibly fail.
- **Backend-only PRs** skip frontend lint/build.
- Reduces CI cost and queue time for the common case where changes are scoped to one area.

## Implementation details

- On `push` to `main` and `workflow_dispatch`, the `changes` job is skipped entirely (it has `if: github.event_name == 'pull_request'`). Downstream jobs use `always() && (... || needs.changes.result == 'skipped')` to ensure they **always run** on non-PR events — preserving the existing safety net.
- The existing fork-check condition on `backend-remote-checks` is preserved alongside the new path filter.
- Shared paths like `pnpm-lock.yaml`, `package.json`, and `Cargo.toml` trigger both frontend and backend jobs, since dependency changes can affect either side.
- Uses `step-security/paths-filter@v3` (maintained fork of `dorny/paths-filter`) since the original is unmaintained.
- Backend filter explicitly lists workspace crates rather than using `crates/**` with negation, because `dorny/paths-filter`'s negation patterns (`!crates/remote/**`) match all files outside that path under the default `predicate-quantifier: 'some'`.
- Backend-remote filter only includes `crates/remote/**` and its path dependencies (`api-types`, `utils`) rather than all crates.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
       backend-remote: ${{ steps.filter.outputs['backend-remote'] }}
     steps:
       - uses: actions/checkout@v6
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@0bc4621a3135347011ad047f9ecf449bf72ce2bd # v3.0.0
         id: filter
         with:
           filters: |


.github/workflows/test.yml
```diff
@@ -25,7 +24,55 @@ env:
...
+    if: github.event_name == 'pull_request'
+    permissions:
+      pull-requests: read
+    outputs:
+      frontend: ${{ steps.filter.outputs.frontend }}
+      backend: ${{ steps.filter.outputs.backend }}
+      backend-remote: ${{ steps.filter.outputs.backend-remote }}
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dorny/paths-filter@v3
```
ggordonhall: This action is unmaintained. Can you find another one?

anastasiya1155: you are right, it seems abandoned, but there is a good fork that is maintaned and is backed by a reputable security company

anastasiya1155: ah, actually it is a paid action, only dorny/paths-filter@v3 seems a widely used standard action for this, alternatively we can write our own script but not sure it is a better approach

ggordonhall: Looks like Sentry are actually using it; https://github.com/getsentry/sentry/blob/d92ba5b00599c3e5b2cc11c7730ad471b2a3cb8a/.github/workflows/backend.yml#L38. Let's go for it. But let's pin the commit like Sentry.

anastasiya1155: what if it gets updated, don't we want to use latest, e.g. with updated node version?

ggordonhall: there will likely be a major version upgrade for that. Always wary of these types of repos being hijacked, especially when they are unmaintained, so safer to pin.

PR #2752, opened 2026-02-15
perf: optimize workspace loading performance

kavorite:
## Summary

- Add server-side normalized log cache (IndexMap, 200 entries, Arc-wrapped values) that populates as messages stream through the normalizer, avoiding repeat DB reads and re-normalization
- Add frontend module-level entry cache (Map, 100 entries) that persists across component mount/unmount, avoiding repeat WebSocket round-trips
- Load execution processes in parallel (batches of 6) instead of sequentially in both `useConversationHistory` hooks
- Respect `HOST` env var in Vite dev server config

## Test plan

- [x] Verified single WebSocket connection returns correct JsonPatch count (302 patches)
- [x] Verified 6 concurrent WebSocket connections all return correct data
- [x] Measured ~10x speedup for parallel vs sequential loading on cache miss
- [x] Verified cache hits return identical data to cache misses
- [x] Manually tested in dev build: both user and agent messages render correctly


crates/local-deployment/src/container.rs
```diff
@@ -83,6 +84,7 @@ pub struct LocalContainerService {
     analytics: Option<AnalyticsContext>,
     approvals: Approvals,
     queued_message_service: QueuedMessageService,
+    normalized_log_cache: Arc<RwLock<IndexMap<Uuid, Arc<Vec<LogMsg>>>>>,
```
abcpro1: `lru` would be a better pick. However, task switching is not frequent enough to justify the permanent over-time memory usage increase that this will add. Caching to disk would be more appropriate.

kavorite: Reworked this; replaced in-memory cache with a disk-backed approach:

- `NormalizedLogCache` writes pre-serialized JSON lines to `{cache_dir}/{execution_process_id}.jsonl` files
- LRU eviction is mtime-based with a 128 MiB size budget — no persistent memory overhead
- Cache hits serve the pre-serialized lines directly over WebSocket, bypassing DB + serde entirely
- Only serves cached data for completed processes to avoid partial results
- Eviction runs in `spawn_blocking` to stay off the async runtime

The other optimizations from before are still in (direct `Patch` construction, parallel WS loads, batched patch application).

Force-pushed

Juanlucasbg:
## Rebase verification — PR #2752

- `crates/executors/src/logs/utils/patch.rs`

Single-file conflict on the log-patch utility. Same file as #2588 (perf: replace structuredClone with immer) — both PRs target the WS-stream patch hot path. Likely overlapping concerns.

**Action**: rebase, then check if #2588's already-suggested approach overlaps with this PR's "optimize workspace loading performance" change. If they target the same allocation hot spot, one supersedes the other; if different, both can land in sequence.

— Automated rebase-verification: confirmed single-file conflict; possible overlap with #2588.

PR #2755, opened 2026-02-16
Ask whether to unlink from issue when deleting a workspace (Vibe Kanban)

anastasiya1155:
## Summary

When deleting a workspace that is linked to a remote issue, the link was silently left behind — either always removed without asking (issue page) or not removed at all (workspaces page). This PR adds an explicit "Also unlink from issue" checkbox to the delete workspace dialog so users can choose.

## Changes

- **DeleteWorkspaceDialog**: Added an optional `isLinkedToIssue` prop that, when true, renders a new "Also unlink from issue" checkbox (defaulted to checked). The dialog result now includes `unlinkFromIssue` so callers can act on the user's choice.
- **DeleteWorkspace action** (`actions/index.ts`): Looks up `ctx.remoteWorkspaces` to detect if the workspace is linked to an issue, passes `isLinkedToIssue` to the dialog, and calls `unlinkFromIssue` only if the user opted in.
- **IssueWorkspacesSectionContainer**: Now passes `isLinkedToIssue: true` to the dialog and conditionally unlinks based on the dialog result, instead of always unconditionally unlinking.
- **Ordering fix**: The unlink call happens *after* successful deletion, not before. This prevents the workspace from being unlinked but not deleted when deletion fails (e.g. processes still running).
- **Translation**: Added `unlinkFromIssueLabel` key to `en/common.json`.

## Why

Deleting a workspace from the workspaces page left an orphaned remote workspace link on the issue. The issue page always silently unlinked, giving no user control. Now both paths go through the same dialog with an explicit opt-in/out checkbox.

### DIFF
--- a/frontend/src/components/ui-new/dialogs/DeleteWorkspaceDialog.tsx
+++ b/frontend/src/components/ui-new/dialogs/DeleteWorkspaceDialog.tsx
@@ -44,9 +44,13 @@ const DeleteWorkspaceDialogImpl = NiceModal.create<DeleteWorkspaceDialogProps>(
 
     // Fetch issue data via Electric sync to show issue simple_id
     const isLinkedToIssue = !!linkedIssueId;
-    const { data: issues } = useShape(PROJECT_ISSUES_SHAPE, {
-      project_id: linkedProjectId ?? '',
-    }, { enabled: !!linkedProjectId });
+    const { data: issues } = useShape(
+      PROJECT_ISSUES_SHAPE,
+      {
+        project_id: linkedProjectId ?? '',
+      },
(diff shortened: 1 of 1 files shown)


stunningpixels:
Screenshot 2026-02-19 at 16 22 22

consistency - we should show name of issue or issue short id, like we do with branch

stunningpixels:
Might be worth tackling VIB-189 in the same PR

anastasiya1155:
> consistency - we should show name of issue or issue short id, like we do with branch

> Might be worth tackling VIB-189 in the same PR

both are done

stunningpixels (approved):

Issue #2751, opened 2026-02-15
draft:feat: use full remote issue UUID in merge commit messages


CPUtester5465:
## Problem

Merge commit messages generated by VK use a **truncated local task UUID** (8 hex chars) instead of the full remote issue UUID.

**Current behavior** (`task_attempts.rs:546-548`):
```rust
let task_uuid_str = task.id.to_string();
let first_uuid_section = task_uuid_str.split('-').next().unwrap_or(&task_uuid_str);
let mut commit_message = format!("{} (vibe-kanban {})", task.title, first_uuid_section);
```

This produces:
```
My task title (vibe-kanban 7fd20218)
```

### Two issues:

1. **UUID truncation**: Only the first 8 hex characters are used — the ID cannot be resolved back to its source
2. **Wrong ID type**: `task.id` is a **local SQLite task UUID**, not the **remote issue UUID**. These are completely different UUIDs — the local task ID has zero correlation to the remote issue ID returned by `get_context` or MCP tools

### Impact

- Agents (Claude Code, etc.) see commit messages with `(vibe-kanban 7fd20218)` and copy these 8-char IDs into cross-references, issue descriptions, and xref URIs
- These short IDs are **not resolvable** via `get_issue` or any MCP tool
- After PR #2678 migrated MCP to issue-first workflows, the commit message format was not updated to match

## Proposed Fix

Replace lines 541-548 in `task_attempts.rs` with logic that:
1. Looks up the remote workspace to get the `issue_id` (same mechanism as `get_context`)
2. Uses the **full remote issue UUID** in the commit message
3. Falls back to full local task UUID if no remote issue is linked

Expected result:
```
My task title (vibe-kanban f1403a5c-b868-4902-af15-7771af2cbe98)
```

### Bonus: consider using `simple_id`

PR #2244 added human-readable `simple_id` (e.g., `MCC-42`) to issues. If the simple_id is available, the commit could use:
```
My task title (vibe-kanban MCC-42)
```
This would be even more user-friendly while remaining resolvable.

## Also affected

`frontend/src/components/dialogs/tasks/CreatePRDialog.tsx:84` sets PR titles with no UUID at all:
```tsx
setPrTitle(`${task.title} (vibe-kanban)`);
```

This should include the issue identifier too.

## Related PRs

- #2244 — Added `simple_id` system (merged)
- #2734 — Exposes `simple_id` via MCP (open)
[...]

Issue #2732, opened 2026-02-13
Make suqshing commits on merge optional


hugochinchilla:
Sometimes I need to work on long refactors that are best worked on a single workflow for keeping context.
The commits are usually small self-contained units of work, merging with a full history provides big value to me as It makes easy to trace what has been done.

I would like to have an option to disable the current squash behavior.

Korayem:
I stopped merging from inside VK workspace. I do it manually to avoid squashing.
Would be great to have this in Global settings

PR #2786, opened 2026-02-17
feat: optional commit branding

kldzj:
This pull request adds a new user-configurable setting that controls whether Vibe Kanban branding is included in commit messages and pull request (PR) titles and descriptions. By default, branding is enabled, but users can now toggle this in the general settings. The change is implemented across both backend and frontend, with appropriate updates to configuration handling, UI, and localization.

### DIFF
--- a/crates/server/src/routes/task_attempts/pr.rs
+++ b/crates/server/src/routes/task_attempts/pr.rs
@@ -105,16 +105,15 @@ async fn trigger_pr_description_follow_up(
     // Get the custom prompt from config, or use default based on branding setting
     let config = deployment.config().read().await;
     let branding_enabled = config.branding_in_commit_and_pr_enabled;
-    let prompt_template = config
-        .pr_auto_description_prompt
-        .as_deref()
-        .unwrap_or_else(|| {
-            if branding_enabled {
+    let prompt_template =
+        config
+            .pr_auto_description_prompt
+            .as_deref()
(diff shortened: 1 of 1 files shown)


LSRCT:
Hi, did you check out the `Use custom prompt` setting in Settings > General? We have it for both PRs and commits.
Did you experience any limitation with that?

kldzj:
The custom prompt works flawlessly, but this wouldn't address the [merge commit message suffix](https://github.com/BloopAI/vibe-kanban/pull/2786/changes#diff-0719366382b3463d5cc69ada3b60f0e601f58b6b541e9252b4f58559def39d59R520-R524). I think this should be up to the user, whether they want the branding or not.

What's your take on this? Are you open to omit it in general?

LSRCT:
Gotcha. The branding for the merge commit message has a vk specific id and given this only affects merge commits, I'd be inclined to leave it in for now
- it enables agents to relate commits to vk issues (see #2751)
- leaving it out keeps the config simple, the current commit reminder settings sound similar but have nothing to do with merge commits

I'm open to revisiting this if we decide to allow more customisation for merge commits in general (like not squashing as requested in #2732)

Given the PR is already customisable I'd close this for now. Hope this makes sense!

kldzj:
@LSRCT I could tackle #2732 if you want and also add the option to omit the branding, if this is okay for you?

PR #2793, opened 2026-02-17
feat: user whitelisting for OAuth authentication

kldzj:
Adds email-based user whitelisting to restrict OAuth login to specific users when self-hosting the Vibe Kanban remote server.

**Changes:**
- New `ALLOWED_USERS` environment variable (comma-separated emails)
- Whitelist checked during OAuth callback before user creation
- Errors surfaced in login dialog via `postMessage`
- Updated documentation in README and Docker deployment guide

**Usage:**
```
ALLOWED_USERS=admin@company.com,user@company.com
```

Empty/unset = allow all users (backward compatible).

---

> [!NOTE]
> **Medium Risk**
> Adds a configurable gate in the OAuth callback flow that can deny authentication based on email, which affects sign-in behavior and could unintentionally lock out users if misconfigured.
> 
> **Overview**
> Adds optional **email-based user whitelisting** for self-hosted remote OAuth authentication via a new `ALLOWED_USERS` env var (comma-separated, normalized to lowercase).
> 
> When `ALLOWED_USERS` is set, the server checks the resolved OAuth email during callback handling and returns an `access_denied` error before creating/updating the user/session if the email is not on the list. Documentation and Docker deployment examples are updated to include the new setting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13898e4006d53d5b5bd574e24e33912bbb9e733f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Juanlucasbg:
## Aggressive review summary — PR #2793

46-line OAuth-whitelist feature: parses `ALLOWED_USERS` (comma-separated) at config load, then short-circuits the OAuth callback to `access_denied` if the authenticated user's email isn't on the list. Verdict: **clean — recommend merge with the two MED items below addressed**.

### Security review (full pass — this PR is auth-sensitive)

**Stored list normalization (PASS)** — `crates/remote/src/config.rs:258-272` lowercases and trims each entry, drops empties, and collapses an empty resulting list to `None` so an `ALLOWED_USERS=` (no value) in `.env` correctly means "unrestricted" rather than "deny everyone". Defensive defaults are right.

**Comparison (`crates/remote/src/auth/handoff.rs:288-298`) (MOSTLY PASS, 1 MED)** — `email.to_lowercase()` matches the stored normalization. Constant-time isn't required here (whitelist membership isn't secret).

> **MED — Trim incoming email before comparison.** Stored entries are trimmed; incoming `email = ensure_email(...)` is not. If an OAuth provider returns an email with stray whitespace (rare but not impossible — proxy / encoding bugs), the check fails open as deny, locking out a legitimate user. One-liner fix: `email.trim().to_lowercase()`.

**Email source (`ensure_email`, lines 537-547) (MED — verification gap)** —

```rust
fn ensure_email(provider: &str, profile: &ProviderUser) -> String {
    if let Some(email) = profile.email.clone() { return email; }
    match provider {
        "github" => format!("{}@users.noreply.github.com", profile.id),
        "google" => format!("{}@users.noreply.google.com", profile.id),
        _ => format!("{}@oauth.local", profile.id),
    }
}
```

The whitelist check uses whatever the provider returns in `profile.email`. GitHub's primary email is verified; Google's is always verified. The `_ =>` synthetic-email branch produces `<id>@oauth.local`, which won't match any human-readable whitelist entry — safe by construction.

> **MED — No explicit `email_verified` check.** If a future provider is added (or the GitHub/Google client is reconfigured to return secondary/unverified emails), an attacker could register `admin@yourcompany.com` they don't own and pass the whitelist. Mitigation: make `ensure_email` return `Option<String>` (or a `(String, bool)` pair) and reject the OAuth flow when the provider's email-verification claim is `false` or absent — fail closed.

**Audit/logging (LOW)** — `tracing::info!(count = users.len(), "User whitelist enabled")` logs on startup. Denied attempts are NOT logged. For a security control this is the wrong direction:

> **LOW — Log denied attempts.** Add `tracing::warn!(email = %email, "Whitelist denied OAuth login")` before returning `access_denied`. Operators of a self-hosted deployment will want this for brute-force / unauthorized-attempt detection. Non-blocking.

**Error opacity (PASS)** — Returns `error: "access_denied"` without leaking whether the email matched a known account. Good.

**Configuration (PASS)** — README + Docker docs both updated with the new env var. Default is unrestricted, which is the right behavior for OSS quickstart.

### Other findings

- **Structural**: 0 HIGH. Plumbing through `Server::new → OAuthHandoffService::new` is straightforward; the `Option<Vec<String>>` shape correctly distinguishes "unset" from "empty list".
- **Adversarial**: 0 HIGH. All of: empty string, whitespace-only, comma-only, leading/trailing commas, `Some("")` from env — all collapse to `None` (unrestricted) per the `filter(|v| !v.is_empty())` and `filter(|list| !list.is_empty())` guards.
- **Conventions**: PASS. Matches the `Option<T>` config pattern used for `github_app`, `review_worker_base_url`, etc.

### NITs

- Whitelist is loaded once at startup. Document this — operators may expect runtime updates.
- Docs example uses `admin@example.com,user@example.com`. Mention case-insensitivity explicitly so the operator knows `Admin@Example.com` is treated identically.
- No tests added. `CLAUDE.md` says "Add tests for new logic and edge cases." A `#[test]` for the env-var parsing edge cases (empty / whitespace / mixed-case / list-with-empty-entries) would lock the contract.

### Verdict

Approve once trim-incoming and `email_verified` (or equivalent) are added. Both are MED, both are one-liners. Without them, the feature is still better than the unrestricted-only status quo, but the verification gap leaves a real vector for any future provider that returns unverified emails.

— Reviewed by automated single-pass review (auth-sensitive — full security pass; structural/adversarial/conventions also covered).

PR #2824, opened 2026-02-18
Add Codex gpt-5.3-codex-spark to the model list

abcpro1:


LSRCT:
Does not work for me.
Tried API key and ChatGPT Plus.

Issue #2377, opened 2026-01-28
can't stop task almost time


TheonChin:
2026-01-28T15:46:02.046725Z ERROR local_deployment::container: Failed to cleanly kill running execution process ExecutionProcess { id: 91e54256-248d-4cad-895f-780579e3b762, session_id: 18d95898-4424-438d-b365-f3fb134f919d, run_reason: CodingAgent, executor_action: Json(ExecutorAction(ExecutorAction { typ: CodingAgentInitialRequest(CodingAgentInitialRequest { prompt: "产品价值讨论\n\n读取仓库的信息，和我一起讨论产品的价值。", executor_profile_id: ExecutorProfileId { executor: Codex, variant: None }, working_dir: Some("ai-awasome-cowork") }), next_action: None })), status: Running, exit_code: None, dropped: false, started_at: 2026-01-28T15:42:38.992363363Z, completed_at: None, created_at: 2026-01-28T15:42:38.992363363Z, updated_at: 2026-01-28T15:42:38.992363363Z }: Other(Child process not found for execution)

dyk98:
same, task is done, but always show loading status

<img width="592" height="254" alt="Image" src="https://github.com/user-attachments/assets/1f8787cd-70a4-4c6c-97f4-fa47f83fa580" />

KevinX41030:
same

simonxmau:
same

Issue #2302, opened 2026-01-25
It has been continuously displaying this status without performing any tasks.


zhilianbi520:
After creating a new task, this status keeps being displayed without any task being executed.
<img width="1064" height="1220" alt="Image" src="https://github.com/user-attachments/assets/97915239-f957-450a-bf95-b49f7932bf3e" />

node:v22.13.0
os:mac
npm:10.9.2
tools:claude code

Issue #2042, opened 2026-01-14
Stuck at  "Started attempt for task"


lironsam:
I try to set up a new project and try to put some tasks but it is stucked. Nothing happen.

On the terminal I have
2026-01-14T13:47:12.566961Z  INFO server::routes::tasks: Started attempt for task a9c8ceb4-5883-49b6-a68a-d699f4957b8c

<img width="1238" height="595" alt="Image" src="https://github.com/user-attachments/assets/15d897e6-5303-4122-9cea-d314ba8825f0" />

I've installed it on Macbook Pro and get the last version. 

Any insights ?

LSRCT:
That's very strange, do you get anything after waiting a bit?
What happens if you run `npx -y @anthropic-ai/claude-code@2.1.2` in the terminal?

Novtopro:
Got the same issue here. I can run npx -y @anthropic-ai/claude-code@2.1.2, chat with the model without any issue, it just stucks with

```
2026-01-14T15:07:29.117046Z  INFO local_deployment::container: Created workspace CLAUDE.md with 1 import(s)
2026-01-14T15:07:29.135917Z  INFO server::routes::task_attempts: Created attempt for task 9a7ad12a-b46c-435f-996b-0d6e61ac6369
```

Opencode is fine, by the way.

lironsam:
> npx -y @anthropic-ai/claude-code@2.1.2

It runs claude code

Novtopro:
@lironsam @LSRCT 

In my case, I solved it by resetting the shell using SHELL=/bin/bash

```
SHELL=/bin/bash RUST_LOG=debug PORT=9999 HOST=0.0.0.0 npx vibe-kanban
```

<img width="1720" height="1134" alt="Image" src="https://github.com/user-attachments/assets/33ab58f3-7985-404c-a194-1998ce14331a" />

<img width="1554" height="952" alt="Image" src="https://github.com/user-attachments/assets/cbba5146-ebdb-4b21-bd59-66af35933a00" />

Issue #1938, opened 2026-01-11
Database Deadlocks and Codex Session Abnormalities


Abel-j:
Some Codex sessions fail to continue after ending and encounter errors. This issue is quite unpredictable and occurs intermittently. Could it be related to running it in WSL2? Disk performance in WSL2 is poor, and I often see database deadlock issues and slow query warnings. Can support be added for configuring external databases such as MySQL?

Codex fail: 
```
I/O error: failed to decode resumeConversation response: unknown variant `context_compacted`, expected one of `error`, `warning`, `task_started`, `task_complete`, `token_count`, `agent_message`, `user_message`, `agent_message_delta`, `agent_reasoning`, `agent_reasoning_delta`, `agent_reasoning_raw_content`, `agent_reasoning_raw_content_delta`, `agent_reasoning_section_break`, `session_configured`, `mcp_startup_update`, `mcp_startup_complete`, `mcp_tool_call_begin`, `mcp_tool_call_end`, `web_search_begin`, `web_search_end`, `exec_command_begin`, `exec_command_output_delta`, `exec_command_end`, `view_image_tool_call`, `exec_approval_request`, `apply_patch_approval_request`, `deprecation_notice`, `background_event`, `undo_started`, `undo_completed`, `stream_error`, `patch_apply_begin`, `patch_apply_end`, `turn_diff`, `get_history_entry_response`, `mcp_list_tools_response`, `list_custom_prompts_response`, `plan_update`, `turn_aborted`, `shutdown_complete`, `entered_review_mode`, `exited_review_mode`, `raw_response_item`, `item_started`, `item_completed`, `agent_message_content_delta`, `reasoning_content_delta`, `reasoning_raw_content_delta`
```
<img width="865" height="1279" alt="Image" src="https://github.com/user-attachments/assets/bd5ee7ad-5ce0-4f3d-bc89-626b65c11c6a" />

Database:
<img width="2560" height="1485" alt="Image" src="https://github.com/user-attachments/assets/e25ea18b-bac4-4485-95bf-91446b72acf7" />

iamriajul:
Same issue.

abcpro1:
The codex session compaction issues have been resolved.

The database deadlock is being solved.

PR #2839, opened 2026-02-19
fix: fix db locked error

KimEbert42:
Being able to tune the sqlite busy timeout allows users using a potato to avoid getting "database is locked" errors, by allowing the queries to finish during IO delays with the disk. This often occurs when running larger queries, and may appear intermittently.

I first noticed the issue when running on a potato of a computer. I found that moving the sqlite database to a ramdisk solved the database issues, but wasn't practical for my use case of using a potato. 

https://www.sqlite.org/c3ref/busy_timeout.html
https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteConnectOptions.html

Example of some of the errors this can resolve.

Failed to fetch scratch:

[0] 2026-02-19T02:40:06.813163Z ERROR local_deployment::container: Failed to update execution process completion: error returned from database: (code: 5) database is locked

[0] 2026-02-19T02:40:01.441862Z ERROR server::middleware::model_loaders: Failed to fetch Workspace 6b783a73-2de4-4a1b-8244-c3aebc5c40d2: error returned from database: (
code: 5) database is locked

This doesn't fix database deadlocks, race conditions, or general query performance, but it does allow the queries to finish. This allows the user, who is using a potato, to be able to get results back and not have to restart the UI after the AI completes a task.

Please test and verify results.

A "Potato of a computer" includes any computer using a spinning disk as the storage for the sqlite database. This frequently causes queries to take beyond the 5 seconds that the default timeout is set to.

### DIFF
--- a/crates/db/src/lib.rs
+++ b/crates/db/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{str::FromStr, sync::Arc};
+use std::{str::FromStr, sync::Arc, time::Duration};
 
 use sqlx::{
     Error, Pool, Sqlite, SqlitePool,
@@ -73,13 +73,22 @@ pub struct DBService {
 }
 
 impl DBService {
+    fn busy_timeout_seconds() -> u64 {
+        std::env::var("SQLITE_BUSY_TIMEOUT_SECONDS")
+            .ok()
+            .and_then(|s| s.trim().parse::<u64>().ok())
+            .unwrap_or(60)
+    }
+
     pub async fn new() -> Result<DBService, Error> {
         let database_url = format!(
             "sqlite://{}",
             asset_dir().join("db.sqlite").to_string_lossy()
         );
+        let busy_timeout = Duration::from_secs(Self::busy_timeout_seconds());
         let options = SqliteConnectOptions::from_str(&database_url)?
             .create_if_missing(true)
+            .busy_timeout(busy_timeout)
             .journal_mode(SqliteJournalMode::Delete);
         let pool = SqlitePool::connect_with(options).await?;
         run_migrations(&pool).await?;
@@ -114,8 +123,10 @@ impl DBService {
             "sqlite://{}",
             asset_dir().join("db.sqlite").to_string_lossy()
         );
+        let busy_timeout = Duration::from_secs(Self::busy_timeout_seconds());
         let options = SqliteConnectOptions::from_str(&database_url)?
             .create_if_missing(true)
+            .busy_timeout(busy_timeout)
             .journal_mode(SqliteJournalMode::Delete);
 
         let pool = if let Some(hook) = after_connect {


KimEbert42:
Maybe related to #2377 #2302 #2042 #1938 #1571 #1059

Juanlucasbg:
## Aggressive review summary — PR #2839

13-line fix in `crates/db/src/lib.rs` adding `SqliteConnectOptions::busy_timeout` to both `SqlitePool` initialization paths, defaulting to 60s and overridable via `SQLITE_BUSY_TIMEOUT_SECONDS`. Verdict: **clean — recommend merge**.

### Why the fix is correct

`SQLITE_BUSY` errors surface when a writer holds the lock and another connection tries to write within the configured wait window. SQLite's `busy_timeout` makes the engine block-and-retry for up to N seconds before erroring. Without it, `SqlitePool::connect_with` defaults to whatever sqlx sets (effectively `0` if not configured at the connection-options level), so any concurrent write hits an immediate `SQLITE_BUSY`. This is the textbook fix.

The fix is consistently applied to both init paths (`DBService::new` and the `new_with_after_connect` variant), so dev/test/prod codepaths all get the same behavior. The env-var escape hatch is well-shaped (default → 60, parse → safe `unwrap_or(60)`).

### Findings

- **Structural**: 0 HIGH, 0 MED, 0 LOW. Symmetric application. `Duration::from_secs(u64)` cannot panic.
- **Adversarial**: 0 HIGH. Adversarial inputs to `SQLITE_BUSY_TIMEOUT_SECONDS` (negative, non-numeric, oversized) all fall through to `unwrap_or(60)`. `u64` overflow would only happen if the value exceeds `u64::MAX` seconds — not a realistic concern.
- **Security**: 0 HIGH. Env-var only reads; no shell exec; no injection surface.
- **Conventions**: PASS. Matches the helper-fn-on-impl style used elsewhere in the crate.

### NITs (non-blocking)

- **WAL would be the more thorough fix.** `journal_mode(SqliteJournalMode::Delete)` (lines 81 / 130 unchanged) keeps the legacy rollback journal — readers and writers contend for the same lock. Switching to `SqliteJournalMode::Wal` allows concurrent reads while a writer holds the lock, which removes most `SQLITE_BUSY` cases entirely. Bigger change, but worth a follow-up issue.
- **60 seconds is a long ceiling.** Common defaults are 5–30s. A 60s `busy_timeout` lets a deadlock pin a request for a full minute before surfacing — fine for desktop UX, less ideal for the server deployment in `crates/remote`. Consider 10s default with the env var staying as the override.
- **Document the env var** in `README.md` under "Environment Variables" and in `CLAUDE.md`'s envs list (currently mentions `FRONTEND_PORT`, `BACKEND_PORT`, `HOST`, etc.).
- **No test added.** `CLAUDE.md` says "Add tests for new logic and edge cases" — a minimal `#[test]` for `busy_timeout_seconds()`'s parsing of malformed values would lock the invariant.

### Verdict

Approve.

— Reviewed by automated single-pass review (db-locking domain triage; full 4-tool battery skipped — diff is 13 lines and the change is a textbook sqlx pattern).

PR #2867, opened 2026-02-20
Fix workspace image uploads dropped during legacy cleanup (Vibe Kanban)

anastasiya1155:
## Summary

Fixes a regression introduced in #2790 (legacy cleanup) where images uploaded during **new workspace creation** were silently dropped. The legacy cleanup rewrote workspace creation from the old task-based flow (`create_task_and_start`) to the new workspace-based flow (`create_and_start_workspace`), but the new flow omitted `image_ids` support entirely. As a result, images uploaded in the first message of a new workspace were never associated with the workspace and never copied to the worktree.

### Backend (Rust)
- **`crates/db/src/models/requests.rs`** — Added `image_ids: Option<Vec<Uuid>>` field to `CreateAndStartWorkspaceRequest`
- **`crates/server/src/routes/task_attempts.rs`** — Added `WorkspaceImage::associate_many_dedup` call in the `create_and_start_workspace` handler to link uploaded images to the newly created workspace
- **`crates/mcp/src/task_server/tools/task_attempts.rs`** — Added `image_ids: None` to the MCP tool's construction of `CreateAndStartWorkspaceRequest`

### Frontend (TypeScript/React)
- **`CreateChatBoxContainer.tsx`** — Wired up the previously unused `getImageIds()` from `useCreateAttachments` and passes the image IDs in the create request payload

### Dead code cleanup
- **`useImageMetadata.ts`** — Removed `taskId` parameter and dead API call to the deleted `/api/images/task/` endpoint
- **`image-node.tsx`** — Removed `useTaskId` import and usage
- **`task-attempt-context.tsx`** — Removed `TaskContext` and `useTaskId` (no longer needed after legacy cleanup)
- **`wysiwyg.tsx`** — Removed `taskId` prop and `TaskContext.Provider` wrapper

### Generated
- **`shared/types.ts`** — Regenerated to include the new `image_ids` field on `CreateAndStartWorkspaceRequest`

## How it works

1. User uploads an image while composing the first message in a new workspace
2. The image is uploaded to the server and gets an ID via `useCreateAttachments`
3. `getImageIds()` collects those IDs when the user submits
4. The IDs are sent as `image_ids` in the `CreateAndStartWorkspaceRequest`
5. The backend handler calls `WorkspaceImage::associate_many_dedup` to link images to the workspace
6. When the workspace executor starts, `copy_images_by_workspace_to_worktree` finds the associations and copies the images into the worktree's `.vibe-images/` directory

### DIFF
--- a/crates/db/src/models/scratch.rs
+++ b/crates/db/src/models/scratch.rs
@@ -146,6 +146,17 @@ pub struct DraftWorkspaceLinkedIssue {
     pub remote_project_id: String,
 }
 
+/// Uploaded image stored in a draft workspace
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+pub struct DraftWorkspaceImage {
+    pub id: Uuid,
+    pub file_path: String,
+    pub original_name: String,
+    #[serde(default)]
+    pub mime_type: Option<String>,
+    pub size_bytes: i64,
+}
+
 /// Data for a draft workspace scratch (new workspace creation)
 #[derive(Debug, Clone, Serialize, Deserialize, TS)]
 pub struct DraftWorkspaceData {
@@ -156,6 +167,8 @@ pub struct DraftWorkspaceData {
     pub executor_config: Option<ExecutorConfig>,
     #[serde(default)]
     pub linked_issue: Option<DraftWorkspaceLinkedIssue>,
+    #[serde(default)]
+    pub images: Vec<DraftWorkspaceImage>,
 }
 
 /// Repository entry in a draft workspace
--- a/crates/server/src/bin/generate_types.rs
+++ b/crates/server/src/bin/generate_types.rs
@@ -24,6 +24,7 @@ fn generate_types_content() -> String {
         db::models::tag::UpdateTag::decl(),
         db::models::scratch::DraftFollowUpData::decl(),
         db::models::scratch::DraftWorkspaceData::decl(),
+        db::models::scratch::DraftWorkspaceImage::decl(),
         db::models::scratch::DraftWorkspaceLinkedIssue::decl(),
         db::models::scratch::DraftWorkspaceRepo::decl(),
         db::models::scratch::DraftIssueData::decl(),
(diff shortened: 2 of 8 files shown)


anastasiya1155:
Responding to the review comments on `useImageMetadata.ts`:

### Why was the `taskId` code block removed? (line 51)

The `taskId` fallback called `/api/images/task/${taskId}/metadata` — this endpoint was **deleted** in the legacy cleanup PR #2790 (commit `139b18158`). The route handler `get_task_image_metadata` was removed from `crates/server/src/routes/images.rs`. So this code block was dead code that would always 404 if reached.

### Why was `shouldFetch` changed? (line 36)

Previously:
```ts
const hasContext = !!taskAttemptId || !!taskId;
const shouldFetch = isVibeImage && hasContext && !localImage;
```

Now:
```ts
const shouldFetch = isVibeImage && !!taskAttemptId && !localImage;
```

Since `taskId` was removed (its API endpoint no longer exists), the `hasContext` check simplifies to just `!!taskAttemptId`. The `|| !!taskId` part was only needed for the now-deleted fallback path. The `queryKey` was also simplified from `['imageMetadata', taskAttemptId, taskId, src]` to `['imageMetadata', taskAttemptId, src]` for the same reason.

Both changes are cleanup of dead code left behind by the legacy cleanup — no behavioral change for any reachable code path.

frontend/src/hooks/useImageMetadata.ts
```diff
@@ -33,28 +32,19 @@ export function useImageMetadata(
...
       if (taskAttemptId) {
         const res = await fetch(
           `/api/task-attempts/${taskAttemptId}/images/metadata?path=${encodeURIComponent(src)}`
         );
         const data = await res.json();
         return data.data as ImageMetadata | null;
       }
-      if (taskId) {
-        const res = await fetch(
-          `/api/images/task/${taskId}/metadata?path=${encodeURIComponent(src)}`
```
anastasiya1155, line 53: endpoint was removed in #2790

anastasiya1155:
> Works for the basic case.
> 
> There is an edge case:
> 
> * upload an image for a new workspace
> * type a prompt
> * reload the page
> 
> In that case the prompt gets restored using the draft mechanism but the image does not. Maybe the `image_ids` can be a field on `DraftWorkspaceData`?

done

PR #2888, opened 2026-02-21
feat: mobile single-panel layout mode for phone screens

yonch:
## Problem

The Vibe Kanban workspace layout assumes a desktop viewport. On phones (~390px wide), fixed-width sidebars (300px each) cause horizontal overflow, `h-screen` (100vh) doesn't account for iOS Safari's dynamic browser chrome, and users must manually toggle panels on/off to see any single piece of content. The navbar toggle buttons work as independent on/off switches, but on mobile you really only want one panel visible at a time.

## Design Decisions

**One panel at a time, radio-style switching**: On screens ≤767px, the 6 existing navbar toggle buttons become mutually exclusive — tapping one shows that panel and hides the others. The six mobile "screens" are: Workspaces list (sidebar), Chat, Changes/Diff, Logs, Preview, and Right sidebar (Git/Terminal/Notes).

**Separate Zustand store**: Mobile panel state lives in its own `useMobileLayoutStore`, completely isolated from `useUiPreferencesStore`. This avoids two problems:
1. The main store's `.subscribe()` fires on _any_ state change with no diffing, which would trigger wasted scratch persistence API calls on every panel switch
2. Desktop panel state stays untouched — no contamination of persisted preferences

**Keep all panels mounted, hide with CSS**: Instead of conditionally rendering (unmount/remount) panels, inactive panels get `display: none` via the `hidden` Tailwind class. This preserves: half-typed messages (session drafts are server-persisted via scratch with 500ms debounce, but cursor/undo history are local), scroll positions, preview iframes (avoids full page reload), and terminal sessions. Components using `ResizeObserver` (xterm `FitAddon`, Virtuoso virtual scroll) automatically re-measure when their container transitions from `display: none` to visible.

**Breakpoint at 767px**: Sits cleanly below Tailwind's `md` breakpoint (768px), avoiding off-by-one ambiguity.

**AppBar kept as-is**: Hiding the vertical AppBar on mobile was considered but would remove project switching, user menu, and sign-in entry points. It stays visible for now.

**Branch name hidden on mobile**: The navbar center section (branch name) is cleared on mobile to save horizontal space. Users can identify their workspace via the sidebar list where the active workspace is highlighted.

**RightSidebar with `rightMainPanelMode={null}`**: On mobile, the right sidebar gets its own dedicated panel, so passing `null` prevents it from showing a redundant upper section (file tree/process list/preview controls) that duplicates the dedicated Changes/Logs/Preview panels.

**`ctx.isMobile` in actions instead of `window.innerWidth`**: Action `isActive`/`isEnabled` callbacks use the reactive context value for consistency and testability. The `execute` callbacks use a `isMobileQuery()` helper (inline `window.matchMedia` check) since `ActionExecutorContext` doesn't carry visibility state.

### New files (2)
- **`frontend/src/hooks/useIsMobile.ts`** — Reactive hook using `useSyncExternalStore` + `matchMedia('(max-width: 767px)')`. Updates on resize/rotation.
- **`frontend/src/stores/useMobileLayoutStore.ts`** — Lightweight Zustand store: `mobileActivePanel` enum (`'sidebar' | 'chat' | 'changes' | 'logs' | 'preview' | 'right-sidebar'`), defaults to `'chat'`, not persisted.

### Modified files (7)
- **`actions/index.ts`** — Added `isMobile` and `mobileActivePanel` to `ActionVisibilityContext`. Six toggle actions (`ToggleLeftSidebar`, `ToggleLeftMainPanel`, `ToggleChangesMode`, `ToggleLogsMode`, `TogglePreviewMode`, `ToggleRightSidebar`) + `ToggleDevServer` now check mobile state: `execute` calls `setMobileActivePanel(...)` on mobile instead of the desktop toggle logic; `isActive` checks `mobileActivePanel` on mobile; `isEnabled` on `ToggleLeftMainPanel` returns `true` on mobile (radio behavior, always enabled).
- **`actions/useActionVisibility.ts`** — Wires `useIsMobile()` and `useMobileLayoutStore` into the `ActionVisibilityContext` returned by `useActionVisibilityContext()`.
- **`containers/SharedAppLayout.tsx`** — Switches root container from `h-screen` to `h-dvh overflow-hidden` on mobile, fixing the iOS Safari 100vh overflow issue.
- **`containers/NavbarContainer.tsx`** — Passes empty string for `workspaceTitle` on mobile.
- **`containers/WorkspacesLayout.tsx`** — Core layout change. On mobile, renders all 6 panels wrapped in `hidden`/`h-full` divs based on `mobileActivePanel`. Guards the desktop-only `rightMainPanelMode === null` reset effect. Adds effect to default to `'sidebar'` when no workspace is selected. Providers (`CreateModeProvider`/`ExecutionProcessesProvider`, `ReviewProvider`, `ChangesViewProvider`) wrap all panels identically to the desktop path.
- **`containers/RightSidebar.tsx`** — Conditionally removes `border-l` on mobile (full-width rendering).
- **`containers/WorkspacesSidebarContainer.tsx`** — Auto-navigates to chat (`setMobileActivePanel('chat')`) when a workspace is selected on mobile.

## Verification

Have been using personally, looks good!

Please also consider #2889

### DIFF
--- a/packages/web-core/src/pages/workspaces/RightSidebar.tsx
+++ b/packages/web-core/src/pages/workspaces/RightSidebar.tsx
@@ -9,6 +9,8 @@ import { useChangesView } from '@/shared/hooks/useChangesView';
 import { useWorkspaceContext } from '@/shared/hooks/useWorkspaceContext';
 import { ArrowsOutSimpleIcon } from '@phosphor-icons/react';
 import { useLogsPanel } from '@/shared/hooks/useLogsPanel';
+import { useIsMobile } from '@/shared/hooks/useIsMobile';
+import { cn } from '@/shared/lib/utils';
 import type { RepoWithTargetBranch, Workspace } from 'shared/types';
 import {
   PERSIST_KEYS,
@@ -48,6 +50,7 @@ export function RightSidebar({
   const { selectFile } = useChangesView();
   const { diffs } = useWorkspaceContext();
   const { setExpanded } = useExpandedAll();
+  const isMobile = useIsMobile();
   const isTerminalVisible = useUiPreferencesStore((s) => s.isTerminalVisible);
   const { expandTerminal, isTerminalExpanded } = useLogsPanel();
 
@@ -187,7 +190,12 @@ export function RightSidebar({
   }
 
   return (
-    <div className="h-full border-l bg-secondary overflow-y-auto">
+    <div
+      className={cn(
+        'h-full bg-secondary overflow-y-auto',
+        !isMobile && 'border-l'
+      )}
+    >
       <div className="divide-y border-b">
         {sections
           .filter((section) => section.visible)
--- a/packages/web-core/src/pages/workspaces/WorkspacesLayout.tsx
+++ b/packages/web-core/src/pages/workspaces/WorkspacesLayout.tsx
@@ -20,6 +20,8 @@ import { CreateChatBoxContainer } from '@/shared/components/CreateChatBoxContain
 import { PreviewBrowserContainer } from './PreviewBrowserContainer';
 import { WorkspacesGuideDialog } from '@/shared/dialogs/shared/WorkspacesGuideDialog';
 import { useUserSystem } from '@/shared/hooks/useUserSystem';
+import { useIsMobile } from '@/shared/hooks/useIsMobile';
+import { useMobileLayoutStore } from '@/shared/stores/useMobileLayoutStore';
 
 import {
   PERSIST_KEYS,
@@ -75,6 +77,12 @@ export function WorkspacesLayout() {
     setLeftMainPanelVisible,
   } = useWorkspacePanelState(isCreateMode ? undefined : workspaceId);
 
+  const isMobile = useIsMobile();
+  const mobileActivePanel = useMobileLayoutStore((s) => s.mobileActivePanel);
+  const setMobileActivePanel = useMobileLayoutStore(
+    (s) => s.setMobileActivePanel
+  );
+
   const {
     config,
     updateAndSaveConfig,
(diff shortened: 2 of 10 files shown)


yonch (commented):
Nice improvement overall — the mobile single-panel experience is much clearer, and reusing the existing action system to switch panels is a good approach. The `h-dvh` tweak in `SharedAppLayout.tsx` is also a good mobile fix.

Main concern (perf/side-effects): in `WorkspacesLayout.tsx` the mobile mode renders *all* panels and toggles visibility via Tailwind’s `hidden` class. That keeps `WorkspacesMainContainer`, `ChangesPanelContainer`, `LogsContentContainer`, `PreviewBrowserContainer`, and `RightSidebar` mounted even when not active, which can lead to unnecessary subscriptions/renders and potentially expensive work (especially preview/logs). Consider conditionally rendering only the active panel’s content (e.g. `mobileActivePanel === 'preview' && ...`) so inactive panels truly unmount.

Smaller nits:
- `Actions` uses an inline `isMobileQuery()` while the rest of the UI uses `useIsMobile`. It’s fine, but I’d prefer sharing the breakpoint constant/util and guarding `window` access (`typeof window !== 'undefined'`) to avoid surprises in non-DOM test environments.
- `useIsMobile` uses `mql.addEventListener('change', ...)`; if older Safari support matters, consider a fallback to `addListener/removeListener`.
- A couple action labels like “Hide Chat Panel” may be a bit misleading on mobile (since it’s more “switch to chat” than hide/show).

Otherwise looks clean and consistent with the existing layout/store patterns.

yonch (commented):
Thanks for the review! Addressing each point:

**Keep-mounted vs conditional rendering**: This was a deliberate design decision. Unmounting panels on mobile would lose:
- Half-typed chat messages (the session draft has a 500ms debounce to server — anything typed in the last 500ms before unmount could be lost, plus cursor position and undo history are local state)
- Preview browser iframes (full page reload on every panel switch)
- Terminal sessions (xterm.js instance destroyed/recreated)
- Scroll positions in conversation list and changes panel

The desktop layout already keeps these components mounted when visible, so the mobile memory cost is comparable. Components using `ResizeObserver` (xterm `FitAddon`, Virtuoso virtual scroll) automatically re-measure when their container transitions from `display: none` to visible, so there's no measurement issue.

**Shared breakpoint constant + typeof window guard**: Fixed in `14bcfbfd` — `isMobileQuery()` is now exported from `useIsMobile.ts` with a `typeof window` guard, and `actions/index.ts` imports it instead of defining its own inline check.

**addEventListener on MediaQueryList**: `addEventListener('change')` is supported since Safari 14 (Sep 2020). The deprecated `addListener` API would only be needed for Safari 13 and below, which is well past EOL. Keeping the modern API.

**Action labels on mobile**: Agreed the "Hide/Show" labels are slightly misleading on mobile (it's more "switch to"). However, these labels only appear in the command bar search results, not on the navbar icon buttons themselves. On mobile the user interacts with the icons, so the impact is minimal. Happy to update if you feel strongly.

ciekawy:
check this https://github.com/BloopAI/vibe-kanban/pull/2889 also - seems we started on the same around same time :)

yonch:
rebased after the upstream frontend refactoring -- ready for review!

vdmkotai:
Thanks a lot for the contribution.

Because we substantially changed the architecture, I had to manually port the changes to another PR #2947

yonch:
Thank you @vdmkotai !

PR #2917, opened 2026-02-25
Create SECURITY.md

lesignals:


mickmister:
@stunningpixels There should probably be a responsible security disclosure process documented (if not already) to avoid public disclosures like this

stunningpixels:
@mickmister I think @lesignals is a bot and the advisory is AI generated, I advise against hosting VK on an open network.

And we just launched: https://www.vibekanban.com/docs/responsible-disclosure

PR #2938, opened 2026-02-26
fix: preserve initial prompt when setup script fails (Vibe Kanban)

### LINKED ISSUE #2847: Initial prompt gets lost when the setup script fails
There is no way to get back to the initial prompt when the setup script fails.

anastasiya1155:
## Summary

When a setup script fails, the initial user prompt was lost because the `CodingAgentInitialRequest` process never starts — so there was no process to render the user message from. This PR preserves the initial prompt so users can see, edit, and retry their original message after a setup script failure.

## Changes

- **Extract prompt from action chain**: Added `extractPromptFromActionChain` helper that walks the `next_action` chain on `ExecutorAction` to find the initial coding agent prompt embedded in the setup script's action chain.
- **Emit user message from setup script branch**: Instead of emitting the initial user message from the `CodingAgentInitialRequest` branch (which caused a Virtuoso list reorder when script and coding agent processes loaded at different times), the user message is now emitted from the `ScriptRequest` branch — but only **after** the setup script finishes (success or failure). This preserves the original UX where the user message appears below the completed script entry.
- **Suppress duplicate user message**: When a setup script exists, the `CodingAgentInitialRequest` branch no longer emits its own user message (the script branch handles it instead), preventing duplicate messages and list reordering.
- **Filter invisible Virtuoso entries**: `next_action` and `token_usage_info` entries that render as `null` in the conversation display are now filtered out before being passed to Virtuoso, preventing empty list items that add unwanted spacing.
- **Distinct patchKey for script entries**: Script tool entries use a `'script'` key suffix instead of numeric `0`, so Virtuoso doesn't reuse stale heights from the streaming state.
- **Eager script loading**: Script processes are loaded first in `loadInitialEntries` to ensure correct ordering from the first render.

## Why

Previously, when the setup script → coding agent chain was interrupted by a script failure, the user's original prompt simply disappeared. The user had no way to see what they originally typed, making it impossible to edit and retry. This fix ensures the prompt is always preserved and visible after the setup script completes, regardless of outcome.

The Virtuoso-related fixes address a secondary visual bug: when the user message and script entry loaded at different times, they would swap positions in the list, causing Virtuoso's cached item heights to produce visible empty space.

Closes #2847

---

### DIFF
--- a/packages/web-core/src/features/workspace-chat/model/hooks/useConversationHistory.ts
+++ b/packages/web-core/src/features/workspace-chat/model/hooks/useConversationHistory.ts
@@ -574,32 +574,11 @@ export const useConversationHistory = ({
 
       if (!executionProcesses?.current) return localDisplayedExecutionProcesses;
 
-      // Always load script processes (setup/cleanup/archive) first so they
-      // appear at their correct position from the very first render.  This
-      // avoids a reorder when the remaining-batch pass loads them later,
-      // which causes Virtuoso to apply stale item sizes.
-      const scriptProcesses = executionProcesses.current.filter(
-        (ep) =>
-          ep.status !== ExecutionProcessStatus.running &&
-          ep.executor_action.typ.type === 'ScriptRequest'
-      );
-      for (const ep of scriptProcesses) {
-        const entries = await loadEntriesForHistoricExecutionProcess(ep);
-        const entriesWithKey = entries.map((e, idx) =>
-          patchWithKey(e, ep.id, idx)
-        );
-        localDisplayedExecutionProcesses[ep.id] = {
-          executionProcess: ep,
-          entries: entriesWithKey,
-        };
-      }
-
       for (const executionProcess of [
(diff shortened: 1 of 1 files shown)


LSRCT:
Looks related to #2861

anastasiya1155:
> Looks related to #2861

@stunningpixels Is it a duplicate? I did not notice the fix on main so not sure

packages/web-core/src/features/workspace-chat/ui/ConversationListContainer.tsx
```diff
@@ -291,7 +277,22 @@ export const ConversationList = forwardRef<
 
       const aggregatedEntries = aggregateConsecutiveEntries(pending.entries);
 
-      setChannelData({ data: aggregatedEntries, scrollModifier });
+      // Filter out entries that render as null in the new design –
```
stunningpixels, line 280: Is this change related to the objective of this PR?

PR #2974, opened 2026-02-28
Simplify WYSIWYG editor: plain-text markdown editing with preview toggle (Vibe Kanban)

anastasiya1155:
## Summary

Replaces the buggy live markdown-rendering WYSIWYG editor with a simpler plain-text editing mode where users type raw markdown syntax, with a preview toggle to see the rendered output.

- **Remove `MarkdownShortcutPlugin`** — the live markdown-as-you-type conversion was the primary source of editor bugs. Users now type `**bold**` as literal text instead of it being auto-converted to rich text.
- **New `MarkdownInsertPlugin`** — intercepts `FORMAT_TEXT_COMMAND` to insert markdown syntax markers (e.g. `**`, `*`, `~~`, `` ` ``) as literal text instead of applying Lexical rich text formatting. Also handles a new `INSERT_MARKDOWN_LIST_COMMAND` for `- ` and `1. ` prefixes.
- **Standalone preview toggle** — every editable editor gets an Eye/PencilLine button (top-right corner) that toggles between edit mode and a read-only rendered preview. The preview renders a nested `<WYSIWYGEditor disabled>` with full display transformers.
- **Split transformer architecture** — `editTransformers` (custom elements + text format transformers only) vs `displayTransformers` (full markdown rendering). Edit mode preserves raw markdown; display mode renders everything.
- **Strip backslash escapes** — added `preserveMarkdownSyntax` option to `MarkdownSyncPlugin` that strips Lexical's automatic escaping of `*`, `_`, `~` etc. in the `onChange` export path, so the stored value contains the user's intended markdown syntax.
- **Toolbar simplification** — removed active state tracking (bold/italic/etc. indicators no longer meaningful), removed Underline button (no markdown equivalent), rewired list buttons to use `INSERT_MARKDOWN_LIST_COMMAND`.

## Why

The WYSIWYG editor's live markdown rendering via `MarkdownShortcutPlugin` was difficult to use and produced many edge-case bugs. By switching to plain-text editing with an explicit preview toggle, users get a predictable editing experience while still being able to format with toolbar buttons and see rendered output on demand. Custom elements (images, PR comments, component info) continue to render inline in both modes.

## Implementation details

- **Frozen selection fix**: Lexical freezes selection objects inside command handlers. The plugin creates fresh `RangeSelection` objects via `$createRangeSelection()` + `$setSelection()` for cursor positioning after marker insertion.
- **`useImperativeHandle` guard**: The nested preview editor doesn't receive a ref prop, which in React strict mode causes a frozen empty object to be passed. A guard checks for a valid ref before calling `useImperativeHandle`.
- **`hideActions` prop**: Prevents the read-only action buttons (Copy, Edit, Delete) from appearing on the nested preview editor.
- **Preview toggle z-index**: The toggle button is positioned outside `<LexicalComposer>` with `absolute top-0 right-0 z-20` to prevent the preview content from occluding it.

## Files changed

| File | Change |
|------|--------|
| `MarkdownInsertPlugin.tsx` | New plugin — intercepts format commands to insert markdown syntax as text |
| `MarkdownSyncPlugin.tsx` | Added `preserveMarkdownSyntax` to strip backslash escapes on export |
| `WYSIWYGEditor.tsx` | Split transformers, added preview toggle, replaced MarkdownShortcutPlugin |
| `ToolbarPlugin.tsx` | Removed active state tracking and Underline button |
| `StaticToolbarPlugin.tsx` | Removed active state, rewired lists to INSERT_MARKDOWN_LIST_COMMAND |
| `KeyboardCommandsPlugin.tsx` | Removed Cmd+E inline code handler (now handled by MarkdownInsertPlugin) |

## Test plan

- [ ] Type markdown syntax (`**bold**`, `*italic*`, `` `code` ``) — should appear as literal text
- [ ] Click toolbar Bold/Italic/Strikethrough/Code buttons — should insert markdown markers
- [ ] Click preview toggle (eye icon) — should render markdown (bold, italic, lists, headings, etc.)
- [ ] Click edit toggle (pencil icon) — should return to raw text editing
- [ ] Verify custom elements (images, PR comments) render in both edit and preview modes
- [ ] Verify keyboard shortcuts (Cmd+B, Cmd+I) insert markdown markers
- [ ] Verify no console errors when toggling preview

### DIFF
--- a/packages/web-core/src/pages/kanban/IssueWorkspacesSectionContainer.tsx
+++ b/packages/web-core/src/pages/kanban/IssueWorkspacesSectionContainer.tsx
@@ -180,8 +180,9 @@ export function IssueWorkspacesSectionContainer({
       return;
     }
 
-    const { WorkspaceSelectionDialog } =
-      await import('@/shared/dialogs/command-bar/WorkspaceSelectionDialog');
+    const { WorkspaceSelectionDialog } = await import(
+      '@/shared/dialogs/command-bar/WorkspaceSelectionDialog'
+    );
     await WorkspaceSelectionDialog.show({ projectId, issueId });
   }, [projectId, issueId]);
 
--- a/packages/web-core/src/pages/kanban/KanbanIssuePanelContainer.tsx
+++ b/packages/web-core/src/pages/kanban/KanbanIssuePanelContainer.tsx
@@ -616,8 +616,9 @@ export function KanbanIssuePanelContainer({
 
         // For statusId, open the status selection dialog with callback
         if (field === 'statusId') {
-          const { ProjectSelectionDialog } =
-            await import('@/shared/dialogs/command-bar/selections/ProjectSelectionDialog');
+          const { ProjectSelectionDialog } = await import(
+            '@/shared/dialogs/command-bar/selections/ProjectSelectionDialog'
+          );
           const result = await ProjectSelectionDialog.show({
             projectId,
             selection: { type: 'status', issueIds: [], isCreateMode: true },
@@ -636,8 +637,9 @@ export function KanbanIssuePanelContainer({
 
         // For priority, open the priority selection dialog with callback
         if (field === 'priority') {
-          const { ProjectSelectionDialog } =
-            await import('@/shared/dialogs/command-bar/selections/ProjectSelectionDialog');
+          const { ProjectSelectionDialog } = await import(
+            '@/shared/dialogs/command-bar/selections/ProjectSelectionDialog'
+          );
           const result = await ProjectSelectionDialog.show({
             projectId,
             selection: { type: 'priority', issueIds: [], isCreateMode: true },
@@ -659,8 +661,9 @@ export function KanbanIssuePanelContainer({
 
         // For assigneeIds, open the assignee selection dialog with callback
         if (field === 'assigneeIds') {
-          const { AssigneeSelectionDialog } =
-            await import('@/shared/dialogs/kanban/AssigneeSelectionDialog');
+          const { AssigneeSelectionDialog } = await import(
+            '@/shared/dialogs/kanban/AssigneeSelectionDialog'
+          );
           await AssigneeSelectionDialog.show({
             projectId,
             issueIds: [],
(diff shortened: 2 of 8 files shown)


vdmkotai:
@anastasiya1155 just let's make sure that click-to-component works as expected for the dev preview

stunningpixels:
I was expecting to see the raw backticks from the code block, but looks like it's rendering the visual code block in both

Screenshot 2026-03-02 at 17 02 35
Screenshot 2026-03-02 at 17 02 32

stunningpixels:
Screenshot 2026-03-02 at 17 33 00
Small thing but images not rendering in preview mode

stunningpixels:
Screenshot 2026-03-02 at 17 38 50
Doesn't make sense to have these in the issue toolbar anymore

stunningpixels:
Maybe we can toggle between raw <> editable preview? Rather than raw <> view-only preview?

anastasiya1155:
> I was expecting to see the raw backticks from the code block, but looks like it's rendering the visual code block in both
> Small thing but images not rendering in preview mode

both are fixed

> Doesn't make sense to have these in the issue toolbar anymore

Why? It is useful to have buttons that help with formatting, just like bold, italic, code etc

> Maybe we can toggle between raw <> editable preview? Rather than raw <> view-only preview?

Not sure if it makes sense. It would be just like it is right now but with a toggle to edit raw markdown and users might not even notice this button. If they try to edit preview they will have the same issues we started from. I think it makes more sense in github where you either edit or preview.

stunningpixels:
> Why? It is useful to have buttons that help with formatting, just like bold, italic, code etc

If you click it, it adds `-` to the start of the sentence, but if you then press enter it doesn't add `-` to the next line. We could do this the same as GitHub, where they have some markdown automation, eg auto add list

anastasiya1155:
> > Why? It is useful to have buttons that help with formatting, just like bold, italic, code etc
> 
> If you click it, it adds `-` to the start of the sentence, but if you then press enter it doesn't add `-` to the next line. We could do this the same as GitHub, where they have some markdown automation, eg auto add list

done

PR #3011, opened 2026-03-03
feat: localStorage-backed scratch storage for remote-web (Vibe Kanban)

vdmkotai:
## Summary

Add a runtime-aware scratch storage adapter so that temporary UI state (draft issues, workspace creation drafts, UI preferences, preview settings, workspace notes, etc.) **persists across page refreshes** on `cloud.vibekanban.com` via `localStorage`, while the local app continues using the existing server-side WebSocket/API scratch backed by SQLite.

## Why

On **remote-web**, there is no scratch API — the remote backend (ElectricSQL + Postgres) doesn't expose the REST/WebSocket scratch endpoints that the local backend provides. This meant every `useScratch` call silently failed on remote, and all temporary UI state was lost on page refresh. Users creating issues, editing workspace drafts, or adjusting UI preferences would lose everything on a simple reload.

`localStorage` is the right abstraction here because scratch data is **ephemeral, per-device state** — not something that needs cross-device server sync. The remote app runs on a stable domain (`cloud.vibekanban.com`), so `localStorage` is reliable. (On local-web, ports are randomized per session, making `localStorage` unsuitable — the existing server-side scratch continues to work there.)

### New Files (3)

| File | Purpose |
|---|---|
| `useLocalStorageScratch.ts` | localStorage-backed hook matching the `UseScratchResult` interface. Includes cross-tab sync via `StorageEvent`, graceful error handling for quota/private-browsing. Exports standalone `localStorageScratchUpdate`/`localStorageScratchDelete` helpers for non-hook code. |
| `useRuntimeScratch.ts` | Thin wrapper — checks `useAppRuntime()` and delegates to `useScratch` (local) or `useLocalStorageScratch` (remote). Both hooks are always called (React rules) but gated via `enabled` flag. |
| `useKanbanIssueComposerScratch.ts` | Syncs the Zustand kanban issue composer store to `localStorage` on remote. Subscribes to store changes, serializes to storage, and hydrates on mount. No-op on local runtime. |

### Modified Files (12)

**6 scratch consumers migrated** from `useScratch` → `useRuntimeScratch`:
- `useUiPreferencesScratch.ts` — UI panel states, sidebar, filters, sort preferences
- `useCreateModeState.ts` — Workspace creation draft state
- `WorkspacesSidebarContainer.tsx` — Workspace draft in sidebar
- `useSessionMessageEditor.ts` — Draft task / follow-up messages
- `useWorkspaceNotes.ts` — Workspace notes
- `usePreviewSettings.ts` — Preview URL, device, zoom settings

**Workspace create draft** (`workspaceCreateState.ts`):
- `persistWorkspaceCreateDraft()` is a plain function (not a hook), so it now accepts a `runtime` parameter and calls `localStorageScratchUpdate()` directly when on remote.

**Action system plumbing**:
- `ActionExecutorContext` (`actions.ts`) — added `runtime: AppRuntime` field
- `ActionsProvider.tsx` — reads `useAppRuntime()` and passes `runtime` into executor context
- `shared/actions/index.ts` — two `persistWorkspaceCreateDraft` call sites now pass `ctx.runtime`
- `useProjectWorkspaceCreateDraft.ts` — passes `runtime` to persist function

**Remote-web root** (`__root.tsx`):
- Wired up `useUiPreferencesScratch()` (was previously only in local-web)
- Wired up `useKanbanIssueComposerScratch()` for issue draft persistence

## Error Handling

All localStorage operations are wrapped in try/catch with **silent failure**:
- Quota exceeded → writes are no-ops
- Private browsing restrictions → writes are no-ops, reads return `null`
- Corrupted JSON in storage → reads return `null`
- StorageEvent with bad data → ignored

No localStorage failure can break the UI. Worst case: data doesn't persist, matching the pre-change behavior.

## Storage Key Format

- Scratch system: `vk-scratch:{ScratchType}:{id}` (e.g. `vk-scratch:DRAFT_TASK:abc-123`)
- Composer store: `vk-kanban-issue-composer`

---

> This PR was written using [Vibe Kanban](https://vibekanban.com)

---

> [!NOTE]
> **Medium Risk**
> Touches the shared `useScratch` hook and kanban create/comment flows, so regressions could impact draft persistence and initialization across both runtimes; local behavior is gated but still shares codepaths.
> 
> **Overview**
> Adds **runtime-aware scratch storage**: `useScratch` now uses WebSocket/API scratch on local runtime and a new `useLocalStorageScratch` adapter on remote runtime, enabling draft/prefs persistence across refreshes on `remote-web`.
> 
> Wires remote-web to actually initialize persistence (`useUiPreferencesScratch`, new `useKanbanIssueComposerScratch`) and updates kanban create flows to restore issue composer drafts from storage; comment composer now autosaves a per-issue draft and clears it on submit.
> 
> Plumbs `runtime` through `ActionExecutorContext`/providers so actions and `persistWorkspaceCreateDraft` can persist workspace-create drafts via localStorage when running remotely.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 92188921b4ffca00d9a8adc801f7a071bec3441e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/web-core/src/pages/kanban/IssueCommentsSectionContainer.tsx
+++ b/packages/web-core/src/pages/kanban/IssueCommentsSectionContainer.tsx
@@ -81,7 +81,8 @@ function IssueCommentsSectionContent() {
     commentDraftScratch?.payload?.type === 'DRAFT_TASK'
       ? commentDraftScratch.payload.data
       : undefined;
-  const hasLoadedCommentDraftRef = useRef(false);
+  const hydratedCommentDraftIdRef = useRef<string | null>(null);
+  const skipNextPersistRef = useRef(false);
 
   const persistCommentDraft = useCallback(
     async (value: string) => {
@@ -111,22 +112,32 @@ function IssueCommentsSectionContent() {
 
   useEffect(() => {
     cancelDebouncedPersistCommentDraft();
-    hasLoadedCommentDraftRef.current = false;
+    hydratedCommentDraftIdRef.current = null;
+    skipNextPersistRef.current = false;
     setCommentInput('');
   }, [commentDraftId, cancelDebouncedPersistCommentDraft]);
 
   useEffect(() => {
     if (isCommentDraftLoading) return;
-    if (hasLoadedCommentDraftRef.current) return;
+    if (hydratedCommentDraftIdRef.current === commentDraftId) return;
 
-    hasLoadedCommentDraftRef.current = true;
-    setCommentInput(commentDraft ?? '');
-  }, [isCommentDraftLoading, commentDraft]);
+    const nextCommentInput = commentDraft ?? '';
+    const shouldSkipNextPersist = nextCommentInput !== commentInput;
+
+    hydratedCommentDraftIdRef.current = commentDraftId;
+    skipNextPersistRef.current = shouldSkipNextPersist;
+    setCommentInput(nextCommentInput);
+  }, [isCommentDraftLoading, commentDraft, commentDraftId, commentInput]);
 
   useEffect(() => {
-    if (!hasLoadedCommentDraftRef.current) return;
+    if (hydratedCommentDraftIdRef.current !== commentDraftId) return;
+    if (skipNextPersistRef.current) {
+      skipNextPersistRef.current = false;
+      return;
+    }
+
     debouncedPersistCommentDraft(commentInput);
-  }, [commentInput, debouncedPersistCommentDraft]);
+  }, [commentInput, commentDraftId, debouncedPersistCommentDraft]);
 
   const handleCommentMarkdownInsert = useCallback((markdown: string) => {
     setCommentInput((prev) =>


packages/remote-web/src/app/providers/RemoteActionsProvider.tsx
```diff
@@ -127,6 +127,7 @@ export function RemoteActionsProvider({
       kanbanProjectId: projectId,
       projectMutations: projectMutations ?? undefined,
       remoteWorkspaces: userCtx?.workspaces ?? [],
+      runtime: "remote",
```
stunningpixels: Is this different from packages/web-core/src/shared/hooks/useAppRuntime.tsx

packages/web-core/src/shared/hooks/useRuntimeScratch.ts
stunningpixels, line 1: Should this be consolidated with useScratch?

packages/remote-web/src/app/providers/RemoteActionsProvider.tsx
vdmkotai: Good catch — replaced the hardcoded `"remote"` with `useAppRuntime()` so it stays consistent with the existing runtime context. Fixed in f37a15f.

packages/web-core/src/shared/hooks/useRuntimeScratch.ts
vdmkotai: Agreed — consolidated into `useScratch` itself. It now checks `useAppRuntime()` internally and delegates to `useLocalStorageScratch` when remote. Deleted `useRuntimeScratch.ts`, all consumers import from `useScratch` as before. Fixed in f37a15f.

packages/web-core/src/shared/hooks/useLocalStorageScratch.ts
```diff
@@ -0,0 +1,141 @@
...
+  id: string,
+  options?: UseLocalStorageScratchOptions
+): UseScratchResult => {
+  const enabled = (options?.enabled ?? true) && id.length > 0;
+  const storageKey = buildStorageKey(scratchType, id);
+
+  const [scratch, setScratch] = useState<Scratch | null>(() =>
+    enabled ? readFromStorage(storageKey) : null
+  );
+  const [isInitialized, setIsInitialized] = useState(false);
```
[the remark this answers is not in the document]
vdmkotai: Fixed in e84a8184. `isInitialized` now starts as `enabled` — since `useState` already synchronously reads from localStorage, data is available on the first render and the extra loading cycle is eliminated.

packages/web-core/src/shared/hooks/useLocalStorageScratch.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const next: Scratch = {
+    id: existing?.id ?? id,
+    payload: update.payload,
+    created_at: existing?.created_at ?? now,
+    updated_at: now,
+  };
+
+  writeToStorage(key, next);
+}
```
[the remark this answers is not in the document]
vdmkotai: Fixed in e84a8184. Extracted `buildScratchEntry()` helper used by both `localStorageScratchUpdate` and the hook's `updateScratch` callback.

Issue #2316, opened 2026-01-26
error sending request for url (https://api.vibekanban.com/v1/tokens/refresh)


ahkimkoo:
```
Remote call failed, retrying after 2.23s: network error: error sending request for url (https://api.vibekanban.com/v1/tokens/refresh)
```

There are a large number of error logs. What is this for? Why is it necessary to access this address?

Teo-ShaoWei:
I am facing a similar issue because my company's Cloudflare WARP is blocking it. Not sure what the ideal solution is in this case. Is there a configuration I can do on Vibe Kanban side, or do I need to approach the folks in my company who maintain the access restriction? Hope to get some advice on this 🙏

Teo-ShaoWei:
Some more context. My company has an always-on Cloudflare WARP. I could disconnect briefly to complete the Vibe Kanban OAuth. However, most of my workflow requires accessing company resources, so I have to turn on Cloudflare WARP most of the time.

Previously, I recall having to log in about once a day in the morning, and then I could continue using it with Cloudflare WARP turned on. However, in the latest Vibe Kanban version, every 30sec or so, it will ask me to OAuth again or get signed out.

Since the token refresh failed to send the request, it doesn't even trigger a refresh on the server side. In this case, I would expect the behaviour to be to continue using the existing token until it expires.

stunningpixels:
I think we can improve login stability when network is interrupted

Teo-ShaoWei:
@stunningpixels specifically on the TLS side for Cloudflare WARP, it might be sufficient for me if `reqwest` is updated to at least [v0.13](https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0), which defaults to `rustls`. Then, instead of `rustls-tls-webpki-roots-no-provider`, use the default [`rustls-platform-verifier`](https://github.com/rustls/rustls-platform-verifier), which will retrieve Cloudflare's CA from my OS cert store.

Would it be something that's worth a try? 🙏

Teo-ShaoWei:
I attempted the fix, feels like it works! 😂 #vibe

Teo-ShaoWei:
@abcpro1 Kudos for the good work on #3090 🚀  I've tried the official build of [Vibe Kanban v0.1.28](https://github.com/BloopAI/vibe-kanban/releases/tag/v0.1.28-20260310182008) and it works 🎉

[...]

PR #3089, opened 2026-03-10
fix(deps): support OS certificate store

Teo-ShaoWei:
## Context

`reqwest` 0.12 used `webpki-roots`, which bundles a hardcoded Mozilla CA set. This breaks TLS connections behind corporate proxies that install their own CA into the OS certificate store but not into the Mozilla bundle.

`reqwest` 0.13 switches to `rustls-platform-verifier`, which reads the OS certificate store (macOS Keychain, Windows cert store, etc.), allowing TLS to work transparently behind corporate proxies without any special configuration.

With `rustls-platform-verifier` handling OS-trusted CAs properly, the `danger_accept_invalid_certs` debug bypass in `remote_client.rs` is no longer needed — it was a workaround for the same class of TLS failures. The method is also deprecated in `reqwest` 0.13, replaced by `tls_danger_accept_invalid_certs`.

## Changes

- Upgrade `reqwest` from 0.12 to 0.13 in workspace, review, and remote `Cargo.toml` files
- Replace `rustls-tls-webpki-roots-no-provider` feature with `rustls-no-provider` (renamed in `reqwest` 0.13)
- Add `query` feature flag (moved out of `json` in `reqwest` 0.13)
- Remove `danger_accept_invalid_certs` debug bypass in `remote_client.rs`
- Add `reqwest` 0.12 to `crates/utils` for cargo feature unification with sentry 0.41's internal `reqwest` 0.12 dependency (sentry has not yet published a version that depends on `reqwest` 0.13)

Contributes to #2316

---

> [!NOTE]
> **Medium Risk**
> Upgrades core HTTP/TLS dependencies (`reqwest`/`rustls` integration) across multiple crates, which can subtly change certificate validation and request behavior. Also introduces a temporary split-brain dependency (`reqwest` 0.12 in `utils`) that could cause feature/unification surprises until `sentry` updates.
> 
> **Overview**
> Upgrades the workspace (and `remote`/`review`) from `reqwest` 0.12 to 0.13, updating feature flags (notably adding `query` and switching to `rustls-no-provider`) so TLS validation can rely on the OS certificate store.
> 
> Removes the `danger_accept_invalid_certs` debug-only bypass in `RemoteClient::new`, and pins `reqwest` 0.12 in `utils` as a temporary compatibility shim to enable TLS features for `sentry`’s still-0.12 `reqwest` dependency. `Cargo.lock` is refreshed accordingly (new `rustls-platform-verifier`/native-cert dependencies and related platform crates).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6fcf6bab069fdb02ff99bcc56f470aa5720ac259. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -20,6 +20,10 @@ rust-embed = "8.2"
 directories = "6.0.0"
 open = "5.3.2"
 regex = "1.11.1"
+# TODO: Remove once sentry publishes a version that depends on reqwest 0.13.
+# reqwest 0.12 is added here to enable TLS features for sentry's internal reqwest 0.12 dependency
+# via cargo feature unification (the workspace uses reqwest 0.13 which doesn't unify with 0.12).
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-webpki-roots-no-provider"] }
 sentry = { version = "0.41.0", default-features = false, features = ["anyhow", "backtrace", "panic", "debug-images", "reqwest"] }
 sentry-tracing = { version = "0.41.0", default-features = false, features = ["backtrace"] }
 json-patch = "2.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -687,6 +687,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cesu8"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
+
+[[package]]
 name = "cexpr"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


Teo-ShaoWei:
I've verified that the fix works for my case by removing `danger_accept_invalid_certs` first. The local dev server becomes unable to use OAuth when my Cloudflare WARP is in use.

After that, the fix was done, and I verified that I can OAuth again 🚀 

While Sentry is still using `reqwest` v0.12, seems like it doesn't affect the OAuth API calls. I guess for me, the telemetry to Sentry will fail since the following is done:
```
[sentry-vite-plugin] Info: Sending telemetry data on issues and performance to Sentry. To disable telemetry, set `options.telemetry` to `false`.
```
but I assume it doesn't affect the usability 🤔 Will continue to test it out locally.

Otherwise, if possible I hope to get this merged first and we fix this along the way 🙏

crates/utils/Cargo.toml
```diff
@@ -20,6 +20,10 @@ rust-embed = "8.2"
 directories = "6.0.0"
 open = "5.3.2"
 regex = "1.11.1"
+# TODO: Remove once sentry publishes a version that depends on reqwest 0.13.
+# reqwest 0.12 is added here to enable TLS features for sentry's internal reqwest 0.12 dependency
+# via cargo feature unification (the workspace uses reqwest 0.13 which doesn't unify with 0.12).
+reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-webpki-roots-no-provider"] }
```
Teo-ShaoWei, line 26: Currently we need this, which will cause `reqwest` to have multiple versions in the dependency tree. However this is only used for `sentry-rust`, and shouldn't result in breakage due to cross-versions.

Anyway, [getsentry/sentry-rust#1000](https://github.com/getsentry/sentry-rust/issues/1000) has been merged and so we should be able to upgrade to the newer `sentry-rust` and do away with this extra import soon.

abcpro1:
Please test #3090

Teo-ShaoWei:
> Please test #3090

That works locally for me, too 👍

Let's publish it? I'll try the tagged release build and let us know if there's any issue.

PR #3100, opened 2026-03-11
Add issue filters to MCP (Vibe Kanban)

### LINKED ISSUE #3088: Feature Request: list_issues MCP tool should support manual kanban sort order
the issues returned by MCP tool `list_issues` are sorted differently from how they are manually ordered in my kanban, 
is it possible to support a "sort" option on this tool? or default to manual sort order?

for example, in my to do list I have manually ordered the 3 tasks to be JLI-16 > JLI-15 > JLI-17 (top > down) 
but MCP always retrieves them in the order of JLI-15 > JLI-16 > JLI-17 no matter how I manually re-order them.

I tried tweaking the parameters but the current specification does not include a "sort" option: 
```
  Tool name: list_issues
  Full name: mcp__vibe_kanban__list_issues

  Description:
  List all the issues in a project. `project_id` is optional if running inside a workspace linked to a remote project.

  Parameters:
    • project_id: string - The ID of the project to list issues from. Optional if running inside a workspace linked to a
    remote project.
    • limit: integer - Maximum number of issues to return (default: 50)
    • offset: integer - Number of results to skip before returning rows (default: 0)
    • status: string - Filter by status name (case-insensitive)
    • priority: string - Filter by priority. Allowed values: 'urgent', 'high', 'medium', 'low'.
    • parent_issue_id: string - Filter by parent issue ID (subissues of this issue)
    • search: string - Case-insensitive substring match against title and description
    • simple_id: string - Filter by issue simple ID (case-insensitive exact match)
    • assignee_user_id: string - Filter to issues assigned to this user ID
    • tag_id: string - Filter to issues having this tag ID
    • tag_name: string - Filter to issues having a tag with this name (case-insensitive)
```

stunningpixels:
## What changed
- added a canonical `SearchIssuesRequest` and moved rich issue listing to `POST /v1/issues/search`
- kept `GET /v1/issues` as the simple `project_id` list path, while routing all filtering, sorting, and pagination through the new JSON search endpoint
- updated the local server proxy, remote client, MCP `list_issues`, and generated remote TypeScript types to use the new shared search contract
- changed MCP issue listing to default to Kanban manual ordering and forward the full Kanban sort model to the API
- fixed the follow-up correctness issues discovered during rollout:
  - search and `simple_id` now escape `%`, `_`, and `\\` so matching stays literal
  - duplicate case-insensitive status/tag names remain inclusive instead of collapsing to the first match
  - `tag_id + tag_name` preserves intersection semantics

## Why
This addresses issue #3088. The MCP `list_issues` tool was fetching the full project issue list, filtering client-side, and then re-sorting by `updated_at desc`, which broke manual Kanban ordering.

The first pass pushed that logic into the API but still carried a lot of transport and compatibility complexity. The final shape in this branch makes the search contract explicit and body-based, so filtering, sorting, and pagination are implemented once in the remote API and MCP becomes a thinner adapter.

## Important implementation details
- manual sorting preserves Kanban semantics by ordering by status `sort_order`, then issue `sort_order`, then `issue_number`
- the shared response now includes `total_count`, `limit`, and `offset`, so pagination is API-driven instead of MCP-driven
- MCP still accepts friendly filters like status names and tag names, but resolves them to canonical API IDs before sending the JSON search request
- the local server simply proxies the JSON search request through to the remote service, and `shared/remote-types.ts` was regenerated to match
- this branch was trimmed back to the core feature and parity-critical fixes; generic MCP error-plumbing changes were intentionally left out

Closes #3088.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new issue search API and rewires multiple layers (DB queries, remote service routes, local proxy, MCP tool) to use it, which could affect pagination/sorting semantics and query performance. Changes also adjust response shapes (`total_count`, `limit`, `offset`), requiring careful compatibility validation across clients.
> 
> **Overview**
> Adds a canonical `SearchIssuesRequest` contract (with filter/sort/pagination fields) and expands `ListIssuesResponse` to include **API-driven pagination metadata** (`total_count`, `limit`, `offset`) plus new enums `IssueSortField` and `SortDirection`.
> 
> Moves rich issue listing into a new `POST /v1/issues/search` endpoint backed by a new `IssueRepository::search` SQL implementation (including LIKE escaping and flexible ORDER BY to preserve kanban/manual ordering), while keeping `GET /v1/issues` as a simple project list that now delegates to the same search path.
> 
> Updates the local server proxy, `RemoteClient`, MCP `list_issues` tool, and generated `shared/remote-types.ts` to use the new search endpoint, including MCP-side resolution of friendly status/tag names to IDs and support for explicit sort field/direction.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0dd3a68121c173896fe2e1988a5902ab71a85b36. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


crates/remote/src/db/issues.rs
```diff
@@ -33,6 +36,235 @@ enum IssueWorkflowSignal {
...
+              )
+              AND (
+                  $10::uuid[] IS NULL
+                  OR EXISTS (
+                      SELECT 1
+                      FROM issue_tags it
+                      WHERE it.issue_id = i.id AND it.tag_id = ANY($10)
+                  )
+              )
+            ORDER BY
```
LSRCT, line 199: The cases int this "order by" aren't compile time checked to be exhaustive, they assume `sort_field` is one of ["sort_order", "priority", "created_at", "updated_at", "title"] but`sort_field` is a string and could be set to a value not hardcoded here.

crates/mcp/src/task_server/tools/remote_issues.rs
```diff
@@ -64,6 +66,12 @@ struct McpListIssuesRequest {
     tag_id: Option<Uuid>,
     #[schemars(description = "Filter to issues having a tag with this name (case-insensitive)")]
     tag_name: Option<String>,
+    #[schemars(
+        description = "Field to sort by. Allowed values: 'sort_order', 'priority', 'created_at', 'updated_at', 'title'. Default: 'sort_order'."
+    )]
+    sort_field: Option<String>,
```
LSRCT, line 72: Could this be `Option<IssueSortField>`? We could then drop the `fn sort_field_key(sort_field: IssueSortField) -> &'static str` mapping in `db/issues.rs`

PR #3138, opened 2026-03-13
fix(auth): Add refresh token refresh grace window and degraded auth UI

abcpro1:
- Remote: align auth implementation with common practices.
  - allow a grace window for refresh token reuse.
  - on reuse, revoke the affected auth session only.
- UX: handle degraded auth gracefully.
  - avoid logging out the user on degraded auth state.
  - display a degraded auth message.
  - retry auth requests on degraded auth state.
  - recover from degraded auth state automatically.

---

> [!NOTE]
> **High Risk**
> Touches security-critical token rotation/refresh logic and adds new DB columns/migrations; mistakes could weaken session revocation or cause unexpected logouts/lockouts. Also changes client/server error handling paths for auth failures vs transient remote outages.
> 
> **Overview**
> Implements **refresh-token overlap** by storing the immediately previous refresh token on `auth_sessions` with a configurable grace expiry (`REFRESH_TOKEN_OVERLAP_SECS`), and updates token refresh to accept the previous token during that window while tightening reuse detection to **revoke only the affected session** (not all user sessions).
> 
> Introduces a **degraded remote auth** state end-to-end: `LoginStatus` can now be `LoggedIn` without a profile, the local backend persists a `remote_auth_degraded` slug in `AuthContext` and exposes it via `/api/info` and `/oauth/status`, and the web clients surface this as a sync error message while `tokenManager` pauses/resumes shapes, avoids forced logout on transient failures, and runs an exponential backoff recovery loop.
> 
> Cleans up related API/error semantics: token refresh expired error code becomes `expired_token`, remote-client error mapping treats specific auth error codes as definitive unauthorized, and the remote client removes the special long refresh timeout in favor of standard request timeouts plus degraded-state reporting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5b281c75951695175babe40f6deba1f1b2a21ad4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/server/src/routes/config.rs
+++ b/crates/server/src/routes/config.rs
@@ -130,6 +130,15 @@ async fn get_user_system_info(
                             .await;
                     }
 
+                    deployment
+                        .track_if_analytics_allowed(
+                            "login_status_timeout",
+                            serde_json::json!({
+                                "has_cached_profile": cached_profile.is_some(),
+                            }),
+                        )
+                        .await;
+
                     LoginStatus::LoggedIn {
                         profile: cached_profile,
                     }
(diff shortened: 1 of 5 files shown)


crates/server/src/routes/config.rs
```diff
@@ -105,18 +107,48 @@ async fn get_user_system_info(
...
+    let login_status = match tokio::time::timeout(
         std::time::Duration::from_secs(2),
         deployment.get_login_status(),
     )
     .await
-    .unwrap_or(LoginStatus::LoggedOut);
+    {
+        Ok(status) => status,
+        Err(_) => {
+            tracing::warn!("timed out determining login status for /api/info");
```
ggordonhall, line 118: Perhaps worth adding an audit event here?

PR #3155, opened 2026-03-16
fix: use ~/Library/Caches on macOS instead of temp dir for worktrees

MinbinLuo:
## Summary

On macOS, `std::env::temp_dir()` returns `/var/folders/.../T/`, which is subject to periodic cleanup by the OS. macOS silently purges small files from this directory, which destroys the `.git` symlink files inside git worktrees (they're typically only ~70 bytes). This causes:

- Persistent ERROR log spam: `could not find repository at '...'`
- Diff streams failing and reconnecting in a loop
- Workspace code becoming inaccessible until the `.git` files are manually recreated

The previous comment claimed `/var/folders/...` was "persistent storage" — this is incorrect.

### Fix

Use `dirs::cache_dir()` (`~/Library/Caches/`) on macOS instead of the temp dir. This directory is not subject to automatic system cleanup. The `dirs` crate is already a project dependency, and the change falls back to the old behavior if `cache_dir()` returns `None`.

Existing worktrees are unaffected — their paths are stored in the database `container_ref` field and continue to work at their current locations. Only newly created workspaces use the new path.

## Test plan

- [x] On macOS, create a new workspace and verify the worktree is created under `~/Library/Caches/vibe-kanban/worktrees/` (or `vibe-kanban-dev` in debug builds)
- [x] Verify existing workspaces with worktrees in the old temp dir continue to function
- [x] `cargo test --workspace` passes
- [x] `pnpm run check` passes

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Low Risk**
> Low risk change limited to macOS path selection for newly created temp/worktree directories, with a fallback to the previous temp-dir behavior if the cache directory is unavailable.
> 
> **Overview**
> On macOS, `get_vibe_kanban_temp_dir()` now prefers `dirs::cache_dir()` (typically `~/Library/Caches`) instead of `std::env::temp_dir()` to avoid OS purges that can break git worktree `.git` symlinks.
> 
> Non-macOS behavior is unchanged, and macOS falls back to the prior temp-dir location if no cache directory is found.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 74e1dca91c8756f7d044ae34b3a987b19b839336. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/utils/src/path.rs
+++ b/crates/utils/src/path.rs
@@ -113,7 +113,12 @@ pub fn get_vibe_kanban_temp_dir() -> std::path::PathBuf {
     };
 
     if cfg!(target_os = "macos") {
-        // macOS already uses /var/folders/... which is persistent storage
+        // Use ~/Library/Caches/ instead of the temp dir (/var/folders/.../T/)
+        // because macOS periodically purges small files from the temp dir,
+        // which destroys the .git symlink files inside worktrees.
+        if let Some(cache_dir) = dirs::cache_dir() {
+            return cache_dir.join(dir_name);
+        }
         std::env::temp_dir().join(dir_name)
     } else if cfg!(target_os = "linux") {
         // Linux: use /var/tmp instead of /tmp to avoid RAM usage


Juanlucasbg:
## Aggressive review summary — PR #3155

Ran the four-reviewer battery against this PR (structural, security, adversarial/codex, conventions). The fix solves a real bug, but **three independent HIGH-severity findings** block merge as-is.

### HIGH — blocks merge

**H1. Stranded worktrees on upgrade, no migration path**
`WorktreeManager::get_default_worktree_base_dir()` now returns `~/Library/Caches/vibe-kanban/worktrees`. Existing macOS users have active workspaces with `container_ref` rows in the DB pointing at `/var/folders/.../T/vibe-kanban/worktrees/<id>`. After upgrade:
- Active workspaces keep working *until* macOS purges the small `.git` symlink files (the very bug this PR fixes), at which point those worktrees break and there is no auto-recreate-at-new-path code.
- `cleanup_orphan_workspaces` (`crates/workspace-manager/src/workspace_manager.rs:546`) only sweeps the new default base dir; orphans in `/var/folders/.../T/vibe-kanban/` are never cleaned.
- `qa-repos` (full git clones, often hundreds of MB) at `/var/folders/.../T/vibe-kanban/qa-repos/` are also never cleaned.

Fix: either migrate (move) old base-dir contents to the new base, or extend the orphan-cleanup loop to also sweep the legacy path on macOS, or document upgrade as requiring a manual `rm -rf /var/folders/.../T/vibe-kanban`.

**H2. Filesystem skip-filter bypass for legacy path**
`crates/services/src/services/filesystem.rs:216` builds the directory-walker skip prefix from `get_vibe_kanban_temp_dir()` — i.e. only the *new* path. After upgrade, walks rooted at `/`, `/var`, or any ancestor of the legacy temp dir will no longer skip `/var/folders/.../T/vibe-kanban/worktrees/`. The walker enumerates the surviving files there as "git repos" in the picker UI, and recurses into thousands of orphan worktree directories. Concrete UX regression and perf hit.

Fix: skip both `get_vibe_kanban_temp_dir()` and the legacy macOS path on macOS until orphans are migrated.

**H3. Cache root inconsistency with the rest of the codebase**
`crates/utils/src/lib.rs:50` already exposes `utils::cache_dir()` which returns `~/Library/Caches/ai.bloop.vibe-kanban/` via `ProjectDirs::from("ai", "bloop", "vibe-kanban")` — the Apple-convention reverse-DNS bundle ID. The PR introduces a *second* cache root at `~/Library/Caches/vibe-kanban/` (no bundle scoping), creating two cleanup surfaces and violating the convention used for images/sounds.

Fix: use `utils::cache_dir().join("worktrees")` (or extract the `ProjectDirs` logic) so all cached data lives under one bundle-scoped root.

### MED

- `dirs::cache_dir()` returning `None` (e.g. launchd / `env -i`) silently falls through to `std::env::temp_dir()` — the buggy path. Either `expect(...)` (consistent with `utils::cache_dir()`'s pattern) or log a loud warning. (`crates/utils/src/path.rs:119-122`)
- `qa_repos.rs:20-27` clones repos once and never refreshes; `clone_qa_repos_if_needed` short-circuits on `.git` presence. Under the old temp dir, OS purge effectively forced re-clones every few weeks. Moving to a non-purged location means stale clones forever. Consider an mtime-based staleness check or `git fetch` on startup.
- Cross-volume `EXDEV` on `git worktree move` if `~` is on an external APFS volume and the legacy worktree was on internal storage.
- No unit test for `get_vibe_kanban_temp_dir()` macOS branch — `CLAUDE.md` says "Add tests for new logic and edge cases."

### LOW

- Spotlight indexes `~/Library/Caches/` by default; full repo checkouts including any non-`.gitignored` files become indexable. Recommendation: write `.metadata_never_index` at the cache root on first creation.
- `crates/utils/src/path.rs:166` test uses literal `/var/folders/.../T/vibe-kanban-dev/` — comment now misleading.
- `crates/utils/src/port_file.rs:21` still uses `env::temp_dir().join("vibe-kanban")` — out of scope but the same anti-pattern.

### Verdict

**Do not merge as-is.** The fix is correct for the originally-reported bug, but introduces an unmigrated upgrade regression. Smallest acceptable change: (a) keep the new path, (b) extend `cleanup_orphan_workspaces` and the `filesystem.rs` skip filter to also cover the legacy `/var/folders/.../T/vibe-kanban/` path on macOS, (c) replace `dirs::cache_dir()` with `utils::cache_dir()` for convention consistency.

— Reviewed by automated four-tool battery (structural / security / adversarial / conventions). Rerun the review by re-requesting after a rebase + the fixes above.

PR #3162, opened 2026-03-16
feat: system notifications via Tauri

ggordonhall:
Send system notifications on Kanban board events using the Tauri invoke command.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new cross-boundary (web -> Tauri) notification path and hooks it into the app’s notification stream, which could affect notification volume/duplication and desktop behavior across platforms.
> 
> **Overview**
> Adds a new Tauri `show_system_notification` command and registers it in the desktop builder, creating a `NotificationService` from the on-disk config to display OS notifications (and any enabled notification side-effects).
> 
> On the local web app, introduces `AppSystemNotifications` to watch grouped notifications and emit a system notification for newly-unseen groups (deduped per group id and reset on user change) via a new `showSystemNotification()` helper that `invoke`s the Tauri command.
> 
> Extends `notificationMessage` with `getGroupedNotificationText()` to render grouped notification segments into a plain string, resolving `user` segments to member display names when available (fallback to "Someone").
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f813106c19cb9d8b6bfe266ea91ae31add33197c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/tauri-app/src/main.rs
+++ b/crates/tauri-app/src/main.rs
@@ -8,7 +8,9 @@ use services::services::{
     config::load_config_from_file,
     notification::{NotificationService, PushNotifier, set_global_push_notifier},
 };
-use tauri::{Emitter, Listener, Manager};
+#[cfg(target_os = "macos")]
+use tauri::Manager;
+use tauri::{Emitter, Listener};
 use tauri_plugin_notification::NotificationExt;
 use tauri_plugin_opener::OpenerExt;
 use tauri_plugin_updater::UpdaterExt;


packages/local-web/src/app/notifications/showSystemNotification.ts
```diff
@@ -0,0 +1,28 @@
...
+): Promise<void> {
+  if (!isTauriApp()) {
+    return;
+  }
+
+  try {
+    await invoke('show_system_notification', {
+      title: notification.title,
+      body: notification.body,
+    });
```
anastasiya1155: if FE is invoking notifications, this means that when window is closed, even if VK runs in the background, there will be no notifications, they will appear only when window is open again. Is this ok for us?

PR #3156, opened 2026-03-16
fix(remote-web): add pkce fallback for insecure dev environments

BeyondWUXF:
## Summary
This PR addresses an issue where the OAuth PKCE flow fails in development environments that are not served over HTTPS (e.g., accessing via a local IP address like `http://192.168.x.x:3000`). In these insecure contexts, browsers disable the `crypto.subtle` API.

## Changes
- Added a check for `crypto.subtle` availability.
- Implemented a fallback hashing mechanism using a simple string hash when `crypto.subtle` is missing.
- **Security**: The fallback is strictly gated behind `import.meta.env.DEV || import.meta.env.MODE === 'development'` to ensuring it never runs in production builds.

## Test Plan
1. Start the remote dev server (`pnpm run remote:dev`).
2. Access the frontend via local IP (e.g., `http://192.168.1.5:3000`) instead of `localhost`.
3. Attempt to log in via GitHub/Google.
4. Verify that the login flow proceeds without crashing due to "crypto.subtle is undefined".

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth PKCE challenge generation used during login, so mistakes could break authentication flows. Risk is mitigated by only enabling the `crypto-js` fallback in `import.meta.env.DEV` and failing fast otherwise.
> 
> **Overview**
> Fixes OAuth PKCE failures in insecure (non-HTTPS) development environments by making `generateChallenge` handle missing `crypto.subtle`.
> 
> When `crypto.subtle` is unavailable, the code now *only in dev* computes the SHA-256 challenge via `crypto-js` (with a warning); in non-dev builds it throws a clear error instead of silently proceeding.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d3b152cc1a455057323d26c4d321ed5ef0df6e3e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3189, opened 2026-03-18
fix(remote-web): add pkce fallback for insecure dev environments

BeyondWUXF:
## Summary
This PR addresses an issue where the OAuth PKCE flow fails in development environments that are not served over HTTPS (e.g., accessing via a local IP address like \`http://192.168.x.x:3000\`). In these insecure contexts, browsers disable the \`crypto.subtle\` API.

It implements the feedback received in #3156 to safely and accurately handle this scenario without compromising production security or performance.

## Changes
* Added a check for \`crypto.subtle\` availability.
* Added a fallback using \`crypto-js\`'s SHA256 to ensure the hash accurately matches the server's expected challenge algorithm.
* Used dynamic importing (\`await import('crypto-js')\`) to prevent \`crypto-js\` from bloating the production bundle.
* **Security Gate**: The fallback is strictly gated behind \`import.meta.env.DEV\` to ensure it NEVER runs in production builds. If \`crypto.subtle\` is missing in a non-dev environment, it will fail fast by throwing an explicit error.

## Test Plan
1. Start the remote dev server (\`pnpm run remote:dev\`).
2. Access the frontend via local IP (e.g., \`http://192.168.1.5:3000\`) instead of \`localhost\`.
3. Attempt to log in via GitHub/Google.
4. Verify that the login flow proceeds smoothly without crashing due to "crypto.subtle is undefined" and without backend verification errors.

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth PKCE challenge generation, so a mismatch could break login flows; risk is mitigated by gating the fallback to `import.meta.env.DEV` and failing fast otherwise.
> 
> **Overview**
> Improves remote-web OAuth PKCE support in insecure (HTTP) *development* environments by making `generateChallenge` fall back when `crypto.subtle` is unavailable.
> 
> In dev-only mode it dynamically imports `crypto-js` to compute a SHA-256 challenge (with a warning), while production continues to require `crypto.subtle` and now throws an explicit error if it’s missing. Adds `crypto-js` and its types to `remote-web` dependencies (and lockfile).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ce2289267f7d71914c1200db18cb161d3f1176ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Juanlucasbg:
## Rebase verification — PR #3189

- `pnpm-lock.yaml`

Lockfile-only conflict. Mechanical:

```bash
git checkout main -- pnpm-lock.yaml
pnpm install   # regenerate
git add pnpm-lock.yaml
git rebase --continue
```

Then push. The PKCE-fallback fix itself (40 LOC) doesn't conflict with anything on `main`; only the lockfile drifted.

— Automated rebase-verification: confirmed lockfile-only conflict. Mechanical to resolve.

PR #3197, opened 2026-03-18
fix: prevent session executor poisoning when switching sessions

### LINKED ISSUE #3196: Bug: switching sessions in workspace permanently poisons executor display via scratch auto-save
## Bug description

When a workspace has multiple sessions with different executors (e.g., Claude Code + Codex), switching between sessions causes the executor selector to **permanently** display the wrong executor for a session — even after browser reload.

### 1. `latestConfig` fallback uses `sessions[0]` instead of the selected session

**File:** `packages/web-core/src/features/workspace-chat/ui/SessionChatBoxContainer.tsx:333`

```typescript
// Fallback: just executor from session metadata
const lastSessionExecutor = sessions?.[0]?.executor;
```

`sessions[0]` is the **most recently used** session (sorted by latest execution timestamp), NOT the currently selected session. When viewing Session A (Claude Code) after running something in Session B (Codex), `sessions[0]` = Session B → fallback returns `CODEX` instead of `CLAUDE_CODE`.

This fires during the brief window while the WebSocket for execution processes is connecting (`processes = []` → `getLatestConfigFromProcesses` returns null → falls through to this fallback).

### 2. Draft auto-save persists the wrong executor to scratch

When the message editor initializes with the wrong `latestConfig` executor, the debounced `saveToScratch` persists `executor_config: { executor: "CODEX" }` to the scratch record keyed by the **Claude Code session's ID**.

### 3. Scratch config takes priority over process-derived config on reload

**File:** `packages/web-core/src/shared/hooks/useExecutorConfig.ts:42-49`

```typescript
const effective = useMemo(
  () =>
    userSelections.executor ??
    scratchConfig?.executor ??       // ← poisoned scratch wins
    lastUsedConfig?.executor ??      // ← correct value never reached
    configExecutorProfile?.executor ??
    options[0] ??
    null,
  ...
);
```

Once the scratch is poisoned, every subsequent load of Session A shows the wrong executor — even though the session's DB record AND all its execution processes have the correct executor.

### Secondary issue: session selection override on re-fetch

**File:** `packages/web-core/src/shared/hooks/useWorkspaceSessions.ts:60-62`

```typescript
setSelection((prev) => {
  if (prev?.mode === 'new' && !workspaceChanged) return prev;
  return { mode: 'existing', sessionId: sessions[0].id };  // always overrides
});
```

This effect runs every time `sessions` re-fetches via React Query. It only preserves `mode: 'new'` selections — explicit existing-session selections get overridden back to `sessions[0]` on every refetch, making it hard to stay on a non-most-recent session.

## Steps to reproduce

1. Create a workspace with a Claude Code session, run something in it
2. Create a second session with a different executor (e.g., Codex), run something in it
3. Switch back to the first session (Claude Code)
4. Observe: executor icon briefly shows correctly, then flips to Codex
5. Reload page — still shows Codex for the Claude Code session

## Confirmed via DB inspection

- Session record: `executor = CLAUDE_CODE` ✓
- All `execution_processes` for that session: `executor_config.executor = CLAUDE_CODE` ✓
- `scratch` table for that session: `executor_config: { executor: "CODEX" }` ✗ — poisoned by the fallback bug

## Proposed fix

**Fix 1 — `SessionChatBoxContainer.tsx:333`**: Use the selected session's executor in existing-session mode:
```typescript
const lastSessionExecutor = isNewSessionMode
  ? sessions?.[0]?.executor
  : session?.executor;
```

**Fix 2 — `useWorkspaceSessions.ts:60-62`**: Preserve explicit user selections on re-fetch:
```typescript
setSelection((prev) => {
  if (prev?.mode === 'new' && !workspaceChanged) return prev;
  if (
    prev?.mode === 'existing' &&
    !workspaceChanged &&
    sessions.some((s) => s.id === prev.sessionId)
  ) {
    return prev;
  }
  return { mode: 'existing', sessionId: sessions[0].id };
});
```

## Environment

- VK v0.1.32
- Observed on self-hosted Coder workspace

ciekawy:
## Summary

- Fix `latestConfig` fallback using `sessions[0]` (most-recently-used) instead of the selected session's executor — caused wrong executor to display during WebSocket loading window, which then got persisted to scratch
- Preserve explicit session selections across React Query refetches instead of always snapping back to `sessions[0]`

## Root cause

Three-step bug chain: wrong fallback → draft auto-save poisons scratch → scratch overrides correct value on every reload. See #3196 for full analysis.

## Test plan

- [ ] Create workspace with two sessions using different executors (e.g., Claude Code + Codex)
- [ ] Run something in both sessions
- [ ] Switch between sessions — verify executor icon stays correct for each
- [ ] Reload browser — verify executor persists correctly
- [ ] Enter "new session" mode — verify it still defaults to most-recently-used executor

Fixes #3196

---

> [!NOTE]
> **Medium Risk**
> Changes session-selection and executor fallback behavior in the chat UI, which can affect which executor/config gets shown and persisted; limited scope but user-facing state logic can be regression-prone.
> 
> **Overview**
> Prevents the chat UI from briefly falling back to `sessions[0].executor` when viewing an existing session, avoiding showing (and persisting to scratch) the wrong executor during the WebSocket/loading window.
> 
> Updates `useWorkspaceSessions` to preserve a user’s explicit existing-session selection across React Query refetches (only auto-selecting `sessions[0]` when the workspace changes or the previously-selected session no longer exists). Also replaces several `Duration::from_mins(...)` constants with equivalent `Duration::from_secs(...)` expressions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 34de52f306511fabc72fbb2fc20808b3eac3d447. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/web-core/src/features/workspace-chat/ui/SessionChatBoxContainer.tsx
+++ b/packages/web-core/src/features/workspace-chat/ui/SessionChatBoxContainer.tsx
@@ -346,8 +346,13 @@ export function SessionChatBoxContainer(props: SessionChatBoxContainerProps) {
     const fromLastSession = getLatestConfigFromProcesses(lastSessionProcesses);
     if (fromLastSession) return fromLastSession;
 
-    // Fallback: just executor from session metadata
-    const lastSessionExecutor = sessions?.[0]?.executor;
+    // Fallback: executor from session metadata.
+    // In new-session mode, default to the most-recently-used session's executor.
+    // In existing-session mode, use the selected session's own executor to avoid
+    // showing a different session's executor during the WebSocket loading window.
+    const lastSessionExecutor = isNewSessionMode
+      ? sessions?.[0]?.executor
+      : session?.executor;
     if (lastSessionExecutor) {
       return {
         executor: lastSessionExecutor as BaseCodingAgent,
@@ -355,7 +360,7 @@ export function SessionChatBoxContainer(props: SessionChatBoxContainerProps) {
     }
 
     return null;
-  }, [processes, lastSessionProcesses, sessions]);
+  }, [processes, lastSessionProcesses, sessions, isNewSessionMode, session]);
 
   const needsExecutorSelection =
     isNewSessionMode || (!session?.executor && !latestConfig?.executor);
--- a/packages/web-core/src/shared/hooks/useWorkspaceSessions.ts
+++ b/packages/web-core/src/shared/hooks/useWorkspaceSessions.ts
@@ -59,6 +59,16 @@ export function useWorkspaceSessions(
       // Only preserve new session mode within the same workspace
       setSelection((prev) => {
         if (prev?.mode === 'new' && !workspaceChanged) return prev;
+        // Preserve explicit existing-session selection if the session still
+        // exists — avoids snapping back to sessions[0] on every React Query
+        // refetch, which would override the user's manual session switch.
+        if (
+          prev?.mode === 'existing' &&
+          !workspaceChanged &&
+          sessions.some((s) => s.id === prev.sessionId)
+        ) {
+          return prev;
+        }
         return { mode: 'existing', sessionId: sessions[0].id };
       });
     } else {


Juanlucasbg:
## Rebase verification — PR #3197

- `crates/services/src/services/remote_client.rs`

`remote_client.rs` is part of the actively-evolving services layer (multiple recent merges). The "prevent session executor poisoning when switching sessions" change targets shared mutable state that has likely been refactored.

**Action**: rebase, then verify the session-switching path still has the original poisoning surface. If the executor lifecycle was reworked (e.g. via the scratch / persistence refactors), the original bug may already be fixed elsewhere.

PR #3202, opened 2026-03-18
fix: load non-localhost preview URLs directly in iframe

ciekawy:
## Summary

- When the preview URL is not a loopback address (e.g. Coder port-forwarded URLs like `https://4000--workspace.coder.example.com`), load it directly in the iframe instead of routing through the subdomain-based preview proxy
- The subdomain proxy (`http://{port}.localhost:{proxyPort}`) only works when VK runs on localhost — this breaks preview for remote-hosted VK (Coder, cloud workspaces, tunneled environments)
- Direct mode is **intentionally degraded**: no bridge communication, no injected devtools/Eruda/inspect scripts, no SPA navigation via postMessage — URL bar changes trigger a full iframe reload

## Problem

When VK is hosted remotely (e.g. on Coder), the current subdomain-based proxy routing doesn't work:
1. Dev server starts on `localhost:4000` (on the server)
2. VK rewrites the iframe URL to `http://4000.localhost:{proxyPort}` 
3. The browser can't resolve `4000.localhost` from outside → preview is broken

Users can access their dev server via Coder's port-forwarding (`https://4000--workspace.coder.example.com`), but VK still forces the URL through the proxy.

## Changes

- **`LOOPBACK_HOSTS` constant** — shared set matching `usePreviewUrl.ts` and Rust `is_loopback_redirect_host` (includes IPv6 `[::1]`, `[::]`)
- **`isLoopbackPreview` memo** — determines routing mode, reused across the component
- **`iframeUrl`** — non-loopback URLs loaded directly (no proxy needed)
- **`displayedPreviewUrl`** — strips `_refresh` param for direct navigation URLs
- **`handleUrlSubmit`** — bridge SPA navigation gated on `isLoopbackPreview` (devtools script only injected by proxy); non-loopback falls through to `setOverrideUrl` for full iframe reload
- **Eruda/inspect mode** — disabled in direct mode (no injected scripts to handle commands)

## Test plan

- [ ] Local dev: start a dev server, verify preview iframe loads via proxy as before
- [ ] Remote/Coder: set an override URL to a non-localhost address, verify iframe loads it directly
- [ ] URL bar: submit a new path on a non-localhost preview, verify full iframe reload (not silent no-op)
- [ ] Eruda/inspect buttons: verify disabled/no-op in non-localhost mode, functional in localhost mode
- [ ] Refresh: verify refresh button works for both localhost and non-localhost preview URLs
- [ ] IPv6: verify `http://[::1]:3000` is treated as loopback (uses proxy)

### DIFF
--- a/packages/web-core/src/pages/workspaces/PreviewBrowserContainer.tsx
+++ b/packages/web-core/src/pages/workspaces/PreviewBrowserContainer.tsx
@@ -33,6 +33,23 @@ import type { PreviewDevToolsMessage } from '@/shared/types/previewDevTools';
 const MIN_RESPONSIVE_WIDTH = 320;
 const MIN_RESPONSIVE_HEIGHT = 480;
 
+const noop = () => {};
+
+/**
+ * Hostnames that are considered loopback/local.
+ * Kept in sync with `LOOPBACK_HOSTS` in usePreviewUrl.ts and
+ * `is_loopback_redirect_host` in the Rust preview-proxy crate.
+ */
+const LOOPBACK_HOSTS = new Set([
+  'localhost',
+  '127.0.0.1',
+  '0.0.0.0',
+  '[::1]',
+  '::1',
+  '[::]',
+  '::',
+]);
+
 function parsePreviewUrl(rawUrl: string, baseUrl?: string): URL | null {
   const trimmed = rawUrl.trim();
   if (!trimmed) return null;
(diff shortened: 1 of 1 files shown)


vdmkotai:
Thanks for the PR! The approach of bypassing the subdomain proxy for non-loopback URLs makes sense - clean way to unblock Coder/tunneled environments without touching the backend.

A few things to address before merging:

1. **Missing `[::]` in loopback check** - `usePreviewUrl`'s `LOOPBACK_HOSTS` includes `[::]` (servers can print `http://[::]:3000`), but the new `isLoopbackPreview` list doesn't. This would cause those URLs to bypass the proxy unintentionally. Ideally extract a shared helper so the lists don't drift.

2. **Eruda & inspect mode controls stay enabled in direct mode** - Since `devtools_script.js` isn't injected without the proxy, the Eruda toggle and inspect mode buttons silently do nothing. They should be disabled when `!isLoopbackPreview`.

3. **PR description says the URL bar tracks in-iframe navigation for direct URLs** - but without the injected devtools script, `navigation` stays null, so the URL bar won't actually update on SPA navigation. Worth correcting the description to set expectations.

4. **Small comment in code** - A brief note near the `isLoopbackPreview` branch explaining that direct mode is intentionally degraded (no bridge, no script injection, no header stripping) would help future maintainers understand the tradeoff.

All small changes - the core logic is solid and localhost flow is unaffected.

PR #3221, opened 2026-03-20
Add pre-commit hook for Rust and frontend quality checks

yonch:
Adds a pre-commit hook that:
- detects changed Rust crates (with transitive reverse deps via scripts/affected-crates.py) and runs cargo fmt --check, clippy, and tests. 
- runs prettier/tsc/eslint on affected frontend packages.

Only runs tests/checks on changed code for the commit, to speed up checks -- don't want to check everything.

Should reduce the back-and-forth of contributors with the Github Actions checks...

---

> [!NOTE]
> **Low Risk**
> Adds developer tooling and documentation only; no runtime behavior changes, but the hook could slow or block local commits if mis-scoped or if required tools aren’t installed.
> 
> **Overview**
> Adds an opt-in `.githooks/pre-commit` hook that inspects staged files and runs **scoped quality checks**: Rust `cargo fmt --check` plus `clippy`/`test` limited to affected workspace crates (or `--workspace` when global Rust config files change), and frontend `prettier`/`tsc`/`eslint` limited to affected `pnpm` packages.
> 
> Introduces `scripts/affected-crates.py`, which uses `cargo metadata` to compute impacted crates (including reverse-dependents) from changed paths, and updates the README with setup instructions and tool prerequisites for enabling the hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac117d77cedba86410f7b6612e0a8895dcc16ac1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -7,8 +7,8 @@ set -euo pipefail
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
-# Staged files (added, copied, modified).
-STAGED=$(git diff --cached --name-only --diff-filter=ACM)
+# Staged files (added, copied, modified, renamed).
+STAGED=$(git diff --cached --name-only --diff-filter=ACMR)
 
 has_staged() {
   echo "$STAGED" | grep -qE "$1"
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+# pre-commit hook: run quality checks gated on which files are staged.
+#
+# Enable for all checkouts (including worktrees) with:
+#   git config core.hooksPath .githooks
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+
+# Staged files (added, copied, modified).
+STAGED=$(git diff --cached --name-only --diff-filter=ACM)
+
+has_staged() {
+  echo "$STAGED" | grep -qE "$1"
+}
+
+FAILED=0
+
+# --- Rust: fmt, clippy, test (scoped to affected crates) ---
+# affected-crates.py is the single source of truth for whether Rust checks
+# are needed and which crates to check.  It handles .rs, .toml, Cargo.lock,
+# rust-toolchain.toml, etc.
+if command -v cargo &>/dev/null && command -v python3 &>/dev/null; then
+  # Capture exit code without tripping set -e.
+  SCOPE_EXIT=0
+  SCOPE=$(echo "$STAGED" | python3 "$REPO_ROOT/scripts/affected-crates.py" --stdin) || SCOPE_EXIT=$?
+
+  if [ "$SCOPE_EXIT" -eq 2 ]; then
+    # Global file changed (Cargo.lock, rust-toolchain, etc.) — full workspace.
+    echo "pre-commit: global Rust file changed, checking full workspace..."
+    SCOPE="--workspace"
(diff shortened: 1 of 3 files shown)


Juanlucasbg:
## Aggressive review summary — PR #3221

267-line pre-commit hook with Rust + frontend quality checks scoped to staged files. New files: `.githooks/pre-commit`, `scripts/affected-crates.py`, README updates. Verdict: **clean — recommend merge**.

### Why the change is well-shaped

The hook's design choices are exactly the ones that distinguish a tolerable pre-commit from a frustrating one:

- **Affected-files scoping**: `affected-crates.py` is the single source of truth for "which Rust crates need checking based on staged files". Cargo.lock / rust-toolchain.toml changes correctly escalate to `--workspace`. Frontend packages similarly filter via `pnpm --filter @vibe/$pkg`.
- **Bash 3.2 compatibility**: Uses `case`-based functions instead of associative arrays so it runs on macOS's stock `/bin/bash`. This is the kind of detail that gets missed and breaks every macOS contributor.
- **`cargo fmt` always workspace-wide**: Format is fast (no compile) and per-crate gating would mask issues in unstaged code. Clippy + test stay scoped.
- **Run-all-then-report**: `FAILED=1` accumulator pattern means a developer sees every issue at once, not one-at-a-time.
- **Three targeted scans**: legacy-frontend-paths, unused-i18n-keys checks fire only when matching files are staged. Cheap when irrelevant.
- **Opt-in via `git config core.hooksPath .githooks`**: doesn't surprise anyone who clones the repo.

### Findings

- **Structural — PASS**: Single-file bash + Python scope analyzer. No new global tooling deps beyond what CI already uses.
- **Adversarial — LOW**:
  - **No bypass guidance**: `--no-verify` always works, but a developer might not know it. A line in the README ("If a check is broken, use `git commit --no-verify` and ping maintainers") would be friendly.
  - **Affected-crates.py edge cases**: Renaming a file (`R` in `--diff-filter`) — does the script see both old and new path? If yes, both crates compile-check. If no, the new crate might be missed. Worth a smoke test.
  - **`set -euo pipefail` + grep that returns 1 on no-match**: `has_staged` uses `grep -q` which exits 1 when nothing matches. Inside an `if has_staged ...; then` block, the conditional swallows that, but not in a context where the result is captured. Looks correct in this script.
- **Performance**: Worst case is "developer touched Cargo.lock" → workspace clippy + test on a large Rust workspace. That can be 1-3 minutes locally on a warm cache. Acceptable for the rare global-change commit, painful if it triggers on every commit. Verify `affected-crates.py` doesn't escalate too aggressively — e.g. a stylistic change to `Cargo.toml` shouldn't trigger workspace mode.
- **Conventions — PASS**: README update tells developers how to enable. Correct path.

### NITs

- **`pnpm --filter @vibe/$pkg run format:check` failures bunch up**: when the formatter fails, the hook continues and runs `tsc --noEmit` against unformatted code, which can produce confusingly cascading errors. Consider stopping the per-package chain on format failure (still continue to the *next* package).
- **No frontend test invocation**: the hook runs `format:check`, `check`, and `lint` for frontend, but not unit tests. If the project has Vitest tests, consider adding them to the hook (or document why they're CI-only).
- **`scripts/affected-crates.py` not visible in this excerpt** — the maintainer should make sure exit code 10 (the "global file changed" sentinel) is documented in a comment at the top of the script.

### Verdict

Approve.

— Reviewed by automated single-pass review (developer-tooling addition; full 4-tool battery skipped — change is well-scoped tooling that operators opt into).

PR #3228, opened 2026-03-22
fix(ws): reconnect after mobile tab suspension in useJsonPatchWsStream

### LINKED ISSUE #3227: fix(ws): session stuck 'running' and messages lost on mobile due to WebSocket not reconnecting after tab suspension
## Summary

On mobile browsers, two related symptoms occur during a conversation session:

1. **Session stuck "waiting/running"** — after the agent finishes its turn, the UI remains in a running/waiting state. Switching to another session/workspace and back fixes it.
2. **Follow-up message "disappears"** — after sending a message, the conversation appears unchanged (as if the agent never received it). Same switch-away-and-back workaround makes everything appear and the agent has already started responding.

Both symptoms were introduced at some point around 0.1.30–0.1.33 and reproduce reliably on mobile browsers (tested on iOS Safari and Chrome for Android).

## Root Cause

**File:** `packages/web-core/src/shared/hooks/useJsonPatchWsStream.ts`

The `onclose` handler contained:

```js
if (cancelled || finishedRef.current || (evt?.code === 1000 && evt?.wasClean)) {
  return; // don't reconnect
}
```

Mobile browsers close WebSocket connections with **code `1000` / `wasClean = true`** when a tab is backgrounded or suspended — the same values as a server-initiated clean close. The client incorrectly treated this as "stream permanently ended" and skipped reconnection.

`finishedRef.current` already correctly guards the legitimate "server sent `{finished:true}`, stream is done forever" case, making the `wasClean` condition redundant **and harmful**.

The switch-away-and-back workaround forces a remount of `ExecutionProcessesProvider`, which opens a fresh WebSocket and receives the up-to-date snapshot (including completed process status and any new follow-up execution processes).

**Symptom 1 — stuck running:** The execution-process `status: running → success` patch never arrived after the WS was closed without reconnect.

**Symptom 2 — message disappears:** The user's follow-up is sent via HTTP POST (which succeeds), but the resulting new execution process is only delivered via WS patch. With the dead stream, the client never learns about it.

## Fix

Two changes in `useJsonPatchWsStream.ts`:

1. **Remove `(evt?.code === 1000 && evt?.wasClean)`** from the reconnect guard — `finishedRef.current` is sufficient.
2. **Add a `visibilitychange` listener** — mobile browsers also throttle/freeze `setTimeout` in the background, so the exponential-backoff retry timer may never fire. The visibility trigger reconnects immediately (with backoff reset to 0) when the user returns to the tab.

## Affected versions

Reproducible on **0.1.33** (and likely 0.1.30–0.1.32). The fix applies to the current `main`.

ciekawy:
Fixes #3227

## Summary

- Remove the erroneous `(evt?.code === 1000 && evt?.wasClean)` reconnect guard from `useJsonPatchWsStream` — mobile browsers close WebSocket connections with code 1000/wasClean=true when backgrounding a tab, which is not a permanent end-of-stream
- Add a `visibilitychange` listener to reconnect immediately when the user returns to the tab (mobile browsers also throttle/freeze `setTimeout` in the background, so the backoff timer alone is insufficient)

## Test plan

- [ ] On a mobile browser (iOS Safari / Android Chrome), open a workspace session and start an agent turn
- [ ] Switch to another app/tab while the agent is running, wait for it to finish, return → session should show completed state (not stuck "running")
- [ ] Open a workspace session, switch away briefly, return, send a follow-up message → message should remain visible and agent should respond (not disappear)
- [ ] On desktop: verify normal WebSocket streams still work (no regressions to reconnect behavior on non-mobile)
- [ ] Verify that a legitimately finished stream (server sends `{finished:true}`) still does not attempt to reconnect after completion

---

> [!NOTE]
> **Medium Risk**
> Changes WebSocket reconnection behavior for all consumers of `useJsonPatchWsStream`, which could cause extra reconnect attempts or subtle lifecycle issues if the visibility/close logic misfires.
> 
> **Overview**
> Improves resiliency of `useJsonPatchWsStream` on mobile by **reconnecting after tab suspension/backgrounding**.
> 
> Removes the guard that skipped reconnects on clean `1000` WebSocket closes, and adds a `visibilitychange` listener that cancels any pending backoff and forces an immediate reconnect when the page becomes visible again (unless the stream is marked `finished`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ca5c00adb7bdc6a5f324057b055c2a64d590679. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/web-core/src/shared/hooks/useJsonPatchWsStream.ts
+++ b/packages/web-core/src/shared/hooks/useJsonPatchWsStream.ts
@@ -170,20 +170,21 @@ export const useJsonPatchWsStream = <T extends object>(
             // that was already received.
           };
 
-          ws.onclose = (evt) => {
+          ws.onclose = () => {
             setIsConnected(false);
             wsRef.current = null;
 
-            // Do not reconnect if we received a finished message or clean close
-            if (
-              cancelled ||
-              finishedRef.current ||
-              (evt?.code === 1000 && evt?.wasClean)
-            ) {
+            // Do not reconnect if we explicitly cancelled or the server sent a
+            // finished message.  Note: we intentionally do NOT skip reconnect
+            // for clean code-1000 closes here — mobile browsers close WebSocket
+            // connections with code 1000 / wasClean=true when a tab is
+            // backgrounded/suspended, which is NOT a permanent end of stream.
+            // finishedRef already covers the legitimate "stream ended" case.
+            if (cancelled || finishedRef.current) {
               return;
             }
 
-            // Otherwise, reconnect on unexpected/error closures
+            // Reconnect on any unexpected closure (including mobile tab suspension).
             retryAttemptsRef.current += 1;
             // Only show error if we haven't received any data yet
             if (!dataRef.current && retryAttemptsRef.current > 6) {
@@ -237,6 +238,34 @@ export const useJsonPatchWsStream = <T extends object>(
     deduplicatePatches,
     retryNonce,
   ]);
+
+  // Reconnect when the page becomes visible again.  Mobile browsers kill
+  // WebSocket connections when a tab is suspended; the onclose retry timer
+  // is also throttled/frozen in the background, so we need an explicit
+  // visibilitychange trigger to recover promptly when the user returns.
+  useEffect(() => {
+    if (!enabled || !endpoint) return;
+
+    const handleVisibilityChange = () => {
+      if (
+        document.visibilityState === 'visible' &&
+        !finishedRef.current &&
+        !wsRef.current
+      ) {
+        // Cancel any pending backoff timer and reconnect immediately.
+        if (retryTimerRef.current) {
+          window.clearTimeout(retryTimerRef.current);
+          retryTimerRef.current = null;
+        }
+        retryAttemptsRef.current = 0;
+        setRetryNonce((n) => n + 1);
+      }
+    };
+
+    document.addEventListener('visibilitychange', handleVisibilityChange);
+    return () =>
+      document.removeEventListener('visibilitychange', handleVisibilityChange);
+  }, [enabled, endpoint]);
 
   const isInitializedForCurrentEndpoint =
     isInitialized && initializedForEndpointRef.current === endpoint;


Juanlucasbg:
## Aggressive review summary — PR #3228

45-line fix to `packages/web-core/src/shared/hooks/useJsonPatchWsStream.ts` correcting WebSocket reconnect behavior on mobile tab suspension. Two coordinated changes. Verdict: **clean — recommend merge**.

### Why the change is correct

**Change 1**: Removes the `evt.code === 1000 && evt.wasClean` early-return from `onclose`. The original assumption (clean code 1000 means "stream ended") is wrong on mobile — Safari/Chrome on iOS/Android close WebSockets with `code=1000, wasClean=true` whenever the tab gets backgrounded or the OS suspends the WebView. Pre-PR, returning to the tab left the user staring at a frozen stream with no auto-recovery.

**Change 2**: Adds an explicit `visibilitychange` listener. Pre-PR, the existing exponential backoff would *eventually* reconnect, but mobile browsers throttle/freeze JS timers when the tab is hidden, so the backoff timer doesn't fire on resume — recovery is slow or never. The new listener cancels any pending backoff timer (`retryTimerRef`), resets `retryAttemptsRef` to 0, and bumps `retryNonce` to trigger an immediate reconnect.

### Findings

- **Structural**: 0 HIGH. Cleanup function correctly removes the listener. All refs (`retryTimerRef`, `wsRef`, `finishedRef`) are accessed via `.current`, so no stale-closure trap. The `[enabled, endpoint]` dep array is correct — both refs are stable identifiers.
- **Adversarial**: 0 HIGH. The "what if server cleanly closes when stream is genuinely done" scenario is covered by `finishedRef.current` — set when server emits a finished message before the close — so reconnect is still suppressed for legitimate completions. Reconnect storm risk is bounded by `retryAttemptsRef` + backoff in the existing reconnect path.
- **Security**: 0 HIGH. No new auth surface; same WebSocket endpoint, same connect path.
- **Conventions**: PASS. The added comment block clearly explains the *why* of removing the code-1000 short-circuit — exactly the case where a comment is warranted (subtle behavior, easily misread as a regression).

### NITs

- The visibilitychange handler reset `retryAttemptsRef.current = 0`. If the tab was backgrounded *during* a real backoff sequence (e.g. server is genuinely flaky), this loses the backoff state and may bias toward more aggressive reconnects. Probably fine — visibilitychange is a strong "user is back" signal — but worth a thought.
- A 60-90s timeout watchdog inside the visibility handler ("if we haven't reconnected in N seconds, surface UI") would be a nice follow-up.

### Verdict

Approve.

— Reviewed by automated single-pass review (mobile-WS-reconnect triage; full 4-tool battery skipped — diff is 45 lines and the comment block makes the rationale explicit).

PR #3251, opened 2026-03-24
fix: link issue-scoped workspaces during create/start

### LINKED ISSUE #3250: Workspace created from issue context can remain unlinked and invisible on the issue page
## Summary

When a workspace is created from an issue-scoped flow, the local workspace and first session can be created successfully, but the workspace may still remain unlinked to the issue in the remote store. When that happens, the issue page does not show the workspace at all, even though the workspace URL and session work normally.

## Observed behavior

- Create a workspace from issue context.
- The workspace is created locally and the first session starts.
- The issue page still shows no workspaces for that issue.
- Manually POSTing to `/api/workspaces/{workspace_id}/links` makes it appear immediately.

## Root cause

`CreateAndStartWorkspaceRequest` already includes `linked_issue`, but the local backend `create_and_start_workspace` route does not actually persist the remote workspace link. The current design relies on a second best-effort client-side link call after create/start succeeds.

That means issue linkage is fragile in two ways:

1. If the second request is skipped or never sent, the workspace stays local-only.
2. If the second request fails, the issue page becomes inconsistent even though the workspace/session already exist.

The same two-step pattern also affects MCP `start_workspace(issue_id=...)` flows.

## Expected behavior

If a create/start request includes `linked_issue`, the backend should ensure the remote issue link exists before reporting success. Repeated link attempts for the same workspace/issue pair should be treated as idempotent success.

## Suggested fix

- Move issue-link persistence into the backend create/start path.
- Make `/api/workspaces/{id}/links` idempotent when the workspace is already linked to the same issue.
- Keep returning a conflict if the workspace is already linked to a different issue.

I have a PR ready for this.

Jack47:
## Summary

Fixes #3250.

This change makes issue-scoped workspace creation persist the remote issue link during the backend create/start flow instead of relying on a second best-effort client request.

It also makes explicit workspace linking idempotent when the workspace is already linked to the same issue, while still returning a conflict if the workspace is already linked elsewhere.

## What changed

- added a shared server-side helper to sync a local workspace into the remote issue/workspace store
- called that helper from `create_and_start_workspace` whenever `linked_issue` is present
- updated `/api/workspaces/{id}/links` to reuse the same helper instead of always attempting a blind create
- added unit coverage for create-vs-update-vs-conflict link classification

## Why this fixes the bug

Before this patch, `CreateAndStartWorkspaceRequest` already carried `linked_issue`, but the backend ignored it for remote linkage. The issue page only showed workspaces present in the remote store, so a workspace could exist locally, run a session, and still be invisible on the issue page until someone manually POSTed to `/api/workspaces/{id}/links`.

After this patch, create/start ensures that the remote link exists before it reports success.

## Testing

- `cargo test -p server classify_remote_workspace_sync`
- `cargo check -p server`

---

> [!NOTE]
> **Medium Risk**
> Adds remote workspace create/update logic into the create/start path and changes link behavior to update-or-conflict based on existing remote state, which could affect workspace visibility and error handling across integrations.
> 
> **Overview**
> Issue-linked workspaces are now **synced to the remote issue/workspace store during `create_and_start_workspace`** (via new `sync_workspace_to_issue`), so the remote link exists immediately without relying on a follow-up client call.
> 
> `POST /api/workspaces/{id}/links` now reuses the same sync helper and becomes *idempotent*: it creates the remote workspace when missing, updates it when already linked to the same issue, and returns a `409 Conflict` when the workspace is linked to a different issue. Basic unit tests were added for the create/update/conflict classification.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87f2c2cf1f627aa77eed1235f4777f57606cdc67. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/server/src/routes/workspaces/create.rs
+++ b/crates/server/src/routes/workspaces/create.rs
@@ -15,8 +15,9 @@ use uuid::Uuid;
 use crate::{
     DeploymentImpl,
     error::ApiError,
-    routes::workspaces::attachments::{
-        ImportedIssueAttachment, import_issue_attachments_from_remote,
+    routes::workspaces::{
+        attachments::{ImportedIssueAttachment, import_issue_attachments_from_remote},
+        links::sync_workspace_to_issue,
     },
 };
 
@@ -238,6 +239,10 @@ pub async fn create_and_start_workspace(
         .workspace_manager()
         .load_managed_workspace(create_workspace_record(&deployment, name).await?)
         .await?;
+    let remote_client = match linked_issue.as_ref() {
+        Some(_) => Some(deployment.remote_client()?),
+        None => None,
+    };
 
     for repo in &repos {
         managed_workspace
@@ -250,9 +255,7 @@ pub async fn create_and_start_workspace(
         managed_workspace.associate_attachments(ids).await?;
     }
 
-    if let Some(linked_issue) = &linked_issue
-        && let Ok(client) = deployment.remote_client()
-    {
+    if let (Some(linked_issue), Some(client)) = (linked_issue.as_ref(), remote_client.as_ref()) {
         match import_issue_attachments_from_remote(
             &client,
             deployment.file(),
@@ -294,6 +297,17 @@ pub async fn create_and_start_workspace(
 
     let workspace = managed_workspace.workspace.clone();
     tracing::info!("Created workspace {}", workspace.id);
+
+    if let (Some(linked_issue), Some(client)) = (linked_issue.as_ref(), remote_client.as_ref()) {
+        sync_workspace_to_issue(
+            &deployment,
+            client,
+            &workspace,
+            linked_issue.remote_project_id,
+            linked_issue.issue_id,
+        )
+        .await?;
+    }
 
     let execution_process = deployment
         .container()
--- a/crates/server/src/routes/workspaces/links.rs
+++ b/crates/server/src/routes/workspaces/links.rs
@@ -1,4 +1,7 @@
-use api_types::{CreateWorkspaceRequest, PullRequestStatus, UpsertPullRequestRequest};
+use api_types::{
+    CreateWorkspaceRequest, PullRequestStatus, UpsertPullRequestRequest,
+    Workspace as RemoteWorkspace,
+};
 use axum::{
     Extension, Json, Router,
     extract::{Path as AxumPath, State},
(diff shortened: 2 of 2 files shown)


Juanlucasbg:
## Aggressive review summary — PR #3251

192-line fix to `crates/server/src/routes/workspaces/create.rs` and `links.rs` ensuring that workspaces created in an issue context are auto-linked to that issue (calling `sync_workspace_to_issue` post-create). Verdict: **clean — recommend merge**.

### Why the change is correct

Pre-PR `create_and_start_workspace` imported attachments from the linked issue but didn't actually link the workspace ↔ issue association. The user got a workspace that mentioned the issue's content but wasn't navigable from the issue's "workspaces" list. Post-PR:

1. Resolve `remote_client` once at the top, reuse for attachments + linking.
2. After the workspace record is committed, if there's a `linked_issue` + `remote_client`, call `sync_workspace_to_issue(deployment, client, workspace, project_id, issue_id)`.
3. Errors propagate via `?` — link failure aborts the create flow with the underlying error.

### Findings

- **Structural — PASS**: The `let (Some(linked_issue), Some(client)) = (linked_issue.as_ref(), remote_client.as_ref())` pattern is repeated for both the attachments path and the new linking path. Slightly verbose; not worth abstracting.
- **Adversarial — LOW**:
  - The link is a hard requirement now (`?` propagation): if `sync_workspace_to_issue` fails (network, remote-side validation, race with someone deleting the issue), the whole create call fails. Pre-PR, attachment-import failures were caught and only the link was missing; post-PR, link failure also tears down the create. PR #3270 (which I reviewed earlier in this sweep) added a UX path that surfaces link errors via a confirm dialog — its `linkErrorMessage` flow probably covers this case at the frontend layer, so the hard error here will be visible to the user. Worth confirming the two PRs compose correctly.
  - **Partial-success scenario**: workspace record created → link fails → user retries → second attempt would create a *second* workspace if this endpoint isn't idempotent. Is there a check to avoid duplicate workspaces for the same issue? Worth a sanity check in the maintainer's mental model.
- **Security — PASS**: `remote_client` is the same authenticated client used elsewhere; no new auth surface.
- **Conventions — PASS**: Re-uses existing `sync_workspace_to_issue` helper; doesn't reinvent.

### NITs

- **Test gap**: Integration test that drives the create endpoint with a `linked_issue` payload and asserts the post-create workspace appears in the issue's linked-workspace list would lock the contract. Without it, this regresses silently if `sync_workspace_to_issue` is later changed.
- **Compose with #3270**: Recommend the maintainer review-merge in this order: #3270 first (surfaces link errors via `ConfirmDialog`), then #3251 (now produces real link errors), so the UX side is ready when the backend starts emitting them. Otherwise users will see raw error toasts during the gap.
- **Telemetry**: Adding a `tracing::info!(workspace_id, issue_id, "linked workspace to issue")` log line on success would help operators correlate workspace-creation traces with link sync.

### Verdict

Approve.

— Reviewed by automated single-pass review (cross-service plumbing; full 4-tool battery skipped — diff is a contained refactor that calls an existing helper at the right point in the lifecycle).

PR #3270, opened 2026-03-25
fix: surface issue link failures after workspace creation

### LINKED ISSUE #3269: Issue-created workspace can silently fail to appear on the issue page
## Summary

When a workspace is created from an issue, the local workspace and agent session can start successfully, but the workspace may never show up in that issue's Workspaces section.

## Root cause

The issue page only shows remote workspaces whose `issue_id` matches the current issue. The local create-and-start flow creates the workspace first, then performs a follow-up `/api/workspaces/:id/links` request to associate it with the issue. If that second request fails, the UI currently only logs to the console and still opens the workspace, so the user gets no visible indication that the issue link is missing.

## Reproduction

1. Open an issue page.
2. Create and start a workspace from that issue.
3. Let the post-create link request fail because of a transient remote/API/auth/network problem.
4. Observe that the workspace opens and the session runs, but the issue page does not list the workspace.

## Expected behavior

If linking fails, the UI should explicitly tell the user that the workspace was created but not linked to the issue, and that it will not appear in the issue workspace list until linked.

## Actual behavior

The user sees a running workspace, but the issue page looks as if the workspace was never created.

Jack47:
Fixes #3269

## Summary

- return structured link-failure information from the workspace create/start mutation
- show an explicit dialog when the workspace is created but linking it to the issue fails
- preserve the existing successful flow while making the failure mode visible to the user

## Why

The issue page only lists workspaces whose remote `issue_id` matches the current issue. The local create/start flow links the issue in a follow-up request, so if that request fails the workspace still opens but silently disappears from the issue page. This change makes that state obvious to the user instead of swallowing it in `console.error`.

## Testing

- `pnpm run format`
- `pnpm --filter @vibe/web-core run check` *(fails on clean `origin/main` due unrelated existing type/dependency issues, including missing `@vibe/ui` component exports and missing `@tanstack/react-virtual` / markdown-related modules in `packages/web-core`)*

---

> [!NOTE]
> **Low Risk**
> Low risk: UI/UX and mutation return-shape changes that only affect the post-create issue-linking path; main workspace creation flow remains unchanged.
> 
> **Overview**
> Surfaces failures when linking a newly created workspace to an issue, instead of only logging them.
> 
> `useCreateWorkspace` now returns a structured `linkErrorMessage` alongside the created `workspace` when the follow-up `linkToIssue` call fails, and `CreateChatBoxContainer` displays a blocking `ConfirmDialog` warning the user that the workspace won’t appear in the issue’s workspace list until manually linked.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ec2586e2440e14077a5515e0ed77979c64742061. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/web-core/src/shared/components/CreateChatBoxContainer.tsx
+++ b/packages/web-core/src/shared/components/CreateChatBoxContainer.tsx
@@ -16,6 +16,7 @@ import {
 } from '@/shared/lib/string';
 import type { BaseCodingAgent, Repo } from 'shared/types';
 import { CreateChatBox } from '@vibe/ui/components/CreateChatBox';
+import { ConfirmDialog } from '@vibe/ui/components/ConfirmDialog';
 import { SettingsDialog } from '@/shared/dialogs/settings/SettingsDialog';
 import { CreateModeRepoPickerBar } from './CreateModeRepoPickerBar';
 import { ModelSelectorContainer } from '@/shared/components/ModelSelectorContainer';
@@ -253,6 +254,19 @@ export function CreateChatBoxContainer({
       data,
       linkToIssue,
     });
+
+    if (result.linkErrorMessage) {
+      await ConfirmDialog.show({
+        title: t('common:error'),
+        message: t('workspaces.linkAfterCreateError', {
+          defaultValue:
+            'The workspace was created and started, but linking it to the issue failed. It will not appear in the issue workspace list until you link it. Error: {{error}}',
+          error: result.linkErrorMessage,
+        }),
+        confirmText: t('common:ok'),
+        showCancelButton: false,
+      });
+    }
 
     if (result.workspace) {
       onWorkspaceCreated(result.workspace.id);
--- a/packages/web-core/src/shared/hooks/useCreateWorkspace.ts
+++ b/packages/web-core/src/shared/hooks/useCreateWorkspace.ts
@@ -1,6 +1,9 @@
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { workspacesApi } from '@/shared/lib/api';
-import type { CreateAndStartWorkspaceRequest } from 'shared/types';
+import type {
+  CreateAndStartWorkspaceRequest,
+  CreateAndStartWorkspaceResponse,
+} from 'shared/types';
 import { workspaceSummaryKeys } from '@/shared/hooks/workspaceSummaryKeys';
 
 interface CreateWorkspaceParams {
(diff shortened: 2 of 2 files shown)


Juanlucasbg:
## Aggressive review summary — PR #3270

43-line UX fix that surfaces issue-linking failures via `ConfirmDialog` instead of swallowing them to `console.error`. Verdict: **clean — recommend merge**.

### Why the change is correct

Pre-PR, `useCreateWorkspace` caught link errors and logged silently. The user got a created workspace that wasn't linked to the issue with no visible signal — the workspace appeared to "work" but never showed up in the issue's workspace list. The PR:

1. Captures the error message from the catch block (with proper instanceof Error narrowing + fallback).
2. Returns `{ workspace, linkErrorMessage? }` instead of just `{ workspace }` — a clean type extension.
3. The caller (`CreateChatBoxContainer`) checks `result.linkErrorMessage` and shows a `ConfirmDialog` with the explicit failure copy plus the underlying error message.
4. The confirm-dialog `await` correctly blocks navigation (`onWorkspaceCreated(result.workspace.id)`) until the user acknowledges the failure.

### Findings

- **Structural**: 0 HIGH. `linkErrorMessage` is `string | undefined`; the caller's `if (result.linkErrorMessage)` correctly narrows. Type-safe end-to-end.
- **Adversarial**: 0 HIGH. The error path explicitly normalizes non-`Error` thrown values to `'Unknown error while linking workspace to issue'`. No silent failure mode introduced.
- **Security**: 0 HIGH. The error message could carry sensitive info (e.g. backend stack trace fragments). Mild info-leak risk mitigated by the fact that this only displays to the local user who initiated the action.
- **i18n**: PASS. Uses `t('common:error')` and `t('workspaces.linkAfterCreateError', {defaultValue, error})` with a proper interpolation. `defaultValue` keeps the dialog functional even before translators add the new key.
- **Conventions**: PASS. Matches the existing `defaultValue`-fallback pattern used elsewhere in this codebase.

### NIT

- The console.error payload is now richer (`{workspaceId, projectId, issueId, error}`) which is helpful for support; consider a structured-log line via the project's telemetry sink rather than just `console.error` so the failure can be aggregated. Non-blocking — out of scope for this PR.

### Verdict

Approve.

— Reviewed by automated single-pass review (UX-error-surfacing triage; full 4-tool battery skipped — diff is 43 lines and the pattern is straightforward error-propagation refactor).

PR #3273, opened 2026-03-26
fix(server): return 404 when attempt-context path does not match a workspace

### LINKED ISSUE #3272: /api/containers/attempt-context returns 500 when cwd is outside any workspace
## Summary
When `GET /api/containers/attempt-context?ref=<path>` is called with a path that is not inside any Vibe Kanban workspace, the server currently maps `sqlx::Error::RowNotFound` to a `500 Internal Server Error`.

This is misleading for callers and logs. The request does not represent a server failure; it means there is no workspace context for that path.

## Reproduction
1. Start a local Vibe Kanban server.
2. Call `/api/containers/attempt-context?ref=<path>` with a path outside any known workspace.
3. Observe a `500` response and `Database(RowNotFound)` in server logs.

## Expected
The endpoint should return a not-found style response for paths that do not map to a workspace context.

## Actual
The endpoint returns `500 Internal Server Error`.

## Notes
`crates/mcp/src/task_server/mod.rs` already treats non-success responses from this endpoint as "no context", so returning `404` here is compatible with current MCP behavior.

Jack47:
## Summary
- map `sqlx::Error::RowNotFound` to `404 Not Found` instead of `500 Internal Server Error`
- add a regression test covering the row-not-found mapping

## Why this is safe
`crates/mcp/src/task_server/mod.rs` already treats non-success responses from this endpoint as "no context", so returning `404` preserves the current caller behavior while fixing the HTTP semantics and logs.

## Validation
- `cargo test -p server`
- `cargo fmt --all`
- `pnpm run format` reaches frontend formatting, but this local worktree does not have `node_modules`, so `prettier` is unavailable

Fixes #3272

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only HTTP error mapping for `sqlx::Error::RowNotFound` from 500 to 404 and adds a focused regression test, with no data writes or auth logic impacted.
> 
> **Overview**
> Returns **`404 Not Found`** (instead of `500`) when `ApiError::Database` wraps `sqlx::Error::RowNotFound`, improving semantics and reducing noisy internal-error logging.
> 
> Adds a small unit test in `crates/server/src/error.rs` to ensure the `RowNotFound` mapping produces a `NOT_FOUND` response.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dd250ceaa28647e7ccad9f61fbe1647b5b0391c0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/server/src/error.rs
+++ b/crates/server/src/error.rs
@@ -488,6 +488,9 @@ impl IntoResponse for ApiError {
             },
             ApiError::Executor(_) => ErrorInfo::internal("ExecutorError"),
             ApiError::CommandBuilder(_) => ErrorInfo::internal("CommandBuildError"),
+            ApiError::Database(sqlx::Error::RowNotFound) => {
+                ErrorInfo::not_found("DatabaseError", "Resource not found.")
+            }
             ApiError::Database(_) => ErrorInfo::internal("DatabaseError"),
             ApiError::Worktree(_) => ErrorInfo::internal("WorktreeError"),
             ApiError::Config(_) => ErrorInfo::internal("ConfigError"),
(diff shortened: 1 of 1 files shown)


Juanlucasbg:
## Rebase verification — PR #3273

- `crates/server/src/error.rs`

`error.rs` has been touched on `main` (most recently by #3315 which added a similar `RowNotFound → 404` rewrite for two adjacent endpoints). The PR's 17-LOC attempt-context change likely overlaps with #3315's pattern.

**Action**: rebase, then verify that #3315 didn't already cover the same surface. If yes, close as superseded. If the attempt-context route is genuinely separate from the container-context route #3315 fixed, re-apply this PR's mapping.

— Automated rebase-verification: confirmed conflict locally before posting. Likely superseded by #3315 — worth a maintainer check.

PR #3296, opened 2026-03-29
Add npx command normalization to fix command not found

chethanuk:
## Summary
- remove the redundant pre-override `npx` delimiter normalization in the Claude executor command builder
- remove the same redundant normalization in slash command discovery
- keep post-override normalization so custom `npx` base-command overrides still receive `--` when needed

## Files Changed
- `crates/executors/src/executors/claude.rs`
- `crates/executors/src/executors/claude/slash_commands.rs`

## Verification
- `cargo test -p executors test_build_command_builder_normalizes_npx_override`
- `cargo test -p executors test_discovery_builder_normalizes_npx_override`

---

> [!NOTE]
> **Medium Risk**
> Touches Claude executor command construction; incorrect normalization could change how arguments are passed to `npx` and break spawning or slash-command discovery for some configurations.
> 
> **Overview**
> Ensures custom Claude `base_command_override` values that start with `npx` are **post-override normalized** to include the `--` delimiter when missing, preventing downstream args from being interpreted as `npx` options.
> 
> Updates both the main Claude command builder and the slash-command discovery builder to apply this normalization after `apply_overrides`, and adds targeted async tests asserting an `npx` override becomes `... --`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5113860f6535399341b7e6f3fc1a5fbc68633bbb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -176,10 +176,9 @@ impl ClaudeCode {
             );
         }
 
-        let mut builder = CommandBuilder::new(ensure_npx_delimiter(base_command(
-            self.claude_code_router.unwrap_or(false),
-        )))
-        .params(["-p"]);
+        let mut builder =
+            CommandBuilder::new(base_command(self.claude_code_router.unwrap_or(false)))
+                .params(["-p"]);
 
         let plan = self.plan.unwrap_or(false);
         let approvals = self.approvals.unwrap_or(false);
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -66,6 +66,29 @@ fn base_command(claude_code_router: bool) -> &'static str {
     }
 }
 
+fn ensure_npx_delimiter(base: &str) -> String {
+    let mut tokens = base.split_whitespace();
+    let is_npx = tokens.next() == Some("npx");
+    if !is_npx {
+        return base.to_string();
+    }
+
+    if base.split_whitespace().any(|token| token == "--") {
+        return base.to_string();
+    }
+
+    format!("{base} --")
+}
+
+fn normalize_npx_base_command(builder: CommandBuilder) -> CommandBuilder {
+    let normalized_base = ensure_npx_delimiter(&builder.base);
+    if normalized_base == builder.base {
+        return builder;
+    }
+
+    builder.override_base(normalized_base)
+}
+
 fn normalize_claude_stderr_logs(
     msg_store: Arc<MsgStore>,
     entry_index_provider: EntryIndexProvider,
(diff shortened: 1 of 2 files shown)


Juanlucasbg:
## Aggressive review summary — PR #3296

94-line bug fix in `crates/executors/src/executors/claude.rs` and `crates/executors/src/executors/claude/slash_commands.rs` adding npx-delimiter normalization to user-supplied `base_command_override`. Verdict: **clean — recommend merge**.

### Why the change is correct

`npx -y @anthropic-ai/claude-code@2.1.62` is a common override. The downstream code appends args like `--print --replay-user-messages` to the base. Without a `--` delimiter between the npx flags and the package args, npx swallows the appended flags as its own options and fails with "command not found" or similar. The fix:

```rust
fn ensure_npx_delimiter(base: &str) -> String {
    let mut tokens = base.split_whitespace();
    let is_npx = tokens.next() == Some("npx");
    if !is_npx { return base.to_string(); }
    if base.split_whitespace().any(|token| token == "--") {
        return base.to_string();
    }
    format!("{base} --")
}
```

Three correctness gates:
1. **Only fires for npx** — `pnpm dlx`, `bunx`, etc. fall through unchanged.
2. **Idempotent** — re-running on a string that already has `--` is a no-op.
3. **Token-aware** — `any(|token| token == "--")` matches only on whole-word `--`, not on `--something` flags.

### Findings

- **Structural — PASS**: The helper takes `&str` and returns `String`; the wrapper `normalize_npx_base_command(builder: CommandBuilder)` short-circuits when no rewrite is needed. Clean separation between pure logic and the impure `CommandBuilder` mutation.
- **Adversarial — 0 HIGH**:
  - `npx` alone (no package): `tokens.next() == Some("npx")`, no `--` in any token, returns `"npx --"`. Harmless.
  - `npx -- pkg arg`: `--` already present, returns unchanged.
  - `npx pkg --version`: `--version` is one token, `--` is a separate token; the `any(|token| token == "--")` predicate compares whole tokens, so `--version != "--"`. Returns `"npx pkg --version --"` — adds the delimiter even though it isn't strictly needed (because `--version` already happens to be the only argument). Harmless: `npx pkg --version --` works the same as `npx pkg --version`.
  - `NPX foo`: case-sensitive match fails; passes through. Reasonable.
  - `  npx foo`: leading whitespace; `split_whitespace().next()` skips empty tokens, so `Some("npx")` is the first non-empty token. Correctly handled.
- **Security — 0 HIGH**: Pure string manipulation; no command exec.
- **Conventions — PASS**: Two new test cases (one per file) cover the main happy path. Both files import the same helper symmetrically.

### NITs

- `ensure_npx_delimiter` calls `split_whitespace()` twice (once via `tokens` after `.next()`, once in the `any` check). Negligible but trivially mergeable: collect into a `Vec` once, or check `.position(|t| t == "--").is_some()` on the same iterator. Not worth blocking.
- The `pnpm dlx` and `bunx` cases (similar tooling, similar delimiter need) aren't handled — out of scope but worth a follow-up. `pnpm dlx pkg` likewise needs `pnpm dlx pkg --` to forward subsequent flags.
- Tests exist but only cover the affirmative case. Add: `npx -- pkg` (idempotent), `pnpm exec foo` (no-op), `npx` (degenerate) — three more 5-line test cases that lock in the gates.

### Verdict

Approve.

— Reviewed by automated single-pass review (CLI-args correctness; full 4-tool battery skipped — change is contained, has tests, gates are sound).

PR #3338, opened 2026-04-09
Feature: HTTP Proxy support for Rust Binaries and binary download

NaN-KL:
Add HTTP ystem-proxy support to Reqwest library features for deployments that require HTTP proxy support. This enables the rust binaries and the webui to respect any system proxy directive set.

---

> [!NOTE]
> **Medium Risk**
> Changes how binaries/manifests are fetched by enabling proxy support in both Rust (`reqwest` features) and the Node download path (CONNECT tunneling + `NO_PROXY`), which could impact connectivity in varied network/proxy setups.
> 
> **Overview**
> **Adds proxy-aware networking for deployments behind proxies.** The workspace `reqwest` dependency now enables `socks` and `system-proxy` features so Rust binaries can respect system proxy configuration.
> 
> The `npx-cli` downloader now routes `fetchJson` and `downloadFile` through a new `httpsGet` helper that honors `HTTP(S)_PROXY`/`ALL_PROXY` and `NO_PROXY`, using HTTP `CONNECT` tunneling for HTTPS requests and preserving existing redirect/checksum handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71db8df63422cc41ab6bf149725c76302d22e1ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Juanlucasbg:
## Aggressive review summary — PR #3338

234-line proxy-support feature: Rust workspace `reqwest` gets `socks` + `system-proxy` features, npx-cli implements a custom HTTPS-over-HTTP-CONNECT tunnel for binary downloads. Verdict: **clean — recommend merge with two MED items**.

### What changed

**Rust side** (`Cargo.toml`): Adds `socks` and `system-proxy` to `reqwest`. This makes every Rust HTTP client in the workspace automatically respect `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` / `ALL_PROXY` / SOCKS proxies. Zero-touch for the rest of the codebase.

**npx-cli side** (`download.ts`): Custom `httpsGet` that:
1. Reads `HTTPS_PROXY` / `HTTP_PROXY` / `ALL_PROXY` / lowercase variants in priority order, skips `socks://` URLs.
2. Applies `NO_PROXY` matching with `*`, exact-match, dotted-prefix, and parent-domain rules.
3. Sends an `http.request({method: 'CONNECT'})` to the proxy, hooks `connect` event, and tunnels TLS through the returned socket via `https.get({socket, servername: ...})`.

### Findings

- **Structural — PASS**: `httpsGet` returns an interface compatible with `https.get`'s call sites (`.on('error', ...)`); existing callers don't need to change. Good adapter pattern.
- **Security — MED**: `parseInt(proxy.port || '80', 10)` defaults to port 80 when `proxy.port` is empty. `https://` proxies (rare but they exist for some corporate setups) typically use 443. Better: `proxy.port || (proxy.protocol === 'https:' ? '443' : '80')`. Otherwise `HTTPS_PROXY=https://corp-proxy/` silently connects to corp-proxy:80, fails confusingly.
- **Adversarial — MED**: `NO_PROXY` parsing has a subtle gap. The `host.endsWith('.' + pattern)` check matches `foo.example.com` against `example.com` — good. But `pattern.startsWith('.') && host.endsWith(pattern)` is redundant (already covered) and means `*.example.com` and `.example.com` and `example.com` all behave identically. CIDR notation (`192.168.0.0/16`) and port-specific entries (`example.com:8080`) are not supported. Document the supported subset.
- **Security — LOW**: TLS hostname pinning works correctly via `servername: target.hostname` — proxy can't MITM unless the operator has trusted the proxy's CA.
- **Conventions — PASS**: Comment block above `httpsGet` ("Make an HTTPS GET request, tunneling through an HTTP proxy via CONNECT if configured") explains the behavior.

### NITs

- **Reqwest's `system-proxy` feature** does reliably pick up env vars on first-client construction — but if vibe-kanban builds long-lived `reqwest::Client` instances at startup and the operator changes proxy env vars at runtime, the existing clients won't notice. Document this limitation. Probably OK in practice (you restart the process to change config).
- **No tests added** for `getHttpProxyUrl`, `isNoProxy`, or the CONNECT path. The pure helpers are easy to unit-test (just feed `process.env` mocks).
- **Error path on `connectReq.emit('error', ...)`**: the `connectReq` is returned to the caller as the resolve target, but the caller registers `.on('error')`. If the CONNECT itself fails before `connect` event fires (proxy unreachable), does `connectReq.emit('error', ...)` reach the registered listener? Verify with a quick smoke test against an unreachable proxy.
- **SOCKS skipped** is fine but worth surfacing in user-facing logs: when `HTTPS_PROXY=socks5://...` is set and the npx-cli silently goes direct, the user wonders why their proxy isn't being used.

### Verdict

Approve.

— Reviewed by automated single-pass review (proxy-feature add; full 4-tool battery skipped — Rust side is feature-flag-only, npx-cli side is a contained CONNECT implementation).

Issue #3340, opened 2026-04-09
MCP client (`vibe-kanban-mcp`) is incompatible with the self-hosted `remote-server` (path + envelope mismatch)


toddamerrill:
## Summary

Running `vibe-kanban@latest --mcp` against a self-hosted `remote-server`
(built from `crates/remote` in this repo) fails on every tool call with
`Failed to parse VK API response — error decoding response body`. The root
cause is that `vibe-kanban-mcp` is hardcoded to the hosted cloud API
surface, which the self-hosted `remote-server` does not implement.

## Environment

- Self-hosted deploy built from commit `576fd3e` (tagged `v0.1.41-20260403182044`)
- Backend URL: `https://<host>` (valid Let's Encrypt, `/v1/health` → `{"status":"ok","version":"0.1.26"}`)
- MCP client: `npx vibe-kanban@latest --mcp` → `vibe-kanban-mcp v0.1.36-20260323174633`
- Env: `VIBE_BACKEND_URL=https://<host>`

## What's happening

1. The self-hosted `remote-server` only mounts routes under `/v1/*`:
   ```rust
   // crates/remote/src/routes/mod.rs:147-148
   .nest("/v1", v1_public)
   .nest("/v1", v1_protected)
   ```
   Any request to `/api/*` falls through to the SPA index.html fallback
   and returns `200 text/html`.

2. The MCP crate at `crates/mcp/src/task_server/tools/*` (same commit, same
   repo) hardcodes `/api/*` paths everywhere, e.g.:
   ```
   crates/mcp/src/task_server/tools/organizations.rs:68:     "/api/organizations"
   crates/mcp/src/task_server/tools/remote_issues.rs:307:    "/api/remote/issues"
   crates/mcp/src/task_server/tools/mod.rs:346:             "/api/workspaces/{}/links"
   ```
   These paths appear to target the BloopAI-hosted cloud product, not
   `remote-server`.

3. When the MCP client hits (for example) `GET /api/organizations` against
   a self-host, it gets the SPA index.html with a 200 status. The response
   deserializer then fails trying to parse HTML as `ApiResponseEnvelope<ListOrganizationsResponse>`,
   which surfaces as the generic "Failed to parse VK API response" error.

4. Even if paths were rewritten, the response shape differs: the cloud API
   appears to wrap responses in `ApiResponseEnvelope { success, data, ... }`,
   while the self-hosted `remote-server`'s `/v1/*` handlers return the
[...]

Issue #3205, opened 2026-03-19
OpenCode executor: zombie processes after macOS sleep/wake, no lifecycle management


Jack47:
## Summary

After macOS sleep/wake cycles, `opencode serve` processes become zombie-like — process stays alive but HTTP server is unresponsive. VK continues sending requests to the stored port, resulting in I/O errors. VK has no mechanism to detect or recover from this state.

## Environment

- macOS (darwin arm64)
- VK running via `com.vibekanban.webui` launchd service
- OpenCode executor (`opencode-ai@1.1.59`)
- ~20 concurrent workspaces (normal usage)

## Reproduction

1. Start several workspaces with OpenCode executor
2. Let Mac sleep overnight (lid close / idle sleep)
3. Wake Mac in the morning
4. Try to run any session in an existing workspace
5. → `OpenCode executor error: I/O error: error sending request for url (http://127.0.0.1:{PORT}/session/...)`

## Root Cause Analysis

Each VK workspace session spawns `opencode serve` via:
```
npm exec opencode-ai@1.1.59 serve --hostname 127.0.0.1 --port 0
```

This creates ~3 processes per session (npm exec → node shim → opencode binary). VK stores the assigned port in its DB.

After sleep/wake:
- TCP connections are torn down by the OS
- `opencode serve` HTTP server doesn't recover, but the **process stays alive**
- RSS drops from ~60-200 MB to ~5 MB (zombie shell)
- VK sends requests to the stored port → connection refused / timeout → I/O error

With 20 concurrent workspaces, this results in **~50 zombie processes** consuming memory with no way to recover except manually killing them.

## Evidence

```
# Zombie process (started Mon 7AM, RSS collapsed to 4.8 MB)
PID 14464  RSS 4996 STARTED Mon07AM  opencode serve --hostname 127.0.0.1 --port 50466

# Port 50466 — no listener
$ lsof -i :50466
(empty)

# pmset log showing sleep/wake cycling
Maintenance Sleep at 10:19, 10:20, 10:28, FullWake at 10:38
```

## Requested Fix (VK side)

1. **Health check before request**: Before sending a request to an `opencode serve` process, verify the HTTP server is responsive (e.g., GET `/health` or TCP connect check)
2. **Auto-restart on failure**: If health check fails, kill the zombie process and spawn a new `opencode serve` instance
3. **Process cleanup on workspace archive/delete**: Kill associated `opencode serve` processes when workspace is no longer active
[...]

Issue #2920, opened 2026-02-25
Feature request: allow MCP to connect to an already-running VK instance


jonzarecki:
### Use Case

I run Vibe Kanban as a long-lived local server (`npx vibe-kanban` on port 13000) and want to manage tasks from an external MCP client (Cursor, Claude Desktop, Raycast, etc.) while keeping the web UI open (and containerized)

The current `--mcp` flag starts an entirely new VK process rather than connecting to the existing one. The second process tries to bind its own backend on an ephemeral port, which either conflicts with or ignores the running instance. Task tools (`list_tasks`, `create_task`, etc.) fail with `Failed to connect to VK API` because the MCP's own backend never fully initializes.

### Current Behavior

```
npx vibe-kanban@latest --mcp
```

- Spawns a full second VK backend (new port, new process)
- Does not connect to the VK server already running on `localhost:13000`
- Task management tools fail; the user's running server is ignored

### Proposed Behavior

A way to point the MCP at an existing VK instance, e.g.:

```bash
npx vibe-kanban@latest --mcp --connect http://localhost:13000
```

Or via environment variable:

```json
{
  "mcpServers": {
    "vibe_kanban": {
      "command": "npx",
      "args": ["-y", "vibe-kanban@latest", "--mcp"],
      "env": {
        "VIBE_KANBAN_URL": "http://localhost:13000"
      }
    }
  }
}
```

### Why This Matters

- **Primary workflow:** Users keep VK open in the browser for the board view and want MCP for AI-driven task creation/updates — these should work together
- **No workaround exists:** Third-party MCPs (e.g. [`mcp-better-vibe-kanban`](https://github.com/yigitkonur/mcp-better-vibe-kanban)) attempt to connect to an existing instance via REST, but VK v0.1.18 doesn't expose `/api/tasks` over HTTP (issues are cloud-synced via Electric-SQL), so those fail too

### Possible Approaches

1. **`--mcp-connect <url>` flag** — MCP acts as a thin client to the existing VK backend's internal API
2. **Auto-detect running instance** — Before spawning a new process, check if VK is already listening on the default port
3. **Expose a stable local API** — Have the running VK server offer an HTTP or IPC endpoint that MCP clients can consume directly

### Environment

- **vibe-kanban:** v0.1.18
- **OS:** macOS 15.3 (arm64)
- **MCP client:** Cursor IDE

janaka:
[...]

PR #3350, opened 2026-04-11
fix(discovery): persist and prefer canonical backend URL

### LINKED ISSUE #3349: MCP backend discovery can latch onto a stale port file and wrong loopback host, missing a healthy desktop backend on `localhost`/`::1`
## Summary

When MCP starts without an explicit `VIBE_BACKEND_URL`, it can lock onto a stale temp port file and reconstruct the backend URL as `http://127.0.0.1:{port}`.

That breaks against a healthy desktop backend when both of these are true:

1. the port file is stale, and
2. the live backend is reachable at `http://localhost:{live_port}` but not at `http://127.0.0.1:{live_port}`.

On macOS, this appears to be a real edge case for desktop/local setups because `localhost` may resolve through IPv6 (`::1`), while MCP recovery/default behavior still assumes IPv4 loopback.

The strongest control in this repro is that **manually bridging the stale `127.0.0.1:{stale_port}` endpoint to the real `http://localhost:{live_port}` backend immediately restored MCP reads**. That strongly suggests the MCP tools themselves are fine once discovery picks the correct backend.

---

## Environment

- macOS
- Desktop backend already running locally
- MCP started without overriding discovery via `VIBE_BACKEND_URL`
- Reproduced with a stale temp port file under:

```text
/var/folders/fv/mf1hrly52x52xn37lc4wq3y00000gn/T/vibe-kanban/vibe-kanban.port
```

Example stale file contents:

```json
{"main_port":54965,"preview_proxy_port":65154}
```

Live backend in the same repro was actually serving on:

```text
http://localhost:65301
```

but **not** on:

```text
http://127.0.0.1:65301
```

---

## Reproduction

1. Leave a stale `vibe-kanban.port` file behind, e.g.

   ```json
   {"main_port":54965,"preview_proxy_port":65154}
   ```

2. Start or keep running a healthy desktop backend that serves API responses on:
   - `http://localhost:65301/api/workspaces`
   - `http://localhost:65301/api/repos`

   while equivalent `http://127.0.0.1:65301/...` requests fail.

3. Start MCP **without** setting a correct `VIBE_BACKEND_URL`.

4. Invoke MCP read tools that depend on `/api/*`, for example:
   - `list_workspaces`
   - `list_repos`
   - `list_organizations`
   - `list_projects`

5. Observe MCP trying the stale port-file endpoint on `127.0.0.1:54965`, where requests fail.

6. As a control, add a local bridge from:

   ```text
   127.0.0.1:54965 -> http://localhost:65301
   ```

7. Re-run the same MCP tools.

---

## Actual behavior

Before bridging, MCP reads fail against stale-port IPv4 endpoints such as:

- `http://127.0.0.1:54965/api/workspaces`
- `http://127.0.0.1:54965/api/repos`

After bridging stale-port/IPv4 traffic to the real desktop backend, the same MCP reads succeed:

- `list_workspaces`
- `list_repos`
- `list_organizations`
- `list_projects`

This points to **backend discovery / address selection**, not to the individual tool handlers.

---

## Expected behavior

MCP should reliably connect to a healthy already-running local backend even if:

- a stale temp port file exists, and/or
- the backend is reachable on `localhost` but not on `127.0.0.1`.

At minimum, stale port metadata plus an IPv4-only loopback assumption should not be enough to break discovery.

---

## Why this looks like an upstream discovery bug

The relevant pieces line up with a combined failure mode:

- the port file stores **ports**, not the effective host/bind address
- MCP can reconstruct a backend URL using `127.0.0.1`
- the desktop path can run under `localhost` semantics, which on macOS may effectively mean `::1`
- MCP appears not to sufficiently validate that the discovered backend is both:
  - alive, and
  - the same address family / host actually serving the desktop API

In other words:

1. desktop runtime advertises or binds via `localhost`
2. MCP reconstructs via `127.0.0.1`
3. stale temp-dir port state is still present
4. MCP trusts that stale candidate too early and does not recover to the real live backend

---

## Why the existing upstream work seems adjacent but not sufficient

This looks related to the same general problem family as:

- **#2920** — MCP should be able to connect to an already-running VK instance
- **#3205** — stale/dead local endpoint recovery gaps
- **#3340** — MCP/backend assumption mismatch causing valid backends to be unusable
- **PR #3119** — fixed hardcoded local address selection by carrying forward the actual bind address
- **PR #3238** — fixed more `127.0.0.1` vs actual bind-address issues, especially around IPv6 / desktop behavior
- **#1585** — includes MCP logs showing historical reliance on port-file recovery and `127.0.0.1`

Those changes/issues are all adjacent to this report, but this repro still shows a remaining failure mode when **stale port state** and **loopback host-family mismatch** combine.

---

## Suggested fix directions

Any of these would likely make this class of failure go away:

1. **Persist host + port together**
   - Store the effective backend address (or full URL), not only port numbers, in the port file.

2. **Validate discovered backends before accepting them**
   - When MCP gets a candidate from env or port-file recovery, probe a lightweight endpoint first.
   - If it is dead, treat the port file as stale and continue recovery instead of failing hard.

3. **Try both local loopback forms during discovery**
   - If `127.0.0.1` fails, try `localhost` (and vice versa) before concluding the backend is unavailable.

4. **Prefer the runtime-advertised backend URL end-to-end**
   - If the server already knows its canonical local URL, preserve that exact value instead of reconstructing it later.

5. **Improve stale-port cleanup**
   - If the port file points to a dead endpoint, refresh or discard it automatically.

---

## References

- #1585
- #2920
- #3205
- #3340
- PR #3119
- PR #3238

naicud:
# fix(discovery): persist and prefer canonical backend URL

## What

- add an optional `backend_url` field to `PortInfo` and keep legacy raw-port / JSON discovery parsing intact
- write the canonical backend URL into `vibe-kanban.port` during server startup
- update MCP base URL resolution to prefer `VIBE_BACKEND_URL`, then explicit env host/port, then the discovered canonical backend URL before falling back to legacy host+port reconstruction
- add unit coverage for discovery parsing and base-URL precedence

## Why

Issue #3349 showed MCP could recover a stale or incomplete discovery record and rebuild `http://127.0.0.1:<port>` even when the healthy desktop backend was serving on `http://localhost:<port>` / `::1`.

Persisting the backend's canonical runtime URL lets MCP reuse the same endpoint the server actually bound, avoiding loopback-family mismatches while preserving existing overrides and legacy discovery compatibility.

## How

- extend `crates/utils/src/port_file.rs` with `PortInfo::backend_url`, a `read_port_info` helper, and compatibility tests for legacy formats
- write discovery metadata through `write_port_file_with_proxy_and_backend_url(...)` from `crates/server/src/startup.rs`
- refactor `resolve_base_url` in `crates/mcp/src/bin/vibe_kanban_mcp.rs` into source-precedence logic that prefers the canonical discovered URL when present

## Testing

- `cargo test -p utils --lib`
- `cargo test -p mcp --bin vibe-kanban-mcp`
- `cargo test -p server --lib --no-run`

## Related

- Closes #3349
- Adjacent: #2920, #3205, #3340, #3119, #3238

---

> [!NOTE]
> **Medium Risk**
> Changes backend discovery/URL resolution between server and MCP; mistakes could break local connectivity if the port file or precedence logic is wrong, but scope is limited and covered by new unit tests.
> 
> **Overview**
> Persists a *canonical* `backend_url` alongside ports in the temp discovery file (`PortInfo`), while keeping backwards-compatible parsing for legacy raw-port and older JSON formats.
> 
> On server startup, writes this canonical URL into the discovery record, and updates MCP base-URL resolution to prefer `VIBE_BACKEND_URL`, then explicit env host/port, then the discovered canonical URL before falling back to reconstructing `http://<host>:<port>`. Adds unit tests covering discovery parsing and URL precedence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9e7babd5d1fb65d6efec07f3d3767ace7d4dd97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/mcp/src/bin/vibe_kanban_mcp.rs
+++ b/crates/mcp/src/bin/vibe_kanban_mcp.rs
@@ -2,7 +2,7 @@ use mcp::task_server::McpServer;
 use rmcp::{ServiceExt, transport::stdio};
 use tracing_subscriber::{EnvFilter, prelude::*};
 use utils::{
-    port_file::read_port_file,
+    port_file::{PortInfo, read_port_info},
     sentry::{self as sentry_utils, SentrySource, sentry_layer},
 };
 
@@ -98,7 +98,38 @@ where
 }
 
 async fn resolve_base_url(log_prefix: &str) -> anyhow::Result<String> {
-    if let Ok(url) = std::env::var("VIBE_BACKEND_URL") {
+    let backend_url = std::env::var("VIBE_BACKEND_URL").ok();
+    let host = std::env::var(HOST_ENV)
+        .or_else(|_| std::env::var("HOST"))
+        .unwrap_or_else(|_| "127.0.0.1".to_string());
+
+    let explicit_port =
+        match std::env::var(PORT_ENV)
+            .or_else(|_| std::env::var("BACKEND_PORT"))
+            .or_else(|_| std::env::var("PORT"))
+        {
+            Ok(port_str) => Some(port_str.parse::<u16>().map_err(|error| {
+                anyhow::anyhow!("Invalid port value '{}': {}", port_str, error)
+            })?),
+            Err(_) => None,
+        };
+
+    let port_info = match explicit_port {
+        Some(_) => None,
+        None => Some(read_port_info("vibe-kanban").await?),
+    };
+
+    resolve_base_url_from_sources(log_prefix, backend_url, host, explicit_port, port_info)
+}
+
+fn resolve_base_url_from_sources(
+    log_prefix: &str,
+    backend_url: Option<String>,
+    host: String,
+    explicit_port: Option<u16>,
+    port_info: Option<PortInfo>,
+) -> anyhow::Result<String> {
+    if let Some(url) = backend_url {
         tracing::info!(
             "[{}] Using backend URL from VIBE_BACKEND_URL: {}",
             log_prefix,
@@ -107,28 +138,29 @@ async fn resolve_base_url(log_prefix: &str) -> anyhow::Result<String> {
         return Ok(url);
     }
 
-    let host = std::env::var(HOST_ENV)
-        .or_else(|_| std::env::var("HOST"))
-        .unwrap_or_else(|_| "127.0.0.1".to_string());
+    if let Some(port) = explicit_port {
+        tracing::info!("[{}] Using port from environment: {}", log_prefix, port);
+        let url = format!("http://{}:{}", host, port);
+        tracing::info!("[{}] Using backend URL: {}", log_prefix, url);
+        return Ok(url);
+    }
 
-    let port = match std::env::var(PORT_ENV)
-        .or_else(|_| std::env::var("BACKEND_PORT"))
-        .or_else(|_| std::env::var("PORT"))
-    {
-        Ok(port_str) => {
-            tracing::info!("[{}] Using port from environment: {}", log_prefix, port_str);
-            port_str
-                .parse::<u16>()
-                .map_err(|error| anyhow::anyhow!("Invalid port value '{}': {}", port_str, error))?
-        }
-        Err(_) => {
-            let port = read_port_file("vibe-kanban").await?;
-            tracing::info!("[{}] Using port from port file: {}", log_prefix, port);
-            port
-        }
-    };
+    let port_info = port_info.ok_or_else(|| anyhow::anyhow!("Missing port file information"))?;
+    if let Some(url) = port_info.backend_url {
+        tracing::info!(
+            "[{}] Using canonical backend URL from port file: {}",
+            log_prefix,
+            url
+        );
+        return Ok(url);
+    }
 
-    let url = format!("http://{}:{}", host, port);
+    tracing::info!(
+        "[{}] Using port from port file: {}",
+        log_prefix,
+        port_info.main_port
+    );
+    let url = format!("http://{}:{}", host, port_info.main_port);
     tracing::info!("[{}] Using backend URL: {}", log_prefix, url);
     Ok(url)
 }
@@ -158,7 +190,10 @@ fn init_process_logging(log_prefix: &str, version: &str) {
 
 #[cfg(test)]
 mod tests {
-    use super::{LaunchConfig, McpLaunchMode, resolve_launch_config_from_iter};
+    use super::{
+        LaunchConfig, McpLaunchMode, resolve_base_url_from_sources, resolve_launch_config_from_iter,
+    };
+    use utils::port_file::PortInfo;
 
     #[test]
     fn orchestrator_mode_does_not_require_session_id() {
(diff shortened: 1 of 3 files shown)


Juanlucasbg:
## Aggressive review summary — PR #3350

221-line plumbing fix across `crates/utils/src/port_file.rs`, `crates/server/src/startup.rs`, and `crates/mcp/src/bin/vibe_kanban_mcp.rs`. Replaces the port-only on-disk discovery file with a `PortInfo` struct that captures the canonical backend URL (port + host + presumably scheme). MCP binary now prefers that canonical URL when not overridden by env vars. Verdict: **clean — recommend merge**.

### Why the change is correct

The pre-PR discovery flow was: server writes `port_file` containing just an integer; MCP reads it and constructs `http://${HOST_or_default}:${port}`. The MCP-side host construction defaults to `127.0.0.1`, which broke any deployment where the server bound to a non-default host (Docker, remote SSH, custom binding). The PR makes the server write a richer `PortInfo` so the MCP can use exactly the URL the server itself is listening on.

The refactor of `resolve_base_url` is well-shaped: precedence is `VIBE_BACKEND_URL` env > explicit `PORT/BACKEND_PORT` env > `port_info` from the discovery file. Each branch logs which source it picked, which is exactly the right operator-debugging surface.

### Findings

- **Structural — PASS**: The split into `resolve_base_url` (env reads) and `resolve_base_url_from_sources` (pure logic) makes the latter unit-testable. The `Option<u16>` / `Option<PortInfo>` shape correctly distinguishes "not set" from "set to default".
- **Adversarial — LOW**:
  - **File schema migration**: If a stale `port_file` from an older version exists (pre-PR JSON shape), the new `read_port_info` deserializer needs to either (a) handle the legacy shape or (b) fail with a clear "stale port file, please restart server" message. Worth verifying — not visible in this excerpt.
  - **Race**: Server writes `port_info`, MCP reads. If MCP starts before server has finished writing, the read fails. Existing `read_port_file` likely already handled retries; ensure `read_port_info` preserves that behavior.
- **Security — LOW**: The discovery file is presumably world-readable in temp dir. Holding the canonical URL there isn't sensitive (it's a localhost endpoint), but if a future deployment puts a non-localhost canonical URL in there, the file's permissions become more important. Consider making it `0600`.
- **Conventions — PASS**: tracing-info logging on every branch.

### NITs

- **Test coverage**: The pure `resolve_base_url_from_sources` function is begging for unit tests covering the four branches (backend_url set, explicit_port set, port_info available, all three missing). One test per branch.
- **Backward-compat**: If the new `PortInfo` JSON shape isn't backward-compatible, document the breaking change in PR description so users running an old MCP against a new server don't silently fall back.

### Verdict

Approve.

— Reviewed by automated single-pass review (discovery plumbing; full 4-tool battery skipped — refactor follows clean-input-output split with logged precedence).

PR #3360, opened 2026-04-15
[codex] Fix workspace chat scroll jumps

artinflight:
## What changed

This PR fixes workspace chat views unexpectedly jumping upward while new content arrives or the composer area changes size.

- adds a shared hook to keep the conversation pinned to the latest messages when the chat box height changes
- applies that behavior to the kanban flyout workspace chat
- reuses the same hook in the full workspace and VS Code views so the behavior stays consistent across shells

## Why

In the kanban flyout, the conversation could appear to randomly scroll backward even when the user had not intentionally scrolled up. That made it look like older problems or earlier messages had resurfaced.

## Root cause

The workspace chat list was already tracking whether the user was at the bottom, but the kanban flyout was missing the resize compensation that keeps the viewport visually anchored when the composer/tool area grows or shrinks. When the height changed, the visible conversation window effectively moved upward.

The full workspace and VS Code views had similar logic inline already, so this PR centralizes that logic into one hook and uses it everywhere the workspace chat stack is rendered.

## User impact

- chat stays anchored to the newest messages when you are already at the bottom
- intentional upward scrolling still works normally
- jump-to-point / jump-to-message actions still work normally
- kanban flyout and full-screen workspace chat behavior now match

## Validation

- `pnpm run web-core:check`
- `pnpm run local-web:check`
- `pnpm run web-core:format`
- `pnpm run local-web:format`

### DIFF
--- a/packages/web-core/src/features/workspace-chat/ui/usePinConversationToBottomOnChatBoxResize.ts
+++ b/packages/web-core/src/features/workspace-chat/ui/usePinConversationToBottomOnChatBoxResize.ts
@@ -1,0 +1,63 @@
+import { useEffect, useRef, type RefObject } from 'react';
+
+import type { ConversationListHandle } from './ConversationListContainer';
+
+interface UsePinConversationToBottomOnChatBoxResizeOptions {
+  containerRef: RefObject<HTMLElement | null>;
+  conversationListRef: RefObject<ConversationListHandle | null>;
+  isAtBottom: boolean;
+  scopeKey: string;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


Juanlucasbg:
## Rebase status (informational — draft respected)

Draft PR. Posting rebase data as a heads-up:

- **Rebase against current `main`**: clean (no conflicts).
- Mergeable on GitHub.

When you mark this ready-for-review, the standard merge gate applies — no rebase work needed first.

— Automated rebase status check; no action requested while in draft.

PR #3369, opened 2026-04-18
[codex] Fix remote task variable fallbacks

davidrudduck:
## What changed

- fixed remote task variable preview so Hive-loaded tasks still expose system/runtime variables instead of collapsing to an empty variable set
- blocked variable mutation routes from operating on the wrong task by enforcing `{task_id}` ownership checks before update/delete
- rejected variable creation for remote task contexts and tightened remote node lookup so `get_available_nodes` only falls back for Hive-synthesized tasks

## Why

Cross-node task viewing had mostly been wired up already, but the variable preview path still re-queried local task state and could fail for Hive-only tasks. The variable update/delete routes also trusted only `var_id`, which allowed cross-task mutation through the wrong route scope.

## Impact

- remote task variable preview now degrades gracefully without losing core task context
- task variable writes stay scoped to the task in the URL
- remote node selection avoids using a stale local project id when only a Hive-backed task is available

## Validation

- `cargo test -p server task_variables::tests -- --nocapture`
- `cargo test -p server streams::tests -- --nocapture`
- `cargo test -p server proxy::tests -- --nocapture`
- `cargo test -p server message_queue::tests -- --nocapture`


davidrudduck:
Opened against the wrong upstream repository by mistake. Superseded by the fork-only PR at https://github.com/davidrudduck/vk-swarm/pull/442 .

PR #3391, opened 2026-04-24
[codex] fix workspace live status refresh

artinflight:
## What changed
- refresh workspace summary queries when the live workspace websocket stream changes
- debounce those refreshes slightly and skip the initial snapshot so normal live updates trigger refetches without restoring blanket polling

## Why this changed
Workspace live state in the sidebar and issue cards is assembled from two sources:
- `isRunning` comes from the workspace stream
- review/attention indicators such as `hasUnseenActivity` come from the workspace summaries query

When polling was removed from `useWorkspaces`, the summary side stopped refreshing during normal live updates. That left finished workspaces showing the running state until another action, such as opening the workspace, forced a refresh.

## Impact
- finished workspaces should switch from the running indicator to the needs-review indicator without being opened first
- this keeps the lower-churn event-driven model instead of reintroducing periodic polling

## Root cause
The frontend stopped refetching workspace summaries after live workspace state changes, so the UI could combine fresh stream data with stale summary data.

## Validation
- manual code inspection of `useWorkspaces.ts`
- attempted `pnpm run format` but local dependency install is incomplete (`prettier` not found)
- attempted `NODE_OPTIONS=--max-old-space-size=4096 pnpm --filter @vibe/web-core run check` but local dependency install is incomplete (`tsc` not found)


Juanlucasbg:
## Rebase status (informational — draft respected)

Draft PR (11821 LOC scope), posting rebase data as a heads-up:

- **Rebase against current `main`**: fails with conflict in `crates/local-deployment/src/container.rs` only.

Single-file conflict despite the largest scope of any draft in the queue — that's a strong positive signal. Same conflict as #3397 (also yours) — both touch `container.rs`. Likely a coordinated pair of drafts.

The `container.rs` file has been actively churned by maintainer merges (#3370, #3318, #3314) in early April. Resolvable when you're ready.

— Automated rebase status check; no action requested while in draft.

Issue #3343, opened 2026-04-10
bug: Long agent responses rendered twice in UI — race condition between streaming and status-transition reload


Liyue2341:
### Description

When Claude Code produces a long response, the **entire content is rendered twice** within the same message in the web UI. Short responses display correctly. The duplication is 100% reproducible on every long response.

The backend terminal output shows **no duplication** — this is purely a frontend issue.

### Environment

- **Vibe Kanban:** 0.1.36 (latest as of 2026-04-10)
- **OS:** Windows 11 (win32-x64)
- **Node.js:** v24.13.0
- **Agent:** Claude Code (official)
- **Browser:** Chrome (latest)

### Steps to Reproduce

1. `npx vibe-kanban`
2. Open any workspace with Claude Code
3. Send a prompt that elicits a **long** response (500+ words)
4. Observe: the entire response content appears **twice** in the same message bubble

Short responses (a few sentences) render correctly.

### Root Cause Analysis

I traced this to a **race condition** in `useConversationHistory.ts` between two `useEffect` hooks:

**Effect 1 (line ~437-471)** — Streaming effect:
When a process is `running`, it calls `loadRunningAndEmitWithBackoff()` which opens a WebSocket stream and continuously updates `displayedExecutionProcesses.current[process.id].entries` as chunks arrive. When the stream finishes, `streamingProcessIdsRef` is cleaned up in `.finally()`.

**Effect 2 (line ~473-519)** — Status-transition reload:
When a process transitions from `running` → non-`running` (completed), it calls `loadEntriesForHistoricExecutionProcess()` which **re-fetches the complete entry list** and writes it into `displayedExecutionProcesses.current[process.id]`.

**The race:**
For long responses, when the process status changes to "completed", Effect 2 fires immediately. But Effect 1's WebSocket stream **may still be open** (it hasn't received the `finished` event yet, so `streamingProcessIdsRef` still contains the process ID). Both effects write to the same `displayedExecutionProcesses.current[process.id]` state object. The result is that entries get emitted twice to the UI.

For short responses this doesn't happen because the stream completes and closes **before** the status transition triggers Effect 2.

**Key observation:** Effect 2 (line ~495) does **not** check `streamingProcessIdsRef` before reloading — it has no awareness that a streaming connection for the same process may still be active.

### Suggested Fix

Before reloading in Effect 2, either:

1. **Wait for the streaming connection to close:**
```typescript
// In the status-transition effect, before reloading:
[...]

PR #3394, opened 2026-04-24
fix(claude): skip Result.success re-emit when assistant content was already streamed

Isla-Liu:
Re-submits #3383 (the original PR's head branch became unavailable after I migrated to a cleaner public fork). Same patch, rebased on current `main` — no conflicts with the CLI version bump in #3384.

## Summary

Fixes a backend dedup bug in `ClaudeLogProcessor::normalize_entries` that appends the final assistant message as a second `NormalizedEntry` whenever the Claude CLI's terminal `Result { subtype: "success", result }` event is processed after the stream already delivered assistant content. The UI then renders two back-to-back identical copies of the last reply, and the duplication persists across page refresh because it lives in the persisted normalized conversation (which is re-derived from the raw JSONL on every load).

## The bug

In `crates/executors/src/executors/claude.rs`, the `Result` event handler gates emission of the final AssistantMessage on:

```rust
|| matches!(&self.last_assistant_message, Some(message) if !message.contains(text))
```

Substring containment is the wrong predicate. The terminal `result` string produced by Claude CLI is not a verbatim copy of the streamed content — in real sessions the two commonly differ by a byte or two somewhere in the body. A representative case observed in an actual Claude Code session:

- Streamed `last_assistant_message` length: **3061** chars
- `Result.success.result` length: **3060** chars
- Neither a substring of the other
- Not equal after `.trim()` either

The `!contains` predicate therefore falls through, the `Result.success.result` is emitted as a second `AssistantMessage` entry with the same semantic content, and the UI shows the final reply twice.

## The fix

The `Result.success.result` field is redundant metadata when the stream has already delivered assistant content (via `type=assistant` events or `stream_event` deltas — both paths run through `content_item_to_normalized_entry`, which sets `last_assistant_message`). The correct predicate is simply:

```rust
self.last_assistant_message.is_none()
```

Only fall back to emitting Result.success's `result` as a NormalizedEntry when no assistant content was seen this turn at all.

## Tests

- Updates the regression test to `test_result_message_skips_when_assistant_content_already_streamed`, exercising a divergent-content scenario (`"Hello world"` streamed vs. `"Hello World"` in `Result.success` — not a substring, not equal after `.trim()`) that matches the real-world bug signature.
- `test_result_message_emits_final_text_if_not_seen` continues to pass (it starts with a fresh processor where `last_assistant_message` is `None`, preserving the fallback path).

```
cargo test -p executors test_result_message
```

## Scope

- **Affected:** `ClaudeLogProcessor` — and therefore Claude Code and Amp (`HistoryStrategy::AmpResume` reuses `ClaudeLogProcessor`).
- **Not affected:** audited sister executors do not share this redundant-summary pattern. Codex uses `UpdateMode::Set` (in-place replace). Gemini / Copilot / QwenCode route through the ACP common layer which does not re-emit on `Done`. Cursor explicitly `no-op`s on its `Result` variant. Droid uses a `sent_completion` flag guard. Opencode uses part-based streaming.

## Related

- #3343 — "bug: Long agent responses rendered twice in UI"

The symptom reported there matches the backend scenario fixed by this PR. The reporter hypothesised a frontend race in `useConversationHistory.ts`, but reproducing against an actual affected session shows:

1. Refreshing the page does not make the duplicate go away. It survives page refresh because it is baked into the normalized conversation that the frontend pulls from the backend.
2. The backend JSONL stream on disk contains a single `type=assistant` text content item plus one `type=result` summary event — running the current `normalize_entries` against it produces **two** `AssistantMessage` entries; running the fixed version produces **one**.

These two pieces of evidence rule out a pure frontend race for the observed bug and pinpoint the backend normalization layer as the cause. The frontend-race analysis in #3343 may still describe an independent, separate bug, but it is not the cause of the duplicated final message.

## Test plan

- [x] `cargo test -p executors test_result_message_skips_when_assistant_content_already_streamed` — reproduces the bug on `main`, passes with this change
- [x] `cargo test -p executors test_result_message_emits_final_text_if_not_seen` — continues to pass (legacy fallback)
- [x] Manual: a long Claude Code response in the UI shows a single copy of the final reply instead of two identical copies, and the single copy persists across page refresh

---

> [!NOTE]
> **Medium Risk**
> Adjusts assistant message emission logic in `ClaudeLogProcessor` for Claude CLI `Result.success` events, which can affect what gets persisted to conversation history. Risk is limited in scope but could still change visible output (missing/extra final messages) if the new per-turn reset logic is wrong.
> 
> **Overview**
> Prevents duplicated final Claude assistant replies by changing `Result { subtype: "success" }` handling to **only emit the `result` fallback when no assistant content was streamed for that turn** (removing the prior substring-based dedupe).
> 
> Resets `last_assistant_message` after each `Result` event so the fallback can still emit in later turns when a processor is reused. Adds focused regression tests covering (1) divergent streamed vs. `Result.success.result` text (should skip) and (2) a later turn with only `Result.success` (should emit).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2752e003d00d88c1c78c6da1a0b3d4bc91766d95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/executors/src/executors/claude.rs
+++ b/crates/executors/src/executors/claude.rs
@@ -1826,8 +1826,7 @@ impl ClaudeLogProcessor {
                     patches.push(ConversationPatch::add_normalized_entry(idx, entry));
                 } else if matches!(subtype.as_deref(), Some("success"))
                     && let Some(text) = result.as_ref().and_then(|v| v.as_str())
-                    && (self.last_assistant_message.is_none()
-                        || matches!(&self.last_assistant_message, Some(message) if !message.contains(text)))
+                    && self.last_assistant_message.is_none()
                 {
                     let entry = NormalizedEntry {
                         timestamp: None,
@@ -1840,6 +1839,14 @@ impl ClaudeLogProcessor {
                     let idx = entry_index_provider.next();
                     patches.push(ConversationPatch::add_normalized_entry(idx, entry));
                 }
+
+                // Reset so the skip check's "this turn" semantic is accurate:
+                // only assistant content streamed within the turn that ended
+                // at this Result event should suppress emission. If a
+                // subsequent turn arrives with no streamed assistant content
+                // (e.g. processor reused across turns), its Result.success
+                // fallback must still emit.
+                self.last_assistant_message = None;
             }
             ClaudeJson::ApprovalRequested {
                 tool_call_id,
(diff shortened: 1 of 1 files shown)


Isla-Liu:
## Test plan verification (all three items)

### 1 & 2. `cargo test` (red-green verified)

Both targeted tests run on the fix branch + against an `main` worktree with the new test code cherry-picked alongside the unfixed production code:

| Test | On `main` (no fix) | On fix branch |
|---|---|---|
| `test_result_message_skips_when_assistant_content_already_streamed` | **FAIL** — duplicate `"Hello World"` AssistantMessage emitted (the exact UI symptom) | **pass** |
| `test_result_message_emits_final_text_if_not_seen` | pass | pass |
| `test_result_message_emits_fallback_in_later_turn_after_reset` | pass (coincidental on `main`: buggy `!contains` happens to give the right result for divergent text) | pass |

Failure on `main` (proves the regression test catches the bug):

```
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 41 filtered out
panicked at crates\executors\src\executors\claude.rs:2879:9:
Result.success must not be re-emitted when an assistant message was already streamed this turn,
got duplicate: [NormalizedEntry { ..., entry_type: AssistantMessage, content: "Hello World", ... }]
```

Pass on fix branch:

```
running 3 tests
test executors::claude::tests::test_result_message_emits_final_text_if_not_seen ... ok
test executors::claude::tests::test_result_message_skips_when_assistant_content_already_streamed ... ok
test executors::claude::tests::test_result_message_emits_fallback_in_later_turn_after_reset ... ok
test result: ok. 3 passed; 0 failed; 0 ignored
```

### 3. Manual UI test

Built backend + frontend from `fix/claude-result-success-dedup` on isolated ports (3003 frontend / 3004 backend) with a fresh `dev_assets/` DB. Created a workspace at the repo on the fix branch, submitted a long-response prompt to Claude Code (Opus High), captured a 16,823-char streamed response, then refreshed.

| State | Bubbles in DOM | Lengths |
|---|---|---|
| During streaming | 2 | 2,074 + 16,927 chars (share first 730 chars then diverge — first ends mid-sentence) |
| **After refresh** | **1** | **16,937 chars** |

→ Single copy persists across refresh. ✅

### Caveat that's worth flagging

The Claude CLI 2.1.119 emitted **zero `Result` events** in the entire 840-line JSONL log of this run (counts: `assistant: 9, stream_event: 626, user: 4, system: 10, result: 0`). That means the specific bug condition this PR fixes — `Result.success.result` re-emit when streamed content already exists — wasn't actually triggered in this end-to-end UI run. The unit-level red-green proof above is what dispositively exercises the fixed predicate.

The transient 2-bubble-during-streaming artifact appears to be a separate frontend behavior (it self-resolves on refresh and does not match the dup signature of #3343 — the two bubbles have *different* lengths, not identical), so it isn't the bug this PR is fixing.

### Other checks

- `cargo clippy -p executors --lib`: clean (warnings only in unrelated `crates/utils/`)
- Full `cargo test -p executors`: 41 pass; 3 unrelated pre-existing Windows-path test failures (`test_path_relative_conversion`, `test_ls_tool_content_extraction`, `test_amp_tool_aliases_create_file_and_edit_file`) — they exist on `main` and don't touch any code modified by this PR.

Juanlucasbg:
## Aggressive review summary — PR #3394

86-line bug fix to `crates/executors/src/executors/claude.rs` correcting `Result.success` re-emission logic, plus two new test cases. Verdict: **clean — recommend merge**.

### Why the change is correct

**Pre-PR predicate**:
```rust
self.last_assistant_message.is_none()
    || matches!(&self.last_assistant_message, Some(message) if !message.contains(text))
```

The `contains` check is fragile: in real Claude CLI output, the terminal `Result.success.result` text often differs from the streamed assistant content by a byte or two somewhere in the middle (capitalization, punctuation normalization). When that happens, `contains` returns false and the result gets emitted *again*, producing the duplicated assistant message in the UI that's the symptom.

**Post-PR predicate** (a) drops the substring check entirely and uses the cleaner "did the streaming already deliver assistant content this turn?" test, and (b) **resets `self.last_assistant_message = None` after the Result event** so the gate's per-turn semantic is preserved when the processor is reused across turns.

The reset is the key insight: without it, the gate would stay "an assistant message was seen" for the lifetime of the processor, silently dropping legitimate fallback emissions in subsequent turns where Claude produced no streamed text (tool-use-only turn, etc.).

### Findings

- **Structural — PASS**: The `if !condition && let Some(...) && cond` chain is more readable post-PR. Comment block (lines 1842-1848) explains the per-turn reset rationale — exactly the case where a comment is warranted.
- **Adversarial — 0 HIGH**: Both directions are tested. `test_result_message_skips_when_assistant_content_already_streamed` covers the original bug; `test_result_message_emits_fallback_in_later_turn_after_reset` covers the "what if reset is missing" regression. Both tests use realistic Claude CLI JSON shapes.
- **Test quality**: The `"Hello world"` vs `"Hello World"` test fixture (one capital W) is the exact byte-level difference that real Claude output exhibits. Author saw the actual broken signal, not a synthetic case. Strong.
- **Conventions — PASS**: Test file structure matches the surrounding `mod tests` patterns; `normalize_helper` reused.
- **Security**: N/A — log processing.

### NITs

- The new comment block at lines 1842-1848 is clearly written but slightly long; could be condensed to two lines.
- Consider adding a third test case: tool-use-only turn followed by Result.success → should emit fallback (the "Turn 2 had no streamed text but Turn 1 did" case is covered, but a "single turn with only tool calls" case isn't).

### Verdict

Approve.

— Reviewed by automated single-pass review (Claude executor log-shape correctness; full 4-tool battery skipped — diff is well-tested, the fix premise is correct, and the comment block makes the per-turn reset semantic explicit).

PR #3337, opened 2026-04-08
fix: copy packages/remote-web in Dockerfile (pnpm workspace glob)

Whaleylaw:
> [!NOTE]
> **Low Risk**
> Low risk change limited to the Docker build context; main impact is build behavior (pnpm install/cache layers) rather than runtime logic.
> 
> **Overview**
> Fixes the frontend Docker build by copying `packages/remote-web/package.json` before `pnpm install`, and then copying the full `packages/remote-web/` directory into the image.
> 
> This ensures pnpm workspace globs/dependency resolution include `remote-web` and prevents Docker layer caching/install from failing due to a missing workspace package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 285e1fe7dee04baead61952e59e10f69e874d737. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3395, opened 2026-04-25
chore: disable shutdown banners

everythings-gonna-be-alright:
## Disable shutdown UI for self-hosted deployments

Vibe Kanban Cloud is shutting down, but self-hosted / private-cloud operators are unaffected and should not see the cloud-specific shutdown UI. This PR adds a build-time opt-out flag for those deployments.

#### `VITE_DISABLE_SHUTDOWN_NOTIFICATION` build-time env var

When set to any truthy value, the following shutdown-related UI is suppressed:

- **Cloud shutdown banner** (`CloudShutdownExportBanner`) — returns `null` instead of rendering; also excluded from `showCloudShutdownBanner` logic in `SharedAppLayout` so the grid row is never added
- **Project sunset page** — `ProjectKanban` skips the `ProjectSunsetPage` redirect and renders the full kanban board instead (all previously removed kanban components are restored behind the flag)

#### `Dockerfile`

- Added `ARG VITE_DISABLE_SHUTDOWN_NOTIFICATION` + `ENV` passthrough in the `fe-builder` stage so the flag is available to Vite during image builds
- Fixed missing `patches/` directory copy — `pnpm install --frozen-lockfile` was failing with `ENOENT` because patch files referenced in `patchedDependencies` were not present in the layer before install ran
- Added all workspace crates missing from the Rust builder stage (`relay-client`, `relay-types`, `relay-tunnel`, `remote`, `remote-info`, `client-info`, `preview-proxy`, `embedded-ssh`, `desktop-bridge`) — `cargo build` was failing with manifest-not-found errors

#### `README.md`

- Added `VITE_DISABLE_SHUTDOWN_NOTIFICATION` to the environment variables table
- Added usage examples to the Self-Hosting section for both Docker and plain pnpm builds

### Usage

```bash
# Docker
docker build --build-arg VITE_DISABLE_SHUTDOWN_NOTIFICATION=true .

# pnpm
VITE_DISABLE_SHUTDOWN_NOTIFICATION=true pnpm run build

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the `ProjectKanban` rendering path (restoring full kanban layout, panel sizing persistence, and action mutation wiring) behind an env-gated branch, which could introduce regressions in navigation/layout when the flag is enabled.
> 
> **Overview**
> Adds a build-time env flag, `VITE_DISABLE_SHUTDOWN_NOTIFICATION`, to **disable all cloud shutdown UX** (banner + sunset/retired project page) for private/self-hosted builds.
> 
> When the flag is set, `ProjectKanban` now renders the full kanban experience (including right sidebar split-pane layout and action-based issue mutations like duplicate/remove) and redirects invalid workspace-create draft URLs back to the project view, while `CloudShutdownExportBanner` and `SharedAppLayout` stop rendering/allocating layout space for the shutdown banner.
> 
> Updates the `Dockerfile` to pass the new Vite build arg, copy `patches/` before `pnpm install`, and include missing Rust workspace crates to prevent `cargo build` manifest errors; `README.md` documents the new build-time variable and usage examples.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1efee2062ac0e7cdf0f746e00cdf3223bb2cc74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


tqnghia1998:
Hi, I tried your branch and it seems doesn't work with production deployment

cd crates/remote
docker compose --env-file ../../.env.remote -f docker-compose.prod.yml up -d --build

What else do I need to do?

tqnghia1998:
> Hi, I tried your branch and it seems doesn't work with production deployment
> 
> cd crates/remote docker compose --env-file ../../.env.remote -f docker-compose.prod.yml up -d --build
> 
> What else do I need to do?

Nvm it works now after I updated Dockerfile in crates/remote

dominikj-cf:
This seems to be a step in the right direction :+1:

However, do I understand it correct that this change does not add the capability to the standard npx path since it depends on Vite build with the specific disable banner setting?

yonch:
It seems we're adding a build variable to control the shutdown banner? With the cloud service has been sunset, maybe it makes more sense to just remove the banner unconditionally..

everythings-gonna-be-alright:
Sorry guys, I made this fix in a hurry, and the code is really far from ideal. If done properly, all of this should be passed through runtime variables.
I would redo it, but it probably doesn’t make much sense if the product’s server infrastructure is going to be decommissioned soon.

@stunningpixels would you mind if the banner is removed after May 10 to allow local versions to continue working?

Juanlucasbg:
## Aggressive review summary — PR #3395

269-line PR mixing two distinct concerns: (a) opt-out for the shutdown banner via `VITE_DISABLE_SHUTDOWN_NOTIFICATION` build-time env, and (b) Dockerfile expansion to copy ~10 previously-missing crates. Verdict: **mixed — split into two PRs and decide the banner question separately**.

### The two changes are unrelated

**Part A: Banner opt-out (`packages/web-core/src/shared/components/CloudShutdownExportBanner.tsx`, `SharedAppLayout.tsx`, `ProjectKanban.tsx`, `vite-env.d.ts`)**: Adds a build-time flag that suppresses the banner. This is a policy question — the maintainer explicitly added the banner (#3388) and the export-only routing (#3387) as part of communicating the sunset to users. Making it disable-able defeats that intent for self-hosted deployments. Whether operators should be able to suppress the sunset notice from their users is a maintainer call, not a code-quality call.

**Part B: Dockerfile crate copies**: Adds COPY lines for `client-info`, `desktop-bridge`, `embedded-ssh`, `preview-proxy`, `relay-client`, `relay-tunnel`, `relay-types`, `remote`, `remote-info`, plus a `COPY patches/ patches/` line. This is an objective bug fix — without these copies, `cargo build` inside the Docker context fails on missing workspace members. Same shape as PR #3337 (which I reviewed earlier and recommended merging) — and likely overlaps with it.

### Findings

**Part A (banner opt-out)**:
- **Policy — MED**: Conflicts with the apparent maintainer intent of the sunset banner. If self-hosters want to keep using vibe-kanban past the official sunset, the banner is the user's signal to migrate. Suppressing it puts users at risk of using an unmaintained product without knowing.
- **Structural — PASS**: Standard Vite env-flag pattern; `vite-env.d.ts` correctly typed.
- **NIT**: Default behavior unchanged (banner shows). Only an explicit `VITE_DISABLE_SHUTDOWN_NOTIFICATION=true` build hides it. So existing deployments aren't affected.

**Part B (Dockerfile crate copies)**:
- **Structural — PASS**: Mechanical copy expansion. The list of new crates matches what the workspace has accumulated since the Dockerfile was last touched.
- **Adversarial — LOW**: Same crate-set issue as PR #3337 — merging both will need a clean rebase. If #3337 lands first, this PR's diff for the workspace COPY block will conflict cleanly.
- **`COPY patches/ patches/`**: New addition. Verify this is needed for `pnpm-workspace.yaml`'s `patchedDependencies` (which references `@pierre/diffs@1.1.4`). Without it, `pnpm install --frozen-lockfile` fails on missing patch file. Probably a real fix.

### Recommended action

1. **Split the PR**:
   - "fix(docker): copy all workspace crates and patches/" — merge readily, same shape as #3337.
   - "feat: VITE_DISABLE_SHUTDOWN_NOTIFICATION env" — maintainer decision on policy. If they want operators to control the banner, accept it; if they want the sunset signal to be unsuppressable, reject it.
2. **If kept as one PR**, mark as awaiting-decision until the maintainer signals on Part A.

### Verdict

Conditional approve — Part B is clean, Part A needs a maintainer call.

— Reviewed by automated single-pass review (split-concerns flag; the structural change is fine, the policy change is the decision point).

PR #3415, opened 2026-05-13
Add crates/remote/starter: one-command local self-host bootstrap

domjancik:
A small, self-contained scaffold under `crates/remote/starter/` that gives
people a single `make start` path to a working local self-hosted Vibe Kanban,
without modifying anything in the base `crates/remote/docker-compose.yml`.

Posting as a draft to gauge interest / placement / scope before polishing.

## What it adds

```
crates/remote/starter/
├── Makefile
├── README.md
├── setup.sh
├── env.remote.template
├── docker-compose.override.yml   # bind-mount Postgres + Electric data dirs
├── docker-compose.ports.yml      # Windows-safe host ports (13000/13333/15433)
├── docker-compose.no-ssh.yml     # drop build-time SSH agent forwarding
└── .gitignore
```

All four compose files are layered via `-f` flags from the Makefile — the
base `crates/remote/docker-compose.yml` is **not modified**. If this dir is
deleted, nothing else changes.

## Why this and not just docs

The existing `docs/self-hosting/local-development.mdx` walks through the same
moving parts (JWT secret generation, `.env.remote`, port mapping, OAuth-or-bootstrap
auth, desktop client pointed at `VK_SHARED_API_BASE`), but as prose. This is the
same flow, scripted, with sensible defaults:

- **Windows-safe ports out of the box.** Hyper-V / WSL2 commonly reserve
  3000-3009 and 5432-5433 dynamically at boot; defaulting to 13000 / 13333 /
  15433 avoids the surprise.
- **SSH-agent forwarding off by default.** The build only needs `ssh:` for
  the private `billing` crate; with `FEATURES` empty, removing the SSH mount
  also makes the build work on machines without an SSH agent.
- **Bind-mounted data dirs** under `SELFHOST_DATA_DIR` so `docker volume
  prune` or a Docker VM wipe doesn't take Postgres with it.
- **Random admin password** generated into `.env.remote` (mode `0600`).
  `make start` prints the credentials each time it brings the stack up.
- **Idempotent setup.** Re-running `setup.sh` preserves the generated JWT
  and admin password, only updates ports / data dir / email.

## Usage

```bash
cd crates/remote/starter
make start    # ~10-15 min first build, seconds after
```

`make start` notices there's no `.env.remote` yet, prompts to run setup,
generates secrets, brings docker up, and launches the frontend app
(`npx vibe-kanban@latest`) in the foreground pointed at `localhost:13000`.

Day-to-day:

| | |
|---|---|
| `make stop` | `docker compose down` |
| `make restart` | stop + up |
| `make rebuild` | `up --build` (pulls in upstream changes) |
| `make logs` | tail `remote-server` |
| `make status` | `docker compose ps` |
| `make backup` | `pg_dump` to `backup-<timestamp>.sql` |
| `make clean` | tear down + wipe volumes + data dir (asks) |

`./setup.sh -y` skips all prompts and accepts defaults (CI / re-provisioning).

## Scope

Single-user, local-only. Doesn't try to do OAuth, relay/tunnel, or
attachments — those stay configurable via `.env.remote` and the existing
profiles. Linux/macOS focus; Windows users should run from WSL.

## Happy to adjust

If `crates/remote/starter/` isn't the right place, easy to move (e.g. a
top-level `self-host/`, under `docs/self-hosting/starter/`, or kept out of
tree entirely as a separate repo). Same goes for the scope — could keep
only the script + remove the Makefile, or vice versa.

### DIFF
--- a/crates/remote/starter/.gitignore
+++ b/crates/remote/starter/.gitignore
@@ -1,2 +1,0 @@
-backup-*.sql
-.DS_Store
--- a/crates/remote/starter/.gitignore
+++ b/crates/remote/starter/.gitignore
@@ -1,0 +1,2 @@
+backup-*.sql
+.DS_Store
--- a/crates/remote/starter/Makefile
+++ b/crates/remote/starter/Makefile
@@ -8,7 +8,7 @@
 #   make rebuild   → up with --build (pulls in upstream changes)
 #   make logs      → tail remote-server logs
 #   make status    → docker compose ps
-#   make backup    → pg_dump to backup-<timestamp>.sql
+#   make backup    → pg_dump to $SELFHOST_DATA_DIR/backups/<timestamp>.sql
 #   make clean     → tear down and wipe volumes + data dir (destructive)
 
 SHELL := /bin/bash
@@ -80,7 +80,9 @@ status: _ensure_setup
 
 backup: _ensure_setup
 	@. "$(ENV_FILE)" && \
-	  f="backup-$$(date +%Y%m%d-%H%M%S).sql"; \
+	  dest="$$SELFHOST_DATA_DIR/backups" && \
+	  mkdir -p "$$dest" && \
+	  f="$$dest/backup-$$(date +%Y%m%d-%H%M%S).sql" && \
 	  PGPASSWORD=remote pg_dump -h localhost -p "$${REMOTE_DB_PORT:-15433}" -U remote remote > "$$f" && \
 	  echo "wrote $$f"
 
--- a/crates/remote/starter/Makefile
+++ b/crates/remote/starter/Makefile
@@ -1,0 +1,94 @@
+# crates/remote/starter — local self-host operations.
+#
+# Primary entry points:
+#   make start     → docker up + launch frontend app (prompts to run setup if missing)
+#   make up        → just docker (no frontend app); use from a second terminal
+#   make stop      → docker compose down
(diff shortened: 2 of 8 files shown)


domjancik:
A note on the Makefile choice, since it's the most opinionated bit of the PR and breaks from the rest of the repo's `pnpm run …` convention:

**Why `make` and not pnpm scripts?**

1. **No bootstrap prerequisite.** The starter is meant to be *the* entry point. Asking users to install Node + pnpm just to bring up the docker stack feels backwards. `make` is on every Linux / macOS box out of the box and available via WSL on Windows.
2. **Self-documenting target list.** `make help` parses the leading comment of the Makefile and prints the contract — single source of truth for the available ops.
3. **Recipe dependencies.** `start: up`, `_ensure_setup` as a chained guard before every docker-touching recipe — these express intent more cleanly than `pnpm run a && pnpm run b` or shell-function dispatch.
4. **The substance is shell anyway.** Every recipe wraps `docker compose …`; wrapping that in pnpm scripts adds an indirection without changing what runs.

**Happy to flip to pnpm if that's the preferred direction here.** That would mean a new `crates/remote/starter/package.json` with the targets as `"scripts"`, `bash setup.sh` invoked directly, and a tiny bit of node-y plumbing for the `_ensure_setup` prompt. Let me know which you prefer and I'll re-roll.

PR #3418, opened 2026-05-21
fix: exclude ambiguous characters from relay pairing codes

### LINKED ISSUE #3359: Exclude ambiguous characters from relay pairing codes
## Problem

The relay pairing code charset (`ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789` in `crates/trusted-key-auth/src/spake2.rs:10`) includes visually ambiguous characters like `O`/`0` and `I`/`1`/`L`. This can cause users to mistype the code — e.g., entering `0` when the code contains `O` — which silently consumes the one-time code and returns an "Unauthorized" error requiring a fresh code to be generated.

## Suggested fix

Replace the charset with one that excludes ambiguous characters, e.g. Crockford's Base32 alphabet:

```
0123456789ABCDEFGHJKMNPQRSTVWXYZ
```

This removes `I`, `L`, `O`, `U` to avoid confusion with `1`, `L`, `0`, `V`.

Alternatively, a simpler approach: remove `O`, `I`, `L` from the existing charset.

### Files involved

- `crates/trusted-key-auth/src/spake2.rs` — `ENROLLMENT_CODE_CHARSET` constant and `generate_one_time_code()`
- `packages/web-core/src/shared/lib/relayPake.ts` — `normalizeEnrollmentCode()` (could optionally map `O`→`0`, `I`→`1` as a fallback)
- `packages/web-core/src/shared/dialogs/settings/settings/PairingCodeInput.tsx` — uses a monospace font, but could also benefit from a font that distinguishes these characters

Alexi5000:
﻿## Summary

- **Removes visually ambiguous characters** (`O`, `I`, `L`, `U`) from `ENROLLMENT_CODE_CHARSET` to prevent mistyping relay pairing codes
- **Adds fallback normalization** in both backend (Rust) and frontend (TypeScript) to map commonly confused characters: `O`->`0`, `I`->`1`, `L`->`1`, `U`->`V`
- Uses a Crockford Base32-inspired alphabet that avoids character pairs that look identical in most monospace fonts

## Problem

The current charset includes characters that are visually indistinguishable in many fonts (e.g., `O` vs `0`, `I` vs `1`, `L` vs `1`). When a user mistypes one of these, the one-time code is silently consumed and returns "Unauthorized", requiring a fresh code.

## Changes

| File | Change |
|------|--------|
| `crates/trusted-key-auth/src/spake2.rs` | Updated `ENROLLMENT_CODE_CHARSET` to exclude `O`, `I`, `L`, `U`; added normalization mapping in `normalize_enrollment_code()`; added tests |
| `packages/web-core/src/shared/lib/relayPake.ts` | Added matching normalization in `normalizeEnrollmentCode()` |

## Charset

Before: `ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789` (36 chars)
After: `ABCDEFGHJKMNPQRSTVWXYZ0123456789` (32 chars)

Removed: `I` (looks like `1`), `L` (looks like `1`), `O` (looks like `0`), `U` (looks like `V`)

Entropy impact: 6-character code goes from 36^6 (2.18B) to 32^6 (1.07B) — still well above brute-force threshold for a one-time code.

Fixes #3359

## Test plan

- [ ] `cargo test -p trusted-key-auth` — verify new and existing tests pass
- [ ] Generate 1000+ codes and verify none contain `O`, `I`, `L`, or `U`
- [ ] Type a code containing `O` on the frontend — verify it normalizes to `0`
- [ ] Pair a device using the new charset — verify pairing succeeds

### DIFF
--- a/crates/trusted-key-auth/src/spake2.rs
+++ b/crates/trusted-key-auth/src/spake2.rs
@@ -7,7 +7,8 @@ use crate::error::TrustedKeyAuthError;
 const SPAKE2_CLIENT_ID: &[u8] = b"vibe-kanban-browser";
 const SPAKE2_SERVER_ID: &[u8] = b"vibe-kanban-server";
 pub const ENROLLMENT_CODE_LENGTH: usize = 6;
-const ENROLLMENT_CODE_CHARSET: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+// Excludes visually ambiguous characters: O (vs 0), I (vs 1), L (vs 1), U (vs V)
+const ENROLLMENT_CODE_CHARSET: &[u8] = b"ABCDEFGHJKMNPQRSTVWXYZ0123456789";
 
 #[derive(Debug)]
 pub struct Spake2StartOutcome {
@@ -68,6 +69,15 @@ pub fn normalize_enrollment_code(raw_code: &str) -> Result<String, TrustedKeyAut
         ));
     }
 
+    // Map visually ambiguous characters to their intended equivalents.
+    // New codes no longer contain O/I/L/U, but users may still type them
+    // when reading a code displayed in a font where 0≈O or 1≈I≈L.
+    let code = code
+        .replace('O', "0")
+        .replace('I', "1")
+        .replace('L', "1")
+        .replace('U', "V");
+
     Ok(code)
 }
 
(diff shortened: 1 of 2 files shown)


Alexi5000:
Closing — vibe-kanban looks sunset (export-only, no recent commits), so this won't get merged here. Keeping the change on my fork in case it's useful later. Thanks for vibe-kanban!

PR #3436, opened 2026-06-26
Fix docker build: copy pnpm patches/ and all workspace crates

rafaelfiguereod-stack:
## What

`docker build .` fails on a clean checkout of `main` for two independent, unrelated reasons in the `Dockerfile`. This fixes both so the image builds end to end.

## 1. `pnpm install --frozen-lockfile` can't find the patch

`pnpm-workspace.yaml` declares a patched dependency:

```yaml
patchedDependencies:
  '@pierre/diffs@1.1.4': patches/@pierre__diffs@1.1.4.patch
```

…but the `fe-builder` stage copies the lockfile/manifests without `patches/`, so the install aborts:

```
 ENOENT  ENOENT: no such file or directory, open '/app/patches/@pierre__diffs@1.1.4.patch'
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @vibe/local-web build ...  exit code: 254
```

**Fix:** `COPY patches/ patches/` before the install.

## 2. The per-crate `COPY` list is missing 7 workspace members

The `builder` stage copies crate manifests and sources one-by-one, but the list has drifted from `Cargo.toml`'s `[workspace] members`. These members are never copied: `relay-client`, `relay-types`, `client-info`, `remote-info`, `embedded-ssh`, `desktop-bridge`, `preview-proxy`. So `cargo build --locked` fails immediately:

```
error: failed to load manifest for workspace member `/app/crates/relay-client`
  failed to read `/app/crates/relay-client/Cargo.toml`: No such file or directory
```

**Fix:** replace the enumerated `COPY crates/<name>/…` lines with a single `COPY crates/ crates/`, which can't drift out of sync with the workspace. The manifest-then-source split wasn't enabling dependency-layer caching here (there's no intermediate `cargo build` between the two copy groups), so collapsing it has no caching downside, and `.dockerignore` already excludes `target/` and `node_modules/`. If you'd rather keep the split, I'm happy to instead just add the seven missing crates to the existing lists.

## Verified

- Both failures reproduce on a clean `main` checkout.
- `cargo metadata --locked` resolves the workspace with the change.
- `docker build .` completes end-to-end and the resulting image boots — `/health` returns 200 and the web UI is served.

— Rafael (rafaelfiguereod-stack)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
 COPY packages/local-web/package.json packages/local-web/package.json
 COPY packages/ui/package.json packages/ui/package.json
 COPY packages/web-core/package.json packages/web-core/package.json
+# pnpm-workspace.yaml's patchedDependencies references
+# patches/@pierre__diffs@1.1.4.patch, so the frozen-lockfile install needs the
+# patch present in the build context (otherwise ENOENT, pnpm exits 254).
+COPY patches/ patches/
 
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     pnpm install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


rafaelfiguereod-stack:
Checking back on this one: the fix still applies cleanly to current `main`, and the bug it fixes is still there. Verifiable in about thirty seconds:

`Dockerfile` copies three files and then installs:

```
:19   COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
:25   pnpm install --frozen-lockfile
```

`pnpm-workspace.yaml` declares a patch:

```yaml
patchedDependencies:
  '@pierre/diffs@1.1.4': patches/@pierre__diffs@1.1.4.patch
```

`patches/@pierre__diffs@1.1.4.patch` exists in the repo, but `patches/` is never copied into the build context, so the install at line 25 is asked to apply a file that is not there and the image build fails.

The `Dockerfile` has not changed since `114b114` (2026-03-27), which is before this PR was opened, so nothing has superseded it.

No urgency from my side and I am not trying to nudge the queue. I only wanted to note that this has not gone stale, in case its age made it look that way. Happy to rebase, split it smaller, or close it if you would rather solve it another way.

PR #3423, opened 2026-05-28
Adds pnpm-workspace.yaml overrides to pin transitive dependencies to patched versions, eliminating all 36 npm audit findings (21 high, 10 moderate, 5 low).

rafaelfiguereod-stack:

PR #3447, opened 2026-09-01
fix: correct swapped stderr/stdout labels in git CLI error output

rafaelfiguereod-stack:
When a git command fails with output on only one stream, the labels are reversed. A stdout-only failure prints under `--- stderr`, and a stderr-only failure under `--- stdout`. The both-streams case is already correct.

The effect is that git failures point the reader at the wrong stream.

Split out of #3423, which bundled several unrelated changes under a single title. Filing them separately so each can be reviewed on its own.

### DIFF
--- a/crates/git/src/cli.rs
+++ b/crates/git/src/cli.rs
@@ -824,8 +824,8 @@ impl GitCli {
             let combined = match (stdout.is_empty(), stderr.is_empty()) {
                 (true, true) => "Command failed with no output".to_string(),
                 (false, false) => format!("--- stderr\n{stderr}\n--- stdout\n{stdout}"),
-                (false, true) => format!("--- stderr\n{stdout}"),
-                (true, false) => format!("--- stdout\n{stderr}"),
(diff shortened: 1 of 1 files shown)


rafaelfiguereod-stack:
Small request, and no urgency attached to it.

All seven of these PRs have workflow runs sitting at `action_required`, so none of them currently have a CI signal. That is expected for a first-time contributor, but it does mean there is nothing green for you to look at when triaging them.

If you have a moment to approve the runs, that would help. If seven is more clicking than you want to do, #3453 is the one where CI adds the most: its `Cargo.lock` changes reach the whole workspace, and the local run I posted there only covered the `embedded-ssh` crate. The other six are small enough to read directly, three of them are single-file and this one is two lines.

Also fine to leave them sitting and look on your own schedule. I would rather not add noise to your queue.

For context, these seven were split out of #3423, which I closed. That one was opened from my fork's `main` branch and dragged 21 commits into a single review, which was my mistake.

anastasiya1155 (approved):

PR #3450, opened 2026-09-01
fix: reject non-positive installation_id in GitHub App callback

rafaelfiguereod-stack:
The GitHub App callback accepts any `i64` `installation_id` from the query string and carries it forward with only a missing-parameter check. This adds a range check rejecting values less than or equal to zero.

A code scanning alert classified this as SSRF. To be accurate about scope, it is not: the value is an integer and cannot inject a host. Filing it as plain input validation rather than as a security fix. Happy to close it if you consider the check unnecessary.

Split out of #3423.


anastasiya1155:
Hey @rafaelfiguereod-stack thanks for your contributions, but this one seems unnecessary

PR #3453, opened 2026-09-01
fix(deps): upgrade Rust dependencies with open RUSTSEC advisories

rafaelfiguereod-stack:
Three dependency upgrades for open advisories:

- russh 0.48 to 0.61, for RUSTSEC-2026-0153 and RUSTSEC-2026-0154. This one needs an API migration across `crates/embedded-ssh` (config, handler, sftp). Smoke tests are added covering the SSH and SFTP paths through the new version.
- aws-lc bumped to a patched release
- rustls-webpki 0.103.10 to 0.103.13

The russh upgrade is the substantive change and deserves the most scrutiny. I have not run the full workspace build locally, so CI is the real check here.

Split out of #3423.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,7 +34,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "crypto-common 0.1.7",
- "generic-array",
+ "generic-array 0.14.7",
+]
+
+[[package]]
+name = "aead"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef60ac202874e574ce7a7158cc8bca7313dd344322482e4fadee288bf4a306b8"
+dependencies = [
+ "crypto-common 0.2.2",
+ "inout 0.2.2",
 ]
 
 [[package]]
@@ -44,8 +54,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
- "cipher",
+ "cipher 0.4.4",
  "cpufeatures 0.2.17",
+]
+
+[[package]]
+name = "aes"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f1fc76eaeac4c9164506c466d4ffdd8ec9d0c5bf57ee97177c4d8eceb3a0e138"
+dependencies = [
+ "cipher 0.5.2",
+ "cpubits",
+ "cpufeatures 0.3.0",
+ "zeroize",
 ]
 
 [[package]]
@@ -54,12 +76,27 @@ version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
 dependencies = [
- "aead",
- "aes",
- "cipher",
- "ctr",
- "ghash",
+ "aead 0.5.2",
+ "aes 0.8.4",
+ "cipher 0.4.4",
+ "ctr 0.9.2",
+ "ghash 0.5.1",
  "subtle",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.11.0-rc.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da8c919c118108f144adecad74b425b804ad075580d605d9b33c2d6d1c62a2f8"
+dependencies = [
+ "aead 0.6.0",
+ "aes 0.9.1",
+ "cipher 0.5.2",
+ "ctr 0.10.1",
+ "ghash 0.6.0",
+ "subtle",
+ "zeroize",
 ]
 
 [[package]]
@@ -99,7 +136,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "const-random",
  "getrandom 0.3.4",
  "once_cell",
  "version_check",
@@ -214,7 +250,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -225,7 +261,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -283,6 +319,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a07d1f37ff60921c83bdfc7407723bdefe89b44b98a9b772f225c8f9d67141a6"
 dependencies = [
  "rustversion",
+]
+
+[[package]]
+name = "argon2"
+version = "0.6.0-rc.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7af50940b73bf4e16c15c448a2b121c63f2d68e3e54b6a8731673cb4aa0cdff5"
+dependencies = [
+ "base64ct",
+ "blake2",
+ "cpufeatures 0.3.0",
+ "password-hash",
 ]
 
 [[package]]
@@ -582,9 +630,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.16.0"
+version = "1.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9a7b350e3bb1767102698302bc37256cbd48422809984b98d292c40e2579aa9"
+checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
 dependencies = [
  "aws-lc-sys",
  "untrusted 0.7.1",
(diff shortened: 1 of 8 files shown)


rafaelfiguereod-stack:
Adding test evidence, since there is no CI signal on this one yet. The workflow run is sitting at `action_required`, which is normal for a first-time contributor.

Ran the `embedded-ssh` crate against this repo's own pinned toolchain, in a Linux container.

`cargo test -p embedded-ssh`

```
Running tests/ssh_smoke.rs

test sftp_subsystem_round_trips_a_file ... ok
test exec_channel_roundtrips_stdout ... ok
test rejects_unauthorized_public_key ... ok

test result: ok. 3 passed; 0 failed; 0 ignored
```

`cargo check -p embedded-ssh --all-targets`: 0 errors, 0 warnings.

Toolchain was `nightly-2025-12-04-x86_64-unknown-linux-gnu`, resolved from `rust-toolchain.toml` rather than pinned by hand, so it matches what CI would use. `aws-lc-sys 0.41.0` built with cmake 3.25.1.

Scope, so this is not read as more than it is:

- It covers `embedded-ssh` only, not the whole workspace. The `Cargo.lock` changes reach much further than that, so a workspace-wide build is still worth having and CI is the right place for it.
- The three tests are the ones added in this PR: an exec channel round trip, an SFTP file round trip, and rejection of an unauthorized public key. The last one is the reason I wanted them run rather than just compiled. A crypto dependency migration that builds cleanly while quietly accepting credentials it should reject is the failure mode worth ruling out explicitly.

If it would make review easier, I am happy to split the russh upgrade away from the aws-lc and rustls-webpki bumps. They are in one PR because they share `Cargo.lock`, not because they need to land together.

rafaelfiguereod-stack:
Heads up that the red check on this PR is not the patch.

The `changes` job failed in 3 seconds with an empty step list, and its only annotation is:

> The job was not started because your account is locked due to a billing issue.

Every other job (`backend-clippy`, `backend-test`, `backend-schema-checks`, `backend-remote-checks`, `frontend-checks`, `tauri-checks`) reports `skipping`, because they are all gated on `changes`. So the whole matrix is sitting behind a job that was never allowed to start.

It is not specific to this PR. A recent run on `main` carries the identical annotation, and other branches in the repo are failing the same way, with runs completing as failures without creating any jobs.

Nothing to do on my side, and no push would turn this green. Flagging it in case the red check reads as a broken patch rather than an account lock. Happy to rebase or push an empty commit to re-trigger once Actions is unlocked, just say the word.

Rafael

rafaelfiguereod-stack:
Since CI cannot run this PR at the moment, I reviewed it properly myself and am posting what I found rather than leaving it unverified. Several of these are things I should have put in the description in the first place.

A local `cargo clippy --workspace --all-targets --exclude vibe-kanban-tauri -- -D warnings` and `cargo test` run, matching what `backend-clippy` and `backend-test` do, is still in progress. I will follow up with the result.

## The thing I most owe you up front: this pulls in a lot of pre-release crypto

russh 0.61 brings the RustCrypto stack forward with it. Pre-release crypto crates in `Cargo.lock` go from **2 to 17**.

Base had `curve25519-dalek 5.0.0-pre.6` and `hkdf 0.13.0-rc.6`. This branch adds `ssh-key 0.7.0-rc.10`, `ed25519-dalek 3.0.0-rc.0`, `curve25519-dalek 5.0.0-rc.0`, `rsa 0.10.0-rc.18`, `ecdsa 0.17.0-rc.18`, `elliptic-curve 0.14.0-rc.33`, `p256`/`p384`/`p521 0.14.0-rc.10`, `aes-gcm 0.11.0-rc.4`, `argon2 0.6.0-rc.8`, `blake2 0.11.0-rc.6`, `pkcs1 0.8.0-rc.4`, `ssh-cipher` and `ssh-encoding 0.3.0-rc.9`, `primefield`/`primeorder 0.14.0-rc.10`.

Two majors of `ed25519-dalek` end up in the same binary: `2.2.0` declared directly by `embedded-ssh`, and `3.0.0-rc.0` via `ssh-key`.

This is forced by russh 0.61 rather than anything I chose, but on a PR whose whole point is closing advisories it is a real trade and you should get to weigh it deliberately. If that is not acceptable, the alternative is staying on russh 0.48 and leaving RUSTSEC-2026-0153 and 2026-0154 open, which I do not think is better, but it is your call.

## `cargo audit` will still fail on this branch, and that is not fixable here

`rsa` RUSTSEC-2023-0071 (Marvin attack) has `patched = []`. No fixed release exists. This branch takes the number of affected `rsa` copies in the tree from one to two: `0.9.10` was already there via `jsonwebtoken` and `desktop-bridge -> ssh-key 0.6.7`, and `0.10.0-rc.18` arrives with russh 0.61 and `ssh-key 0.7.0-rc.10`.

I would rather say that plainly here than have someone merge this expecting a clean audit. If you want an `audit.toml` ignore entry with a justification comment, I am happy to add one.

## One defect in my own comment, which I will fix

`Cargo.toml` says `aws-lc-sys >=0.38.0 fixes RUSTSEC-2026-0044/0045/0046/0047/0048`. Two of those five need **0.39.0**, not 0.38.0:

| Advisory | Patched |
|---|---|
| 2026-0044 X.509 name-constraints bypass | >=0.39.0 |
| 2026-0045 AES-CCM tag timing side-channel | >=0.38.0 |
| 2026-0046 PKCS7_verify chain validation bypass | >=0.38.0 |
| 2026-0047 PKCS7_verify signature validation bypass | >=0.38.0 |
| 2026-0048 CRL distribution point scope check | >=0.39.0 |

Nothing vulnerable ships, because the actual pin is `=0.41.0` and that clears all five. But the comment is the only in-repo record of why the pin exists, so it is exactly what someone would read before relaxing it. Also, the parenthetical lists four descriptions for five IDs, because 0046 and 0047 are two distinct PKCS7_verify advisories.

## The `MethodSet` line is correct, but it is not the control it looks like

`config.rs` changes `MethodSet::PUBLICKEY` to `MethodSet::from(&[MethodKind::PublicKey][..])`. The set is unchanged and this does not widen anything.

Worth knowing for whoever reviews it: in russh 0.61 `config.methods` is read once, in `server/encrypted.rs`, and from there only encoded into the `USERAUTH_FAILURE` name-list. `server_read_auth_request` dispatches on the wire method string with no `MethodSet` check, so a client sending `password` would reach `handler.auth_password` regardless of what is in the set. What actually denies non-publickey auth is that `SshSessionHandler` implements only `auth_publickey` and every other `auth_*` falls through to russh's default reject.

So the line is fine today, but if anyone later adds an `auth_password` impl the publickey-only set will not stop it. I can add a one-line comment saying so.

## An undeclared behaviour change I should have called out

The sftp error conversion became `StatusReply::new(err.code).with_message(err.message)`. The `with_message` half is not required by the type migration. The old impl returned only `err.code` and never read `message`, which is why it carried `#[allow(dead_code)]`. `StatusReply::new(err.code)` alone would have preserved the previous wire behaviour exactly.

I checked whether this leaks anything before assuming it does not. All 14 `SftpError` construction sites are in `sftp.rs`, and `message` is either one of six hardcoded strings or `err.to_string()` from `impl From<std::io::Error>`. Rust's `std::fs` entry points do not attach path context to their `io::Error`, so errno text reaches the client but filesystem paths do not. The client is also already authenticated and has shell exec through the same handler.

So I do not think it is a disclosure problem, but it is a behaviour change riding inside a mechanical port and it should have been declared. Happy to drop the `.with_message(...)` if you would rather keep the wire output identical.

## Smaller things, all of which I will fix if you want them fixed

- **The new exec test has no Windows gate.** `exec_channel_roundtrips_stdout` runs `printf` via `SHELL` or `/bin/sh -c`, and `spawn_stdio_session` has no Windows branch. CI is unaffected since `backend-test` runs on `ubuntu-24.04`, but `cargo test --workspace` on a Windows checkout gains a new failure from this PR. `#[cfg_attr(windows, ignore)]` resolves it.
- **The `=0.7.0-rc.10` pin buys nothing today.** russh 0.61.2 already pins `ssh-key` exactly, so a caret resolves identically. The cost is later: `russh = "0.61"` is a caret, so if a russh patch moves to a different `ssh-key` rc, the two `=` requirements collide in the same semver range and `cargo update` hard-fails with `failed to select a version for ssh-key`.
- **`ssh-key` stays duplicated** at `0.6.7` and `0.7.0-rc.10`, because `crates/desktop-bridge/Cargo.toml` still asks for `0.6` and I did not touch it. Not an advisory, the DB holds none against `ssh-key`, but it does mean two independent private-key serialization paths in one binary.
- **Two leftovers:** a now-stale `#[allow(dead_code)]` on `SftpError.message` since it is read again, and a redundant `russh-sftp` entry I added to `[dev-dependencies]` when `[dependencies]` already provides it.

## On the test

It is a real integration test, not a construction test. It stands up a listener, runs the real server, connects a real russh 0.61 client, completes a real key exchange and a real publickey round trip in both the accept and reject shapes, runs a child process over an exec channel and reads its output, and drives a real SFTP session. It would not pass if the upgrade broke the protocol. The one changed line it does not exercise is the stderr path.

Tell me which of the above you want changed and I will push them. I would suggest at minimum the `0.38.0` to `0.39.0` correction and the Windows test gate, since those are unambiguous.

Rafael

rafaelfiguereod-stack:
Local verification result, as promised above. Run in a `rust:1-bookworm` container on `nightly-2025-12-04`, the same toolchain `RUST_TOOLCHAIN` pins.

**Clippy passes.**

```
cargo clippy --workspace --all-targets --exclude vibe-kanban-tauri -- -D warnings
Finished `dev` profile in 43m 33s
exit 0
```

That is `backend-clippy`'s command verbatim. The russh 0.48 to 0.61 port compiles clean across all 33 crates with no warnings, and the pinned `aws-lc-sys 0.41.0` and `ssh-key 0.7.0-rc.10` resolve and build.

**The new integration tests pass.**

```
Running tests/ssh_smoke.rs
test sftp_subsystem_round_trips_a_file ... ok
test exec_channel_roundtrips_stdout ... ok
test rejects_unauthorized_public_key ... ok
test result: ok. 3 passed; 0 failed; finished in 1.12s
```

That covers a real key exchange with the Ed25519 host key built through the new `russh::keys::PrivateKey` path, a publickey round trip in both the accept and the new `Auth::Reject { partial_success: false }` shape, an exec channel through the changed `handle.data(channel_id, Vec<u8>)` signature, and an SFTP session through the new `StatusReply` conversion.

**Three honest limits on the above, so nobody reads more into it than it supports.**

1. I ran `cargo test`, not `cargo nextest run`. Same tests, different harness. nextest runs each test in its own process, so shared-process state could in principle differ.
2. Only `embedded-ssh` was tested, not the whole workspace. A workspace-wide `cargo test` died with `rustc-LLVM ERROR: IO failure on output stream: Cannot allocate memory` on my machine, which is my memory limit, not anything in this branch. So crates outside `embedded-ssh` are compile-verified by clippy but their tests are unverified here.
3. `cargo fmt --all -- --check` fails, but not because of this PR. It is one import line in `crates/executors/src/executors/copilot.rs`, which this branch does not touch, and I reproduced the identical failure on `main` at `d5cbb53`. Filed separately as #3469. It does mean `backend-clippy` will show red on this PR for a reason that is not this PR, since that job runs fmt and clippy together.

Rafael

PR #3459, opened 2026-09-16
fix: an Enter that confirms an IME candidate is not an Enter

isamu:
## Summary

An Enter pressed to **confirm an IME candidate** (Japanese, Chinese, Korean) fires shortcuts and submits, so the line goes out half-composed.

`useSemanticKey` and `Input.tsx` both guard on `event.isComposing`, which is the right idea but not sufficient: **Safari fires `compositionend` BEFORE the confirming Enter's `keydown`**, so the flag is already `false` when the handler runs. Chrome and Firefox fire it after — which is why the naive guard looks correct on those two and the bug reads as Safari-only.

This is a documented WebKit ordering difference, not something specific to this project:

- [Understanding Composition Browser Events](https://developer.squareup.com/blog/understanding-composition-browser-events/) — Square, on the ordering
- [makeplane/plane#8893](https://github.com/makeplane/plane/pull/8893) — the same fix in Plane
- [tiann/hapi#1481](https://github.com/tiann/hapi/issues/1481) — the same bug with a Safari repro, since fixed

A maintainer of another tool measured it on their machine while confirming their own repro: `compositionend` arrived **3 ms** before the `keydown`.

## The fix

```ts
export function isImeConfirmation(event: Pick<KeyboardEvent, 'isComposing'>): boolean {
  if (event.isComposing || composing) return true;
  return performance.now() - lastCompositionEndAt < SAFARI_IME_RACE_WINDOW_MS; // 30
}
```

The `compositionend` → `keydown` pair is synchronous (microseconds); a human pressing Enter a second time takes 100 ms or more, so a deliberate keypress never lands inside the window. **The guard only ever becomes more conservative** — on Chrome and Firefox `isComposing` is already `true` and nothing changes.

Composition is tracked on `window` in the **capture phase** rather than per field, so a composition that starts in one input and ends after focus has moved still closes. `blur` does not bubble, but a capture listener still sees it on the way down, which closes an abandoned composition.

Wired into the two places that guard today:

| | |
| --- | --- |
| `packages/web-core/src/shared/keyboard/useSemanticKey.ts` | covers **every registered shortcut** |
| `packages/ui/src/components/Input.tsx` | the Enter-to-submit path |

## Items to confirm / review

**The helper lives in `@vibe/ui` and `web-core` imports it.** An earlier revision copied it into both packages; that was wrong. I had checked whether `@vibe/ui` depends on `@vibe/web-core`, which is the wrong direction — `@vibe/web-core` already depends on `@vibe/ui` as `workspace:*`, and the package exports `./lib/*`, so the edge was already there.

Disclosure: I maintain a tool in the same space and hit this bug there first; this is the same fix.


anastasiya1155:
Hey @isamu, thanks for your contribution, please leave this helper only in ui package and import it in web-core instead of duplicating. Also make sure all checks pass locally:
```
npx concurrently \
            --kill-others-on-fail \
            --names "local:lint,local:fmt,local:build,remote:fmt,remote:build,ui:check,ui:lint,ui:fmt,core:check,core:fmt,i18n,i18n:unused,legacy" \
            --timings \
            "cd packages/local-web && npm run lint" \
            "cd packages/local-web && npm run format:check" \
            "cd packages/local-web && npm run build" \
            "cd packages/remote-web && npm run format:check" \
            "cd packages/remote-web && npm run build" \
            "cd packages/ui && npm run check" \
            "cd packages/ui && npm run lint" \
            "cd packages/ui && npm run format:check" \
            "cd packages/web-core && npm run check" \
            "cd packages/web-core && npm run format:check" \
            "GITHUB_BASE_REF=${{ github.base_ref || 'main' }} ./scripts/check-i18n.sh" \
            "node scripts/check-unused-i18n-keys.mjs" \
            "./scripts/check-legacy-frontend-paths.sh"
 ```

**Comment by isamu** [comment]
Both done, thank you — and the duplication was my mistake rather than a judgement call.
I checked whether `@vibe/ui` depends on `@vibe/web-core`, which is the wrong direction:
`@vibe/web-core` already depends on `@vibe/ui` as `workspace:*`, and the package exports
`./lib/*`, so the edge I said I did not want to create was already there. The helper now
lives only in `packages/ui/src/lib/imeComposition.ts` and `useSemanticKey` imports it from
`@vibe/ui/lib/imeComposition`. The PR is three files.

Your command, run locally on this branch:

| | |
|---|---|
| `local-web` lint | **fails, and fails the same way on a clean `main`** — see below |
| `local-web` format:check, build | pass |
| `remote-web` format:check, build | pass |
| `ui` check, format:check | pass |
| `ui` lint | **same pre-existing failure** |
| `web-core` check, format:check | pass |
| `check-i18n.sh`, `check-unused-i18n-keys.mjs`, `check-legacy-frontend-paths.sh` | pass |

`format:check` did catch my helper; it is formatted with the repo's prettier config now.

The two `lint` failures are not from this branch. Both scripts run
`eslint . --ext ts,tsx`, and with `eslint.config.js` in place ESLint rejects the flag
outright:

```
Invalid option '--ext' - perhaps you meant '-c'?
You're using eslint.config.js, some command line flags are no longer available.
```

I confirmed it by stashing my changes and running it against a clean tree — same error, no
files linted. It looked out of scope for this PR, so I left it alone; happy to send a
separate one dropping `--ext` from both scripts if that is useful.

**Comment by isamu** [comment]
Both done, and re-reading the helper for this turned up three defects in it. The branch is
now two commits.

## The helper lives in `@vibe/ui`

`web-core` imports it from `@vibe/ui/lib/imeComposition`. Copying it into both packages was
my mistake rather than a judgement call: I checked whether `@vibe/ui` depends on
`@vibe/web-core`, which is the wrong direction. `@vibe/web-core` already depends on
`@vibe/ui` as `workspace:*`, and the package exports `./lib/*`, so the edge I said I did
not want to create was already there.

## Three defects in the helper itself

**A startup window that swallowed Enter.** `lastCompositionEndAt` started at `0`, and
`performance.now()` counts from the page's time origin, so `now - 0` was inside the 30ms
window for the first 30ms of the page — an Enter that no composition preceded was dropped.
Measured at 16ms after start, the comparison was true.

**The window applied to every key.** `useSemanticKey` guards every registered shortcut, so
for 30ms after any composition ended, `esc`, `j`, `k` and the rest were dropped along with
Enter. An open composition still swallows every key, because none of them reached the
application — but the window *after* it is Enter-only: it exists for the keystroke that
confirms a candidate, and that keystroke is Enter. The comment at that call site claimed
"this Enter" while the code applied to everything; it now says what the code does.

**`sideEffects: false` was not true for this module.** `@vibe/ui` declared it while the
helper registers `window` listeners on import. The declaration now names this one file, so
a bundler cannot drop the registration on the strength of that claim. Happy to drop this
hunk if you would rather keep the field as a flat `false`.

## Your command, run locally on this branch

| | |
|---|---|
| `local-web` format:check, build | pass |
| `remote-web` format:check, build | pass |
| `ui` check, format:check | pass |
| `web-core` check, format:check | pass |
| `check-i18n.sh`, `check-unused-i18n-keys.mjs`, `check-legacy-frontend-paths.sh` | pass |
| `local-web` lint, `ui` lint | **fail, and fail the same way on a clean `main`** |

Both `lint` scripts run `eslint . --ext ts,tsx`, and with `eslint.config.js` in place ESLint
rejects the flag outright:

```
Invalid option '--ext' - perhaps you meant '-c'?
You're using eslint.config.js, some command line flags are no longer available.
```

I confirmed it by stashing my changes and running against a clean tree — same error, no
files linted. It looked out of scope here, so I left it; happy to send a separate PR
dropping `--ext` from both scripts.

I did not add tests: none of these packages has a `test` script wired, so there is no runner
to add one to. Say the word if you would like one and I will set it up.

anastasiya1155 (approved):

